PR #2, opened 2019-01-28
handle CORS pre-flight requests

paulswartz:
Previously, we weren't properly matching the OPTIONS requests made by browsers. This moves some code around so that the OPTIONS requests are matched before we hit the router, allowing them to be checked.

The majority of the changes are to split up the `api_user` (which is a key) from `user` (which is a developer who can have multiple keys).

### DIFF
--- a/apps/api_web/test/api_web/router_test.exs
+++ b/apps/api_web/test/api_web/router_test.exs
@@ -57,6 +57,10 @@ defmodule ApiWeb.RouterTest do
 
       response = options(conn, "/_health")
       assert response.status == 200
+      assert ["*"] = get_resp_header(response, "access-control-allow-origin")
+      assert ["GET"] = get_resp_header(response, "access-control-allow-methods")
+      assert [headers] = get_resp_header(response, "access-control-allow-headers")
+      assert headers =~ "x-api-key"
     end
   end
 end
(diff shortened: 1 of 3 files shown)


apps/api_web/test/api_web/router_test.exs
```diff
@@ -43,4 +43,20 @@ defmodule ApiWeb.RouterTest do
...
+    test "returns the appropriate headers when an OPTIONS request is made", %{conn: conn} do
+      conn =
+        conn
+        |> put_req_header("x-api-key", conn.assigns.api_key)
+        |> put_req_header("access-control-request-headers", "x-api-key")
+        |> put_req_header("access-control-request-method", "GET")
+        |> put_req_header("origin", "http://localhost/")
+
+      response = options(conn, "/_health")
+      assert response.status == 200
```
skyqrose: The test name is "returns the appropriate headers when an OPTIONS request is made" but the only thing that's being asserted is the status. Did you mean to check the response headers, too?

apps/api_web/lib/api_web/plugs/cors.ex
```diff
@@ -7,22 +7,36 @@ defmodule ApiWeb.Plugs.CORS do
...
+      cond do
+        preflight_req?(conn) ->
+          send_preflight_resp(conn, opts)
 
-    if origin_allowed(conn, allowed_domains) do
-      put_cors_simple_resp_headers(
-        conn,
-        origins: allowed_domains
-      )
+        allowed_origin?(conn, opts) ->
```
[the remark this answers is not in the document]
paulswartz: It's undocumented, but public: https://github.com/whatyouhide/corsica/blob/master/lib/corsica.ex#L645-L654

apps/api_web/test/api_web/router_test.exs
paulswartz: Fixed.

PR #3, opened 2019-02-04
API version 2019-02-12

paulswartz:
Introduces support for new platforms at:

- Alewife
- Braintree
- Forest Hills
- Government Center
- Kenmore
- Oak Grove
- Park Street


apps/api_web/lib/api_web/controllers/prediction_controller.ex
```diff
@@ -122,11 +122,63 @@ defmodule ApiWeb.PredictionController do
...
+            "Braintree-02"
+          ]
+
+        "70036" ->
+          ["70036", "Oak Grove-01", "Oak Grove-02"]
+
+        "70001" ->
+          ["70001", "Forest Hills-01", "Forest Hills-02"]
+
+        "70199" ->
```
dmaltzan: Are you sure these IDs are right? 70199 is the E branch berth I think, which wasn't involved in these changes

paulswartz: You're right: 70200 is the current Eastbound stop, so that's what I should be looking for.

paulswartz: The Kenmore ones are correct:
- 70150 is Kenmore - Inbound, and 71150 is Kenmore - Inbound - B Branch
- 70151 is Kenmore - Outbound, and 71151 is Kenmore - Outbound - B Branch

apps/api_web/test/api_web/controllers/prediction_controller_test.exs
```diff
@@ -141,6 +141,38 @@ defmodule ApiWeb.PredictionControllerTest do
     end
   end
 
+  test "versions before 2019-02-12 include Alewife platformed stops", %{conn: conn} do
```
dmaltzan: I'm confused about which versions include or don't include which stops - could you describe that? I thought earlier versions would be the ones that _didn't_ have the new platforms

paulswartz: None of the versions make changes to the actual data. All they do is treat queries for old stop IDs as more expansive.

| Version | Stops expanded |
| -- | -- |
| 2019-02-12 | none |
| 2018-07-23 | Park St, Kenmore, Government Center, Alewife, Braintree, Forest Hills, Oak Grove |
| 2018-05-07 (and previous) | 2018-07-23 stops, plus North Station, South Station, Back Bay |

dmaltzan: so if an old-API user queries for a list of stops or something, they'll still get the new stop IDs?

paulswartz: Yep; we didn't make that change even for the CR platform change and it didn't seem to break anything for people.

dmaltzan: OK cool; that seems good then. We'll have to make a post at some point about the new Green Line stop IDs too.

PR #4, opened 2019-02-08
add calendar_attributes to services

skyqrose:
Ticket: [add calendar attributes to service API](https://app.asana.com/0/810933294009540/919723069714588)


apps/api_web/lib/api_web/controllers/service_controller.ex
```diff
@@ -99,6 +99,50 @@ defmodule ApiWeb.ServiceController do
...
+            )
+
+            service_schedule_typicality(
+              :integer,
+              """
+              Describes how well this schedule represents typical service for the listed `service_schedule_type`
+
+              | Value | Description                                                                 |
+              |-------|-----------------------------------------------------------------------------|
+              | `0`   | (or null) Not defined.                                                      |
```
paulswartz: In the API, should never be `null`: we should convert it to `0` at parsing.

apps/api_web/lib/api_web/views/service_view.ex
```diff
@@ -7,6 +7,10 @@ defmodule ApiWeb.ServiceView do
     :start_date,
     :end_date,
     :valid_days,
+    :service_description,
+    :service_schedule_name,
+    :service_schedule_type,
+    :service_schedule_typicality,
```
paulswartz: These are already part of the `service` endpoint, so the `service_` prefix is unnecessary.

apps/parse/lib/parse/calendar_attributes.ex
```diff
@@ -0,0 +1,30 @@
+defmodule Parse.CalendarAttributes do
+  @moduledoc """
+  Parser for GTFS calendar_attributes.txt
+  """
+  @behaviour Parse
```
paulswartz: You can `use Parse.Simple` instead, and then you only need to implement `parse_row/1`.

apps/parse/lib/parse/calendar_attributes.ex
```diff
@@ -0,0 +1,30 @@
...
+  defstruct [
```
paulswartz: Structs should be defined in the `models` app.

apps/state/lib/state/service.ex
```diff
@@ -54,10 +59,15 @@ defmodule State.Service do
...
+  def handle_new_state({
+        calendar,
+        calendar_attributes,
+        calendar_dates
+      }) do
     service_ids =
       [
         calendar |> Enum.map(&Map.get(&1, :service_id)),
+        calendar_attributes |> Enum.map(&Map.get(&1, :service_id)),
         calendar_dates |> Enum.map(&Map.get(&1, :service_id))
```
paulswartz: I realize you're matching the existing style here, but these shouldn't be using pipes as there's only a single one. `Enum.map(calendar_attributes, & &1.service_id)`

apps/parse/lib/parse/calendar_attributes.ex
```diff
@@ -0,0 +1,30 @@
...
+    |> Enum.map(&parse_row/1)
+  end
+
+  defp parse_row(row) do
+    %__MODULE__{
+      service_id: :binary.copy(row["service_id"]),
+      service_description: :binary.copy(row["service_description"]),
+      service_schedule_name: :binary.copy(row["service_schedule_name"]),
+      service_schedule_type: :binary.copy(row["service_schedule_type"]),
+      service_schedule_typicality: :binary.copy(row["service_schedule_typicality"])
```
paulswartz: This should be an integer, not a binary.

apps/api_web/lib/api_web/controllers/service_controller.ex
skyqrose: If there is no calendar_attributes entry for a service, so the other fields are all null, should we still fill this with 0?

paulswartz: Yep; it makes it much easier for clients.

apps/parse/lib/parse/calendar_attributes.ex
skyqrose: The `calendar_attributes` is only used in one place: it gets passed to `state/service.ex`, which uses it to construct the `service` that it stores, but it does not directly store the `calendar_attributes` in the state. Including the struct here was following the pattern of `calendar_dates`. Should there still be a `model/calendar_attributes`?

paulswartz: Fair enough.

apps/parse/lib/parse/calendar_attributes.ex
```diff
@@ -0,0 +1,25 @@
...
+  ]
+
+  def parse_row(row) do
+    %__MODULE__{
+      service_id: :binary.copy(row["service_id"]),
+      description: :binary.copy(row["service_description"]),
+      schedule_name: :binary.copy(row["service_schedule_name"]),
+      schedule_type: :binary.copy(row["service_schedule_type"]),
+      schedule_typicality: row["service_schedule_typicality"]
+    }
```
paulswartz: These will always end up as binaries: you probably need to do some conversions (empty binary -> nil, binary -> integer for typicality).

skyqrose:
@paulswartz this is ready for another review

apps/state/lib/state/service.ex
```diff
@@ -75,11 +85,17 @@ defmodule State.Service do
...
+
       %Service{
         id: service_id,
         start_date: start_date,
         end_date: end_date,
         valid_days: valid_days,
+        description: attributes && attributes.description,
+        schedule_name: attributes && attributes.schedule_name,
+        schedule_type: attributes && attributes.schedule_type,
+        schedule_typicality: (attributes && attributes.schedule_typicality) || 0,
```
paulswartz: You could simplify this code by assigning `attributes` as an empty `%Parse.CalendarAttributes{}` in the case where the ID isn't found?

apps/parse/lib/parse/calendar_attributes.ex
paulswartz: Probably a test of `parse_row/1` would catch this.

apps/parse/test/parse/calendar_attributes_test.exs
```diff
@@ -0,0 +1,22 @@
+defmodule Parse.CalendarAttributesTest do
+  @moduledoc false
+  use ExUnit.Case, async: true
+  alias Parse.CalendarAttributes
+
+  test "parses calendar entries" do
+    blob = """
+    service_id,service_description,service_schedule_name,service_schedule_type,service_schedule_typicality
+    BUS12019-hba19ns1-Weekday-02,Weekday schedule (no school),Weekday (no school),Weekday,1
```
paulswartz: What happens when typicality is the empty string? What about when the others are the empty string?

apps/parse/lib/parse/calendar_attributes.ex
```diff
@@ -0,0 +1,40 @@
...
+          schedule_type: String.t() | nil,
+          schedule_typicality: integer() | nil
+        }
+
+  defstruct [
+    :service_id,
+    :description,
+    :schedule_name,
+    :schedule_type,
+    :schedule_typicality
```
paulswartz: You can avoid the `|| 0` below by making 0 the default here, and the `copy_int` default as well.

skyqrose: In the current way there's only one place in charge of creating the default.

paulswartz: That's true, but it's not near the parsing code at all. In the future, you can do something like this:
```elixir
@default_typicality 0

defstruct [typicality: @default_typicality]

def copy_int(""), do: @default_typicality
def copy_int(binary), do: String.to_integer(binary)
```

PR #6, opened 2019-02-11
make the log level configurable

paulswartz:
`info` is the default level so we don't need a configuration for that.


paulswartz:
Good idea! I documented them in the README.

paulswartz:
Fixed.

PR #7, opened 2019-02-13
work on speeding up prediction/vehicle ingestion

paulswartz:
- remove `track` processing for non-legacy API versions
- fetch vehicle positions more frequently


paulswartz:
Superseded by #64 .

PR #8, opened 2019-02-14
parse status from Protobuf

paulswartz:


paulswartz:
Didn't get traction from the GTFS-RT community.

PR #11, opened 2019-02-26
Return an error code if someone uses a wrong filter

lboyarsky:
Ticket: [Return an error code if someone uses a wrong filter](https://app.asana.com/0/810933294009540/1104657015105420)

PR #18, opened 2019-03-06
revert validation of parameters

paulswartz:
Partially reverts #11 

This leaves the function calls intact so that they can be re-implemented with
the future version-based approach.


paulswartz:
It ended up breaking more clients than expected, so I re-deployed an earlier version last night. I filed a followup ticket to create a new version to use this validation and so I didn't want to fully revert the change and possibly cause merge issues down the road.

PR #20, opened 2019-03-15
StopsOnRoute fixes

paulswartz:
tickets:
- [Plimptonville out of order](https://app.asana.com/0/116562505129567/1112453530163240)
- [101 outbound does not include all stops](https://app.asana.com/0/116562505129567/1112469349487825)

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -105,4 +105,12 @@ config :state, :shape,
     "-S" => -1
   }
 
+# Overrides for the stop ordering on routes where the trips themselves aren't enough
+config :state, :stops_on_route,
+  stop_order_overrides: %{
+    {"CR-Franklin", 0} => [
+      ["Norwood Central", "Windsor Gardens", "Plimptonville", "Walpole"]
+    ]
+  }
+
 import_config "#{Mix.env()}.exs"
--- a/apps/state/lib/state/stops_on_route.ex
+++ b/apps/state/lib/state/stops_on_route.ex
@@ -12,6 +12,9 @@ defmodule State.StopsOnRoute do
 
   @table __MODULE__
 
+  # ensure we recompile this module when the state config changes
+  @external_resource "apps/state/config/config.exs"
+
   @type stop_id_list :: [Model.Stop.id()]
 
   @typep record ::
@@ -72,7 +75,6 @@ defmodule State.StopsOnRoute do
     subscriptions = [
       {:new_state, State.Route},
       {:new_state, State.Trip},
-      {:new_state, State.Shape},
       {:new_state, State.Schedule}
     ]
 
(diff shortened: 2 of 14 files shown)


apps/state/lib/state/stops_on_route.ex
```diff
@@ -161,11 +165,15 @@ defmodule State.StopsOnRoute do
...
-    |> Stream.reject(&ignore_trip_for_route?/1)
-    |> stop_ids_for_trips
-    |> merge_ids
+  defp order_stop_ids_for_trips(route, direction_id, trips) do
+    trip_stops =
+      trips
+      |> Stream.reject(&ignore_trip_for_route?/1)
+      |> stop_ids_for_trips()
+
+    overrides = stop_order_overrides(route.id, direction_id)
```
skyqrose: Could you reference `Application.get_env(:state, :stops_on_route)[:stop_order_overrides]` here directly?
It'd save some complexity with the macro definitions and the `@external_resource` above.

apps/state/config/config.exs
```diff
@@ -105,4 +105,15 @@ config :state, :shape,
...
 
+# Overrides for the stop ordering on routes where the trips themselves aren't enough
+config :state, :stops_on_route,
+  stop_order_overrides: %{
+    {"CR-Kingston", 0} => [
+      ["Hanson", "Halifax", "Kingston", "Plymouth"]
+    ],
+    {"CR-Franklin", 0} => [
+      ["Norwood Central", "Windsor Gardens", "Plimptonville", "Walpole"]
+    ]
```
skyqrose: These don't need to be lists. We probably won't need more than one override per (route,direction), and even if we do, I think one longer override would probably work as well as two smaller ones.

apps/state/lib/state/stops_on_route.ex
```diff
@@ -207,11 +223,18 @@ defmodule State.StopsOnRoute do
...
+  @spec merge_ids([stop_id_list], [stop_id_list]) :: stop_id_list
+  def merge_ids(lists_of_ids, override_lists \\ [])
+
+  def merge_ids([], _), do: []
+
+  def merge_ids(lists_of_ids, override_lists) do
+    sorted_lists = Enum.sort_by(lists_of_ids, &list_merge_key/1, &>=/2)
+    # overrides should be short or empty, so putting that first with ++ is
+    # fine.
+    lists_with_overrides = override_lists ++ sorted_lists
```
skyqrose: if we limit each (route,direction) to one override, as discussed above, this gets simpler.

apps/state/test/state/stops_on_route_test.exs
```diff
@@ -363,6 +429,18 @@ defmodule State.StopsOnRouteTest do
       # know that there are only three items.
       assert [_, _, _] = merge_ids(stop_ids)
     end
+
+    test "can use an override to order individual stops" do
+      stop_ids =
+        Enum.shuffle([
```
skyqrose: This will make the tests only fail intermittently if there's a problem.
Perhaps you should always do a merge in both directions, and check that both work or just do the one order that'd be broken now.

apps/state/config/config.exs
```diff
@@ -105,4 +105,15 @@ config :state, :shape,
     "-S" => -1
   }
 
+# Overrides for the stop ordering on routes where the trips themselves aren't enough
+config :state, :stops_on_route,
+  stop_order_overrides: %{
+    {"CR-Kingston", 0} => [
+      ["Hanson", "Halifax", "Kingston", "Plymouth"]
```
skyqrose: Is this needed?

paulswartz:
Closing for now in favor of #21. Will re-open with separate changes for the 101.

PR #21, opened 2019-03-18
allow passing overrides to the merge function

paulswartz:
This lets us give a short chunk of ordering preference.

Fixes [Plimptonville out of order](https://app.asana.com/0/116562505129567/1112453530163240)

### DIFF
--- a/apps/state/lib/state/stops_on_route.ex
+++ b/apps/state/lib/state/stops_on_route.ex
@@ -115,14 +115,15 @@ defmodule State.StopsOnRoute do
     route.id
     |> State.Trip.by_route_id()
     |> Enum.group_by(& &1.direction_id)
-    |> Enum.flat_map(fn {_direction_id, trips} ->
-      do_gather_route_direction(route, trips)
+    |> Enum.flat_map(fn {direction_id, trips} ->
+      do_gather_route_direction(route, direction_id, trips)
     end)
   end
 
-  @spec do_gather_route_direction(Model.Route.t(), [Model.Trip.t()]) :: [record]
-  defp do_gather_route_direction(route, trips) do
-    global_stop_id_order = order_stop_ids_for_trips(trips)
+  @spec do_gather_route_direction(Model.Route.t(), Model.Direction.id(), [Model.Trip.t()]) ::
+          [record]
+  defp do_gather_route_direction(route, direction_id, trips) do
+    global_stop_id_order = order_stop_ids_for_trips(route, direction_id, trips)
 
     trips
     |> Enum.group_by(fn trip ->
@@ -137,7 +138,7 @@ defmodule State.StopsOnRoute do
 
     stop_ids =
       trip_group
-      |> stop_ids_for_trips
+      |> stop_ids_for_trips()
       |> merge_group_stop_ids(global_order)
 
     if stop_ids == [] do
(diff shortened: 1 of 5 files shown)


apps/state/lib/state/stops_on_route.ex
```diff
@@ -187,6 +192,19 @@ defmodule State.StopsOnRoute do
     Enum.filter(global_order, &MapSet.member?(group_stop_id_map, &1))
   end
 
+  # additional ordered stop IDs to include during the merge
+  defp stop_order_overrides(route_id, direction_id) do
+    config = Application.get_env(:state, :stops_on_route)[:stop_order_overrides]
```
skyqrose: Thoughts on putting this in code instead of config? It'd remove the need to fetch the config, the `@external_resource`, and make it easier to understand the code path.

apps/state/test/state/stops_on_route_test.exs
```diff
@@ -363,6 +416,18 @@ defmodule State.StopsOnRouteTest do
       # know that there are only three items.
       assert [_, _, _] = merge_ids(stop_ids)
     end
+
+    test "can use an override to order individual stops" do
+      stop_ids =
+        Enum.shuffle([
```
skyqrose: This will make the tests only fail intermittently if there's a problem.
Perhaps you should always do a merge in both directions, and check that both work or just do the one order that'd be broken now.

paulswartz: This seems harder to understand:
```elixir
      overrides = [[2, 4, 3, 5]]

      for stop_ids <- [
            [
              [1, 2, 3, 5, 6],
              [1, 2, 4, 5, 6]
            ],
            [
              [1, 2, 4, 5, 6],
              [1, 2, 3, 5, 6]
            ],
          ] do
          assert [1, 2, 4, 3, 5, 6] == merge_ids(stop_ids, overrides)
      end
    end
```

Since `Enum.shuffle` is dependent on the seed, a failure that's caused by a change for this is easily reproducible. We already have some property tests which also depend on randomness, so it's already a possibility we live with.

apps/state/lib/state/stops_on_route.ex
paulswartz: - The `@external_resource` was already removed, as now this happens at run time.
- Having it in configuration (like the shape overrides) limits the possible changes and makes the code easier to understand.

paulswartz:
The only new function is `stop_order_overrides` which is private and doesn't require a `@spec`.

apps/state/test/state/stops_on_route_test.exs
skyqrose: Having some intermittently failing tests doesn't make a new one any less painful. How about this?
```ex
overrides = [[2, 3, 4, 5]]
# try both permutations so we know the result order is due to the override and not the input order.
assert [1, 2, 3, 4, 5, 6] == merge_ids([
  [1, 2, 3, 5, 6],
  [1, 2, 4, 5, 6]
], overrides)
assert [1, 2, 3, 4, 5, 6] == merge_ids([
  [1, 2, 4, 5, 6],
  [1, 2, 3, 5, 6]
], overrides)
```
Even if it is more verbose, that's less bad than intermittent test failures.

skyqrose:
I guess `order_stop_ids_for_trips` isn't new, but could still use a `@spec`

`@spec`s are useful on private functions too.

paulswartz:
They're useful if they provide additional information, but for these:
```elixir
@spec order_stop_ids_for_trips(Model.Route.t(), Model.Direction.id(), [Model.Trip.t()]) :: stop_id_list
@spec stop_order_overrides(Model.Route.id(), Model.Direction.id()) :: stop_id_list
```
they're basically duplicates of the names of the arguments. I did add them, though.

apps/state/test/state/stops_on_route_test.exs
paulswartz: This is hard to understand, since it's not obvious what is the difference between the two lists when you have them twice. You can tell that it's harder to understand by the comment that's almost as long as the test :)

We already use randomness to find issues with the code (property tests, test suite order randomization, sorting tests), and as long as the failure is reproducible with the seed every time, it's not really an intermittent failure in the same way that say a race condition (affected by timing outside the control of the test) would be. Like many things in programming, it's a trade-off, and it's a trade-off that the API uses in other places to balance the amount of test code we need to write and the space covered by our tests.

apps/state/lib/state/stops_on_route.ex
```diff
@@ -187,6 +194,20 @@ defmodule State.StopsOnRoute do
     Enum.filter(global_order, &MapSet.member?(group_stop_id_map, &1))
   end
 
+  # additional ordered stop IDs to include during the merge
+  @spec stop_order_overrides(Model.Route.id(), Model.Direction.id()) :: stop_id_list
```
skyqrose: This should return `:: [stop_id_list]`. It's not obvious just from the name.

paulswartz: Also not obvious to Dialyzer since it didn't find any issues :) Fixed.

PR #24, opened 2019-03-20
use Route Patterns to decide what stops are on routes

paulswartz:
The [101](https://www.mbta.com/schedules/101/line?direction_id=0) has a few different variants (not all represented as shapes), and it ends up with a confusing behavior when looking at what stops the API considers to be a part of that route during the week.

This change is to stop using the shapes to determine the stops on a day, and instead to use the route patterns. This still allows us to ignore shuttles and other disruptions, but does include atypical patterns which received negative shape priorities in the past.

We keep the existing shape code around, as it's used to provide the list of stops for each shape.

Product/design sign-off: 
- [x] @liana 
- [x] @ingridpierre

### DIFF
--- a/apps/state/lib/state/stops_on_route.ex
+++ b/apps/state/lib/state/stops_on_route.ex
@@ -38,7 +38,7 @@ defmodule State.StopsOnRoute do
 
     matchers =
       for service_id <- Keyword.get(opts, :service_ids, [:_]),
-          shape_id <- Keyword.get(opts, :shape_ids, [:_]),
+          shape_id <- Keyword.get(opts, :shape_ids, [:all]),
           route_id <- route_ids do
         {{route_id, direction_id, shape_id, service_id, alternate?, :"$1"}, [], [:"$1"]}
       end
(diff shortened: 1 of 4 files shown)


apps/state/lib/state/stops_on_route.ex
```diff
@@ -125,12 +125,43 @@ defmodule State.StopsOnRoute do
...
+      trips
+      |> Enum.group_by(fn trip ->
+        {ignore_trip_route_pattern?(trip), :all, trip.service_id, trip.direction_id}
+      end)
+      |> Enum.flat_map(&do_gather_direction_group(route, global_stop_id_order, &1))
+
+    Enum.concat(shape_records, other_records)
+  end
+
+  defp do_gather_direction_group(
```
lboyarsky: This looks pretty much identical to the [existing function](https://github.com/mbta/api/blob/340dbcdba656c33c91e3d0c26c7a8f6744074f52/apps/state/lib/state/stops_on_route.ex#L167-L182)? The only difference seems to be that new function uses `:all` atom as shape id, but I don't see why we cannot just pass it as the argument to the existing function. Or am I missing something?

paulswartz: This version uses `ignore_trip_route_pattern?` rather than `ignore_trip_for_route?`. The new function uses the route patterns to decide which trips to use, rather than the shapes.

lboyarsky: I understand that new version of `do_gather_route_direction` is different from the old one; what I'm talking about is that you now have two almost identical versions of `do_gather_direction_group` -- this one, and the one on lines 167-182. These functions don't use `ignore_trip_route_pattern?` or `ignore_trip_for_route?` internally.

paulswartz: Ah, I see; thanks!

paulswartz: fixed!

PR #25, opened 2019-03-27
allow new parent station IDs for Commuter Rail stations

paulswartz:


skyqrose:
Can we revert this / switch to just the new ids once the change to gtfs is merged? If so, is there an asana task or a note on an existing task to do that, so that we don't forget that we can clean this up soon?

PR #30, opened 2019-04-03
allow querying trip by name

handorff:


apps/api_web/test/api_web/controllers/trip_controller_test.exs
```diff
@@ -133,6 +133,17 @@ defmodule ApiWeb.TripControllerTest do
       assert index_data(conn, %{"route" => "2", "direction_id" => "0"}) == []
     end
 
+    test "filters by name", %{conn: conn} do
+      trip = %Model.Trip{id: "1", name: "name"}
+      :ok = State.Trip.new_state([trip])
+
+      [^trip] = State.Trip.filter_by(%{names: ["name"]})
+      [] = State.Trip.filter_by(%{names: ["not_a_name"]})
```
paulswartz: these should be tests in `apps/state`

apps/api_web/test/api_web/controllers/trip_controller_test.exs
```diff
@@ -133,6 +133,17 @@ defmodule ApiWeb.TripControllerTest do
...
+
+      assert index_data(conn, %{"name" => "name"}) == [trip]
+      assert index_data(conn, %{"name" => "not_a_name"}) == []
```
paulswartz: there should also be a test that an empty name returns an error.

paulswartz: I think `index_data(conn, %{"name" => ","})` will return trips, instead of an error.

PR #35, opened 2019-04-08
Allow querying stops by location type

handorff:
Also fixes a bug where, in some cases, we returned the union of different filters rather than the intersection when filtering stops.


apps/state/lib/state/stop.ex
```diff
@@ -215,7 +230,13 @@ defmodule State.Stop do
         end
       end)
 
-    Enum.uniq_by(stops, & &1.id)
+    search_results
+    |> Enum.to_list()
+    |> Enum.reduce(all(), fn results, acc ->
```
paulswartz: 1. You don't need `Enum.to_list`, as `Enum.reduce` can work against anything enumerable (including streams).
1. This will always ask for all the stops, even when there are filters.

PR #36, opened 2019-04-08
Clarify /stops documentation re: including route

handorff:


apps/api_web/lib/api_web/controllers/stop_controller.ex
```diff
@@ -174,7 +179,11 @@ defmodule ApiWeb.StopController do
     #{swagger_path_description("/data")}
     """)
 
-    include_parameters(@includes)
+    include_parameters(@includes,
```
paulswartz: For `show` it's not possible to include a route: we shouldn't include `route` as one of the parameters here.

apps/api_web/lib/api_web/controllers/stop_controller.ex
```diff
@@ -179,10 +179,8 @@ defmodule ApiWeb.StopController do
     #{swagger_path_description("/data")}
     """)
 
-    include_parameters(@includes,
-      description:
-        "Note that `route` can only be included if `filter[route]` is present and has exactly one `/data/{index}/relationships/route/data/id`. Use [/stops](#/Stop/ApiWeb_StopController_index) with `filter[id]` and `filter[route]` to include `route` with a specific stop."
-    )
+    # For show, it's not possible to include a route, so exclude it from the list of include parameters.
+    include_parameters(@includes -- ["route"])
```
paulswartz: You probably want a separate module variable here. That way, you can use it in `show_data/2` when the includes get validated.

PR #37, opened 2019-04-09
Add recommended transfers relationship to stops

handorff:


apps/state/lib/state/transfer.ex
```diff
@@ -0,0 +1,25 @@
...
+  use State.Server,
+    fetched_filename: "transfers.txt",
+    recordable: Model.Transfer,
+    indicies: [:from_stop_id],
+    parser: Parse.Transfers
+
+  def recommended_transfers_from(stop_id) when is_binary(stop_id) do
+    stop_id
+    |> by_from_stop_id()
+    |> Enum.filter(&(&1.transfer_type == 0))
```
paulswartz: You should be able to combine this:
```
%{from_stop_id: stop_id, transfer_type: 0}
|> select(:from_stop_id)
|> Enum.map(& &1.to_stop_id)
```

apps/state/lib/state/transfer.ex
```diff
@@ -0,0 +1,25 @@
...
+    |> Enum.map(& &1.to_stop_id)
+    |> State.Stop.by_ids()
+  end
+
+  def recommended_transfers_from(nil) do
```
paulswartz: When does `nil` come up? All stops should have an ID.

apps/parse/lib/parse/transfers.ex
```diff
@@ -0,0 +1,37 @@
...
+  @doc """
+  Parses (non-header) row of `transfers.txt`
+
+  See `Model.Transfer.t` documentation for descriptions of the individual fields.
+  """
+  def parse_row(row) do
+    %Model.Transfer{
+      from_stop_id: copy(row["from_stop_id"]),
+      to_stop_id: copy(row["to_stop_id"]),
+      transfer_type: String.to_integer(row["transfer_type"]),
```
paulswartz: This works for our feed, but per the documentation it's also possible for this to be empty. That would be the same as a 0.

PR #39, opened 2019-04-19
Add rate limit headers

handorff:
* Adds new headers "x-ratelimit-limit", "x-ratelimit-remaining", "x-ratelimit-reset".
* Changes rate limiting to use time-based buckets rather than memcached ttl, so that we know the time when limits reset.

### DIFF
--- a/apps/api_web/lib/api_web/plugs/rate_limiter.ex
+++ b/apps/api_web/lib/api_web/plugs/rate_limiter.ex
@@ -11,15 +11,32 @@ defmodule ApiWeb.Plugs.RateLimiter do
 
   def call(%{assigns: assigns, request_path: request_path} = conn, _) do
     case ApiWeb.RateLimiter.log_request(assigns.api_user, request_path) do
-      :ok ->
+      %{status: :ok, limit_data: %{limit: limit, remaining: remaining, reset_ms: reset_ms}} ->
+        conn
+        |> put_rate_limit_headers(limit, remaining, reset_ms)
+
+      %{status: :ok} ->
         conn
 
-      {:error, :rate_limited} ->
+      %{
+        status: {:error, :rate_limited},
+        limit_data: %{limit: limit, remaining: remaining, reset_ms: reset_ms}
+      } ->
         conn
+        |> put_rate_limit_headers(limit, remaining, reset_ms)
         |> put_status(429)
         |> put_view(ApiWeb.ErrorView)
         |> render("429.json-api", [])
         |> halt()
     end
+  end
+
+  defp put_rate_limit_headers(conn, limit, remaining, reset_ms) do
+    reset_seconds = div(reset_ms, 1_000)
+
+    conn
+    |> put_resp_header("x-ratelimit-limit", "#{limit}")
+    |> put_resp_header("x-ratelimit-remaining", "#{remaining}")
+    |> put_resp_header("x-ratelimit-reset", "#{reset_seconds}")
   end
 end
--- a/apps/api_web/lib/api_web/plugs/rate_limiter.ex
+++ b/apps/api_web/lib/api_web/plugs/rate_limiter.ex
@@ -11,17 +11,14 @@ defmodule ApiWeb.Plugs.RateLimiter do
 
   def call(%{assigns: assigns, request_path: request_path} = conn, _) do
     case ApiWeb.RateLimiter.log_request(assigns.api_user, request_path) do
-      %{status: :ok, limit_data: %{limit: limit, remaining: remaining, reset_ms: reset_ms}} ->
+      :ok ->
+        conn
+
+      {:ok, {limit, remaining, reset_ms}} ->
         conn
         |> put_rate_limit_headers(limit, remaining, reset_ms)
 
-      %{status: :ok} ->
-        conn
-
-      %{
-        status: {:error, :rate_limited},
-        limit_data: %{limit: limit, remaining: remaining, reset_ms: reset_ms}
-      } ->
+      {:rate_limited, {limit, remaining, reset_ms}} ->
         conn
         |> put_rate_limit_headers(limit, remaining, reset_ms)
         |> put_status(429)
--- a/apps/api_web/lib/api_web/rate_limiter.ex
+++ b/apps/api_web/lib/api_web/rate_limiter.ex
@@ -12,6 +12,17 @@ defmodule ApiWeb.RateLimiter do
   @max_anon_per_interval ApiWeb.config(:rate_limiter, :max_anon_per_interval)
   @max_registered_per_interval ApiWeb.config(:rate_limiter, :max_registered_per_interval)
 
+  @type request_status :: :ok | {:error, :rate_limited}
+  @type limit_data :: %{
+          :limit => non_neg_integer,
+          :remaining => non_neg_integer,
+          :reset_ms => non_neg_integer
+        }
+  @type log_result :: %{
+          :status => request_status,
+          optional(:limit_data) => limit_data
+        }
+
   ## Client
 
   def start_link(_opts \\ []) do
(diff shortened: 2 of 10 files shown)


apps/api_web/lib/api_web/plugs/rate_limiter.ex
```diff
@@ -11,15 +11,32 @@ defmodule ApiWeb.Plugs.RateLimiter do
 
   def call(%{assigns: assigns, request_path: request_path} = conn, _) do
     case ApiWeb.RateLimiter.log_request(assigns.api_user, request_path) do
-      :ok ->
+      %{status: :ok, limit_data: %{limit: limit, remaining: remaining, reset_ms: reset_ms}} ->
```
paulswartz: You don't really need a map for this: `{:ok | :rate_limited, limit_data} | :ok`. `limit_data` itself could be a tuple too, `{limit, remaining, reset_ms}`

apps/api_web/test/api_web/rate_limiter/memcache_test.exs
```diff
@@ -8,8 +8,26 @@ defmodule ApiWeb.RateLimiter.MemcacheTest do
     test "returns true if there is no rate limit remaining" do
       {:ok, _} = start_link(clear_interval: 1000)
       user_id = "#{System.monotonic_time()}"
-      refute rate_limited?(user_id, 1)
-      assert rate_limited?(user_id, 1)
+
+      {has_limit, _} = rate_limited?(user_id, 1)
+      refute has_limit
```
paulswartz:
```elixir
assert {false, _} = rate_limited(user_id, 1)
assert {true, _} = rate_limited(user_id, 1)
```

apps/api_web/test/api_web/rate_limiter/memcache_test.exs
```diff
@@ -8,8 +8,26 @@ defmodule ApiWeb.RateLimiter.MemcacheTest do
...
+      {has_limit, _} = rate_limited?(user_id, 1)
+      assert has_limit
+    end
+
+    test "returns the correct number of requests" do
+      {:ok, _} = start_link(clear_interval: 1000)
+      user_id = "#{System.monotonic_time()}"
+
+      {_, requests} = rate_limited?(user_id, 2)
+      assert requests == 1
```
paulswartz:
```elixir
assert {false, 1} = rate_limited?(user_id, 2)
# &c
```

apps/api_web/lib/api_web/plugs/rate_limiter.ex
```diff
@@ -11,15 +11,32 @@ defmodule ApiWeb.Plugs.RateLimiter do
...
     end
   end
+
+  defp put_rate_limit_headers(conn, limit, remaining, reset_ms) do
+    reset_seconds = div(reset_ms, 1_000)
+
+    conn
+    |> put_resp_header("x-ratelimit-limit", "#{limit}")
+    |> put_resp_header("x-ratelimit-remaining", "#{remaining}")
+    |> put_resp_header("x-ratelimit-reset", "#{reset_seconds}")
```
paulswartz: I don't think there's a test for this?

apps/api_web/lib/api_web/rate_limiter.ex
```diff
@@ -35,19 +46,34 @@ defmodule ApiWeb.RateLimiter do
...
+  @spec log_request(any, String.t()) :: log_result
+  def log_request(_, "/_health" <> _), do: %{status: :ok}
 
   def log_request(user, _request_path) do
     max = max_requests(user)
-
-    if @limiter.rate_limited?(user.id, max) do
-      {:error, :rate_limited}
+    {key, reset_ms} = key_and_reset_time(user)
+    {rate_limited, requests} = @limiter.rate_limited?(key, max)
```
paulswartz: Would it be easier to have this function return the number of requests remaining? It seems like that's what the Memcache version is returning already.

apps/api_web/lib/api_web/rate_limiter/limiter.ex
```diff
@@ -3,14 +3,14 @@ defmodule ApiWeb.RateLimiter.Limiter do
...
-  - `rate_limited?(user_id, max_requests)` returns true if the user_id has used too many requests.
+  - `rate_limited?(user_id, max_requests)` returns true if the user_id has used too many requests, along with the number of requests the user_id has used in this time period.
 
   The main option passed to `start_link/1` is `clear_interval` which is a
   number of milliseconds to bucket the requests into.
 
   """
   @callback start_link(Keyword.t()) :: {:ok, pid}
-  @callback rate_limited?(String.t(), non_neg_integer) :: boolean
+  @callback rate_limited?(String.t(), non_neg_integer) :: {boolean, non_neg_integer}
```
paulswartz: What about a return_type of `{:remaining, non_neg_integer} | :rate_limited`? It doesn't make sense for there to be a number of requests remaining if you've been limited, so that data is irrelevant.

apps/api_web/lib/api_web/plugs/rate_limiter.ex
paulswartz: Also, https://opensource.zalando.com/restful-api-guidelines/#153 says reset is a relative number of seconds: do we want to use that instead?

handorff: I don't have a strong sense of whether developers would prefer to be told a relative number of seconds or a timestamp for this. (Do you?)

I went with timestamp over relative because that's what Twitter, Github, Vimeo and Twitch all do; I didn't come across any examples of APIs from well-known companies that include a response header similar to `X-RateLimit-Reset` where the value is relative.

handorff:
I didn't see a natural place in the swagger docs -- they don't appear to address rate limiting at all, and `api_key` isn't listed as a parameter anywhere.

paulswartz:
> I didn't see a natural place in the swagger docs -- they don't appear to address rate limiting at all, and `api_key` isn't listed as a parameter anywhere.

They're listed under "Authorize" on https://api-v3.mbta.com/docs/swagger/index.html

apps/api_web/lib/api_web/plugs/rate_limiter.ex
paulswartz: I think the rationale for having a relative timestamp is that it avoids the issue where different servers have a different time. However, I don't have a strong preference and other companies have standardized on an epoch time, so let's keep it this way.

apps/api_web/lib/api_web/rate_limiter/limiter.ex
```diff
@@ -3,14 +3,15 @@ defmodule ApiWeb.RateLimiter.Limiter do
   Behavior for backends to the V3 API rate limiter.
 
   - `start_link(opts)` is called to start the backend by the supervisor.
-  - `rate_limited?(user_id, max_requests)` returns true if the user_id has used too many requests.
+  - `rate_limited?(user_id, max_requests)` returns true if the user_id has used too many requests, along with the number of requests the user_id has used in this time period.
```
paulswartz: This isn't quite right anymore.

apps/api_web/lib/api_web/router.ex
```diff
@@ -345,6 +352,13 @@ defmodule ApiWeb.Router do
           ##### Header
           Without an api key as a request header or in the query string, requests will be tracked by IP address and have stricter rate limit. \
           [Register for a key](/register)
+
+          The HTTP headers returned in any API response show your rate limit status:
+          | Header | Description |
+          | ------ | ----------- |
+          | `x-ratelimit-limit` | The maximum number of requests you're allowed to make per time window. |
+          | `x-ratelimit-remaining` | The number of requests remaining in the current time window. |
+          | `x-ratelimit-reset` | The time at which the current rate limit time window ends in UTC epoch seconds. |
```
paulswartz: :+1: good place for this. Can we pull the text into a `@modulevar` so we don't need to duplicate it.

apps/api_web/test/api_web/plugs/rate_limiter_test.exs
```diff
@@ -58,4 +58,20 @@ defmodule ApiWeb.Plugs.RateLimiterTest do
...
+    setup %{conn: conn} do
+      ApiWeb.RateLimiter.force_clear()
+      {:ok, conn: conn}
+    end
+
+    test "have rate limiting headers in response", %{conn: conn} do
+      conn = get(conn, @url)
+      resp_headers = conn.resp_headers
+
+      assert {"x-ratelimit-limit", _} = List.keyfind(resp_headers, "x-ratelimit-limit", 0)
```
paulswartz: You can use [`Plug.Conn.get_resp_header/2`](https://hexdocs.pm/plug/Plug.Conn.html#get_resp_header/2) for this:
```elixir
assert [_] = get_resp_header(conn, "x-rate-limit-limit")
...
```

apps/api_web/test/api_web/rate_limiter/memcache_test.exs
```diff
@@ -8,8 +8,18 @@ defmodule ApiWeb.RateLimiter.MemcacheTest do
...
+      assert {:remaining, _} = rate_limited?(user_id, 1)
+      assert :rate_limited == rate_limited?(user_id, 1)
+    end
+
+    test "returns the correct number of requests" do
+      {:ok, _} = start_link(clear_interval: 1000)
+      user_id = "#{System.monotonic_time()}"
+
+      assert {:remaining, 1} = rate_limited?(user_id, 2)
+      assert {:remaining, 0} = rate_limited?(user_id, 2)
```
paulswartz: :+1:

apps/api_web/test/api_web/rate_limiter/memcache_test.exs
```diff
@@ -8,8 +8,18 @@ defmodule ApiWeb.RateLimiter.MemcacheTest do
     test "returns true if there is no rate limit remaining" do
       {:ok, _} = start_link(clear_interval: 1000)
       user_id = "#{System.monotonic_time()}"
-      refute rate_limited?(user_id, 1)
-      assert rate_limited?(user_id, 1)
+
+      assert {:remaining, _} = rate_limited?(user_id, 1)
```
paulswartz: This should be `{:remaining, 0}` yes?

PR #40, opened 2019-04-23
Test allowing streaming all predictions

handorff:


paulswartz:
Good test, but we learned that it's too slow to work in practice.

PR #41, opened 2019-04-23
Validate that there are no filters at /show endpoints

handorff:


apps/api_web/lib/api_web/params.ex
```diff
@@ -295,4 +295,33 @@ defmodule ApiWeb.Params do
...
+
+        _ ->
+          []
+      end
+
+    case bad_filters ++ bad_query_params do
+      [] ->
+        :ok
+
+      bad_params ->
```
paulswartz: It's probably more efficient to split this up slightly.
```elixir
def validate_show_params(params, conn)
def validate_show_params(params, %{assigns: %{api_version: version}}) when version >= "2019-04-05" do
  bad_query_params ...
  bad_filters ...
  case {bad_filters, bad_query_params} do
    {[], []} -> :ok 
    {a, b} -> {:error, :bad_filter, a ++ b}
  end    
end
def validate_show_params(_params, _conn) do
  :ok 
end

PR #42, opened 2019-04-26
Create zone relationship for /stops

handorff:


apps/api_web/lib/api_web/views/stop_view.ex
```diff
@@ -83,6 +89,10 @@ defmodule ApiWeb.StopView do
     State.Transfer.recommended_transfers_from(stop_id)
   end
 
+  def zone(%{id: stop_id}, _conn) do
+    State.Stop.zone_id(stop_id)
```
paulswartz: I think the first argument here is a `%Stop{}` already, so you could match it out:
```elixir
def zone(%{zone_id: zone_id}, _conn), do: zone_id
```

apps/parse/lib/parse/stops.ex
```diff
@@ -26,7 +26,8 @@ defmodule Parse.Stops do
       longitude: String.to_float(row["stop_lon"]),
       parent_station: copy_if_not_blank(row["parent_station"]),
       wheelchair_boarding: String.to_integer(row["wheelchair_boarding"]),
-      location_type: String.to_integer(row["location_type"])
+      location_type: String.to_integer(row["location_type"]),
+      zone_id: copy_if_not_blank(row["zone_id"])
```
paulswartz: This could also use a test.

apps/api_web/lib/api_web/views/stop_view.ex
```diff
@@ -23,6 +23,12 @@ defmodule ApiWeb.StopView do
     field: :parent_station
   )
 
+  has_one(
+    :zone,
+    type: :zone,
+    field: :zone
```
paulswartz: Actually, can you avoid the extra function definition if you make this `zone_id`?

PR #46, opened 2019-05-06
Fix including stop with route via filter[stop]

handorff:


apps/api_web/lib/api_web/views/route_view.ex
```diff
@@ -63,7 +63,19 @@ defmodule ApiWeb.RouteView do
     relationships = super(route, conn)
 
     if split_included?("stop", conn) do
-      stop = State.Stop.by_id(conn.params["stop"])
+      stop_id =
```
paulswartz: Is it possible to use the data from `Params.validate_filters` rather than partially re-implementing it here?

handorff: I don't see a great way to do this, though I could be missing something.

If we pass `conn.params` to `validate_filters/3`, it will give us an error for default params like `includes`.

If we instead use `filter_params/3`, we still need to know `@filters` from `RouteController`, which I don't think we can access from outside the module.

If we try to just filter out `stop` and ignore the other possible filters for `/route`, we'll get an error if any filters besides `stop` are present.

PR #47, opened 2019-05-06
Allow including the same resource twice

handorff:
Note that this PR will allow users to include exactly the same relationship path multiple times: e.g. https://api-v3.mbta.com/route-patterns/Red-1-1?include=representative_trip.service,representative_trip.service. (This would return the same response as if it were only included once.)

Note also that, currently, there is no validation of anything besides the first part of a relationship path. This is not changed by the current PR. e.g. https://api-v3.mbta.com/route-patterns/Red-1-1?include=representative_trip.bad_path will not error, and will return the same response as https://api-v3.mbta.com/route-patterns/Red-1-1?include=representative_trip. I believe it would require significant changes to change this: in the route pattern controller, we know what includes are permitted for route patterns, but don't have access to the list of allowed includes for trips.

### DIFF
--- a/apps/api_web/lib/api_web/params.ex
+++ b/apps/api_web/lib/api_web/params.ex
@@ -283,7 +283,8 @@ defmodule ApiWeb.Params do
           |> String.split(",", trim: true)
           |> Enum.map(&(&1 |> String.split(".") |> List.first()))
 
-        bad_includes = Enum.filter(split, fn el -> el not in includes end)
+        includes_set = MapSet.new(includes)
+        bad_includes = Enum.filter(split, fn el -> el not in includes_set end)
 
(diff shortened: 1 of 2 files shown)


apps/api_web/lib/api_web/params.ex
```diff
@@ -283,7 +283,7 @@ defmodule ApiWeb.Params do
           |> String.split(",", trim: true)
           |> Enum.map(&(&1 |> String.split(".") |> List.first()))
 
-        bad_includes = split -- includes
+        bad_includes = Enum.filter(split, fn el -> el not in includes end)
```
paulswartz: This is effectively quadratic, as it's iterating over two linked lists. Could you convert `includes` to a `MapSet`?

PR #50, opened 2019-05-07
Add fields[] to Swagger documentation

handorff:
Note that fields is not currently supported for `/route-patterns` and `/live-facilities`.

### DIFF
--- a/apps/api_web/lib/api_web/api_controller_helpers.ex
+++ b/apps/api_web/lib/api_web/api_controller_helpers.ex
@@ -171,7 +171,13 @@ defmodule ApiWeb.ApiControllerHelpers do
   end
 
   defp view_module_for_type(type) do
-    view_name = String.capitalize(type) <> "View"
+    view_name =
+      (type
+       |> String.split("_")
+       |> Enum.map(&String.capitalize/1)
+       |> Enum.join("")) <>
+        "View"
+
     Module.safe_concat([ApiWeb, view_name])
   end
 
--- a/apps/api_web/lib/api_web/api_controller_helpers.ex
+++ b/apps/api_web/lib/api_web/api_controller_helpers.ex
@@ -171,7 +171,7 @@ defmodule ApiWeb.ApiControllerHelpers do
   end
 
   defp view_module_for_type(type) do
-    view_name = Macro.camelize(type) <> "View"
+    view_name = String.capitalize(type) <> "View"
     Module.safe_concat([ApiWeb, view_name])
   end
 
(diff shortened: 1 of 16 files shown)


paulswartz:
Why isn't fields supported? I thought that was built into our JSON-API library?

handorff:
We weren't passing them to the JSON-API library because the way we determine which fields correspond to valid resources didn't work for resources which were more than 1 word. The most recent commit should fix that issue and add the appropriate parameters to the swagger documentation.

apps/api_web/lib/api_web/api_controller_helpers.ex
```diff
@@ -171,7 +171,7 @@ defmodule ApiWeb.ApiControllerHelpers do
   end
 
   defp view_module_for_type(type) do
-    view_name = String.capitalize(type) <> "View"
+    view_name = Macro.camelize(type) <> "View"
```
paulswartz: It doesn't look like this works: http://localhost:4000/live-facilities/park-alfcl-garage?fields[live-facility]=updated_at

apps/api_web/test/api_web/api_controller_helpers_test.exs
```diff
@@ -112,6 +112,15 @@ defmodule ApiWeb.ApiControllerHelpersTest do
       params = %{"stop" => nil}
       assert ApiWeb.ApiControllerHelpers.filter_valid_field_params(params) == %{"stop" => []}
     end
+
+    test "can provide fields for multi-word types" do
+      params = %{"live_facility" => "updated_at", "route_pattern" => "name,typicality"}
```
paulswartz: This should be `live-facility` to match the URL and `type` value for the resource.

apps/api_web/lib/api_web/api_controller_helpers.ex
paulswartz: Oh, that looks like like something we should fix:
- https://api-v3.mbta.com/live-facilities/park-woodl-garage has type `live-facility`
- https://api-v3.mbta.com/route-patterns/Red-1-1 has type `route_pattern`

Can we standardize on `hyphen-separated` like the URLs?

apps/api_web/test/api_web/api_controller_helpers_test.exs
handorff: Currently, the JSON-API library we use expects it to be `live_facility`. I believe this is set by `config :ja_serializer, key_format: :underscored` in `apps/api_web/config.exs`.

handorff: Or does it just use types that we set somewhere?

paulswartz: We must be setting it somewhere if it's different for live facilities and route patterns.

handorff:
Note: omitting `fields[live-facility]` and `fields[route-pattern]` from the documentation for now. Underscores vs hyphens discrepancies will require a new API version and will be resolved separately.

PR #53, opened 2019-05-09
Fix representative_trip relationship for /route-patterns

handorff:
`representative_trip` `data` field should be a single trip, not a list of trips of length 1.

### DIFF
--- a/apps/api_web/lib/api_web/views/route_pattern_view.ex
+++ b/apps/api_web/lib/api_web/views/route_pattern_view.ex
@@ -25,7 +25,14 @@ defmodule ApiWeb.RoutePatternView do
     :sort_order
   ])
 
+  defp fetch_representative_trip(trip_id) do
+    case State.Trip.by_id(trip_id) do
+      [] -> nil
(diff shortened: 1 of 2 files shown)


apps/api_web/lib/api_web/views/route_pattern_view.ex
```diff
@@ -25,7 +25,14 @@ defmodule ApiWeb.RoutePatternView do
     :sort_order
   ])
 
+  defp fetch_representative_trip(trip_id) do
+    case State.Trip.by_id(trip_id) do
```
paulswartz: - This is `List.first`: `trip_id |> State.Trip.by_id() |> List.first()`
- tests?

handorff: Are representative trips guaranteed to be primary?

paulswartz: At least one of them will be. Non-primary trips are those we duplicate based on their presence in `multi_route_trips.txt`

PR #56, opened 2019-05-16
support predictions with a vehicle on a different trip

paulswartz:
As a part of mbta/concentrate#90, we want to start providing the vehicle ID for predictions when we think a vehicle will be running that trip, even if it isn't currently running that trip. This updates the parsing and view to use a vehicle stored on the prediction, rather than looking it up via the trip.


apps/parse/lib/parse/trip_updates.ex
```diff
@@ -24,6 +24,7 @@ defmodule Parse.TripUpdates do
       trip_id: copy(update.trip.trip_id),
       route_id: copy(update.trip.route_id),
       direction_id: update.trip.direction_id,
+      vehicle_id: vehicle_id(update.vehicle),
```
dmaltzan: codecov says this line isn't covered

apps/parse/lib/parse/trip_updates.ex
```diff
@@ -53,6 +54,9 @@ defmodule Parse.TripUpdates do
     nil
   end
 
+  defp vehicle_id(%{id: id}), do: id
```
dmaltzan: codecov also complains about this line (57)

apps/parse/lib/parse/trip_updates.ex
paulswartz: Should be better now. Previously this whole file wasn't tested.

PR #57, opened 2019-05-20
Handle case where includes is a map

handorff:
`includes[key]` in a query will cause includes to be a map rather than a string. In this case:
- Don't crash in `ApiControllerHelpers.split_include`, then
- Catch and return error in `Params.validate_includes`

### DIFF
--- a/apps/api_web/lib/api_web/api_controller_helpers.ex
+++ b/apps/api_web/lib/api_web/api_controller_helpers.ex
@@ -178,9 +178,13 @@ defmodule ApiWeb.ApiControllerHelpers do
   def split_include(%{params: params} = conn, []) do
     split_include =
       case params["include"] do
-        nil -> []
-        %{} -> []
-        include -> include |> String.split(~r"[.,]") |> MapSet.new()
+        include when is_binary(include) ->
+          include
+          |> String.split([",", "."])
+          |> MapSet.new()
+
+        _ ->
+          []
       end
 
     assign(conn, :split_include, split_include)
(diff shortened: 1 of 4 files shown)


apps/api_web/lib/api_web/api_controller_helpers.ex
```diff
@@ -179,6 +179,7 @@ defmodule ApiWeb.ApiControllerHelpers do
     split_include =
       case params["include"] do
         nil -> []
+        %{} -> []
         include -> include |> String.split(~r"[.,]") |> MapSet.new()
```
paulswartz: What about having this as a list match, and letting everything else fall through?
```elixir
case Map.get(params, "include") do
  [_ | _] -> include |> String.split(~r"[.,]") |> MapSet.new()
  _ -> []
end
```

Also, can you test whether `String.split(include, ~r"[.,]")` or `String.split(include, [",", "."])` is more efficient?

apps/api_web/test/api_web/api_controller_helpers_test.exs
```diff
@@ -127,4 +127,14 @@ defmodule ApiWeb.ApiControllerHelpersTest do
       assert result[:fields] == %{"shape" => [:priority, :name]}
     end
   end
+
+  describe "split_include/2" do
+    test "doesn't error when includes is a map" do
+      conn =
+        %Plug.Conn{params: %{"bad" => ""}}
```
paulswartz: This doesn't seem like it's testing the includes?

handorff:
`String.split(include, [",", "."])` seems to be more efficient, so switched to using that.

PR #58, opened 2019-05-22
Filter /service by route

handorff:


apps/state/lib/state/service.ex
```diff
@@ -21,6 +21,18 @@ defmodule State.Service do
     end
   end
 
+  def by_route_ids(route_ids) do
+    route_ids
+    |> State.Trip.by_route_ids()
+    |> Enum.into(MapSet.new(), & &1.service_id)
```
paulswartz: `|> MapSet.new(& &1.service_id)`

apps/api_web/lib/api_web/controllers/service_controller.ex
```diff
@@ -33,6 +32,13 @@ defmodule ApiWeb.ServiceController do
       example: "1,2"
     )
 
+    parameter(
+      "filter[route]",
+      :query,
+      :string,
+      "Filter by route. Multiple `route_id` #{comma_separated_list()}."
```
paulswartz: "Multiple `route`"

PR #59, opened 2019-05-27
New API version - 2019-07-01

lboyarsky:
Tickets: 
1. [schedule: don't have arrival time for drop_off_type: 1 schedules or departure_time for pick_up_type: 1 schedules](https://app.asana.com/0/810933294009540/549739614850509/f)
2. [make use of underscores consistent in the API](https://app.asana.com/0/1122397568109186/1122192854528856/f)
3. [rename name -> long_name in /facilities](https://app.asana.com/0/810933294009540/1120944787243107/f)


lboyarsky:
@paulswartz any advice about additional unit tests? From my point of view, I've already covered both execution paths (older API versions, newer API versions).

apps/api_web/lib/api_web/controllers/live_facility_controller.ex
```diff
@@ -35,29 +35,40 @@ defmodule ApiWeb.LiveFacilityController do
...
-          |> Enum.group_by(& &1.facility_id)
-          |> Enum.map(fn {facilty_id, properties} ->
-            %{
-              facility_id: facilty_id,
-              properties: properties,
-              updated_at: updated_at(properties)
-            }
-          end)
-          |> State.all(Params.filter_opts(params, @pagination_opts))
+  def check_version(conn) do
```
paulswartz: You can probably implement this as a Plug, something like:
```elixir
# at the top
plug :ensure_path_matches version

def ensure_path_matches_version(conn, _) do
  if matches? do
    conn
  else
    conn
    |> do_404
    |> halt
  end
end
```

The `halt/1` will prevent any other processing from happening in the controller, and so the `index_data` and `show_data` function don't need to worry about that.

apps/api_web/lib/api_web/views/live_facility_view.ex
```diff
@@ -14,7 +14,7 @@ defmodule ApiWeb.LiveFacilityView do
 
   attributes([:properties, :updated_at])
 
-  def type(_, _), do: "live-facility"
+  def type(_, _), do: "live_facility"
```
paulswartz: This needs to stay `live-facility` for the current version. I believe the second argument here is a `conn`.

apps/api_web/lib/api_web/controllers/live_facility_controller.ex
lboyarsky: Great idea! Thanks for letting me know. Refactored.

apps/api_web/lib/api_web/views/live_facility_view.ex
lboyarsky: Fixed

apps/state/lib/state/schedule.ex
```diff
@@ -259,6 +260,13 @@ defmodule State.Schedule do
     do_time_filter(schedules, 0, max)
   end
 
+  defp do_post_search_filter(schedules, %{remove_unsupported_times: true}) do
```
handorff: I don't think that you'll actually do this filter if there's a min or max time.

apps/api_web/lib/api_web/views/live_facility_view.ex
```diff
@@ -14,6 +14,7 @@ defmodule ApiWeb.LiveFacilityView do
 
   attributes([:properties, :updated_at])
 
+  def type(_, %{assigns: %{api_version: ver}}) when ver >= "2019-07-01", do: "live_facility"
```
handorff: I don't have a very strong preference, but I find it a bit more readable when the case with the API version is for old versions and the default is for new versions, so I would slightly prefer `def type(_, _), do: "live_facility"` and then check if we're below 2019-07-01 to return `"live-facility"`.

lboyarsky: Good point, refactored.

apps/state/lib/state/schedule.ex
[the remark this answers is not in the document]
lboyarsky: And here I was thinking that I found a clever way to do it 🙂 Refactored.

lboyarsky: Moved it to the view as per Paul's comment.

apps/api_web/lib/api_web/controllers/facility_controller.ex
```diff
@@ -43,6 +43,7 @@ defmodule ApiWeb.FacilityController do
       |> format_filters()
       |> Facility.filter_by()
       |> State.all(Params.filter_opts(params, @pagination_opts))
+      |> format_names(conn)
```
paulswartz: What about?
1) renaming the field in Model.Facility to long_name
2) In FacilityView.attributes/2, check the version and if it's older, do `Map.put(attributes, :name, facility.long_name)`

apps/api_web/lib/api_web/controllers/facility_controller.ex
```diff
@@ -142,8 +156,16 @@ defmodule ApiWeb.FacilityController do
             )
 
             name(
+              :string,
```
paulswartz: We'd just rename the field in the documentation: we don't need to maintain it for older versions.

apps/api_web/lib/api_web/controllers/live_facility_controller.ex
```diff
@@ -35,6 +37,21 @@ defmodule ApiWeb.LiveFacilityController do
     response(429, "Too Many Requests", Schema.ref(:TooManyRequests))
   end
 
+  defp ensure_path_matches_version(conn, _) do
```
paulswartz: Because you know that you came from one of these paths (because of the router), you should only need to check one of them.

apps/api_web/lib/api_web/views/schedule_view.ex
```diff
@@ -61,10 +61,16 @@ defmodule ApiWeb.ScheduleView do
     State.Prediction.prediction_for(schedule, date)
   end
 
+  def arrival_time(%{drop_off_type: 1}, %{assigns: %{api_version: ver}}) when ver >= "2019-07-01",
```
paulswartz: Per Paul's comment, I'd do this the other way:
1) Update the parsing to have the data in the format we want
2) In the view, if we have an older version and a missing time, then copy it from the `departure_time` (and vice versa)

apps/api_web/lib/api_web/controllers/facility_controller.ex
lboyarsky: Refactored.

apps/api_web/lib/api_web/controllers/facility_controller.ex
lboyarsky: Fixed.

apps/api_web/lib/api_web/controllers/live_facility_controller.ex
lboyarsky: Fixed, but it seems less readable to me like that.

apps/api_web/lib/api_web/views/schedule_view.ex
lboyarsky: Done.

apps/api_web/lib/api_web/views/schedule_view.ex
```diff
@@ -61,10 +61,24 @@ defmodule ApiWeb.ScheduleView do
...
+      )
+      when ver < "2019-07-01",
+      do: format_time(seconds_past_midnight, conn)
+
   def arrival_time(%{arrival_time: seconds_past_midnight}, conn) do
     format_time(seconds_past_midnight, conn)
   end
 
+  def departure_time(
+        %{pickup_type: 1, arrival_time: seconds_past_midnight},
```
paulswartz: I think this is right, but it's more forward-compatible if the match is `departure_time: nil, arrival_time: seconds` (and vice versa). That way, if there's another reason we have `nil` times in the future, older versions will still be supported.

apps/api_web/test/api_web/controllers/schedule_controller_test.exs
```diff
@@ -300,6 +300,31 @@ defmodule ApiWeb.SchedulerControllerTest do
       assert index_data(conn, put_in(params["max_time"], "5:30")) == []
     end
 
+    test "doesn't excludes arrival and departure times for older API versions", %{conn: conn} do
```
paulswartz: This test doesn't make sense anymore.

apps/api_web/lib/api_web/views/schedule_view.ex
lboyarsky: Well, if there's another reason we have `nil`, we'd probably want to have different logic to handle it. Anyway, changed like you say.

apps/api_web/test/api_web/controllers/schedule_controller_test.exs
lboyarsky: Removed. I've added similar tests to `schedule_view_test.exs` yesterday.

PR #62, opened 2019-05-29
Prediction state

paulswartz:

PR #65, opened 2019-06-05
feat: multiple connections to Memcache

paulswartz:
Previously, we only kept open one connection to Memcache. However, under high
traffic this can cause a queue for the single connection. This creates a
small worker pool of connections (managed by a Registry), and uses that to
pick a random child for each request.

Pulled out of #62.


lboyarsky:
I see no issues with the code, but don't we need to introduce some tests for this?

paulswartz:
They're covered by the existing memcached tests: https://github.com/mbta/api/blob/master/apps/api_web/test/api_web/rate_limiter/memcache_test.exs

PR #66, opened 2019-06-05
feat: track the number of concurrent requests

paulswartz:
At the moment, we don't know how many concurrent request clients are
making. This will allow us to gather some data.


paulswartz:
I can add a test, but I believe the teardown code we add with `register_before_send` is always called, even if we use `halt` to stop other processing.

PR #68, opened 2019-06-05
set a deadline of 20 seconds for API responses

paulswartz:


apps/api_web/lib/api_web/endpoint.ex
```diff
@@ -24,6 +24,7 @@ defmodule ApiWeb.Endpoint do
   plug(Plug.MethodOverride)
   plug(Plug.Head)
   # CORS needs to be before the router, and Authenticate needs to be before CORS
```
handorff: nit: this comment should probably be one line lower

PR #69, opened 2019-06-07
Added logic to handle nil arrival/departure schedule times when sorting

lboyarsky:
Ticket: [sorting schedules in earlier versions should work as if both values are present](https://app.asana.com/0/810933294009540/1126222348431249/f)

### DIFF
--- a/apps/api_web/lib/api_web/controllers/schedule_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/schedule_controller.ex
@@ -99,7 +99,7 @@ defmodule ApiWeb.ScheduleController do
           # greater than 1 because `date` is automatically included
           filters
           |> Schedule.filter_by()
-          |> populate_extra_dates(conn)
+          |> populate_extra_times(conn)
           |> State.all(Params.filter_opts(params, @pagination_opts))
 
         _ ->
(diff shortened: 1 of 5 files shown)


apps/state/lib/state.ex
```diff
@@ -182,10 +182,20 @@ defmodule State do
     Enum.sort_by(results, &mapper_fn(&1, key), &>=/2)
   end
 
-  defp mapper_fn(result, key) do
+  defp mapper_fn(result, key, first_run \\ true) do
```
paulswartz: This doesn't seem like the right place to handle it, as this doesn't have anything to do with the version. (However, this will be a good place to address [sort=time](https://app.asana.com/0/810933294009540/1126222348431251) in the future.)

lboyarsky: > Maybe an extra function here between 101-102 to post-process the data and insert the missing values? Then you could also remove the extra code in the ScheduleView.

This is how I initially implemented it but then I realized that we were going to have different data based on the sort order (those _inserted missing values_ are going to reach clients, correct?), which didn't feel right to me. Do you think it is ok? 

> This doesn't seem like the right place to handle it

I agree but on the other hand we have special case for distance based sorting here, why don't have special case for schedule times as well? 

> as this doesn't have anything to do with the version. (However, this will be a good place to address sort=time in the future.)

I think what I have implemented here is more or less the same as what is asked in "sort=time" task, I just haven't introduced the new parameter name. I don't really see a use-case for any other way of sorting the schedule times (e.g. one, which would put nil's to the top or to the bottom of the list), or do you think it could be useful?

paulswartz: We'd only be inserting the extra values for the older versions where we're displaying them, so it makes sense that they affect the sort order as well as the display. The difference w/ distance is that it's a special case of sorting, not a special case of the data: your code would make the change for all types, including predictions where we don't want to change the sorting behavior.

lboyarsky: Ok, refactored

apps/api_web/lib/api_web/controllers/schedule_controller.ex
```diff
@@ -99,6 +99,7 @@ defmodule ApiWeb.ScheduleController do
           # greater than 1 because `date` is automatically included
           filters
           |> Schedule.filter_by()
+          |> populate_extra_dates(conn)
```
paulswartz: nitpick: they're times, not dates :)

lboyarsky: Fixed

PR #70, opened 2019-06-11
improved memory usage

paulswartz:
Changes to improve the memory usage of the API

- when using Gather to wait for multiple pieces of data from the GTFS parsing, erase the data once all of it is received (rather than keeping it in the process memory)
- Hibernate (run GC and pause the process) most of the Server processes (as they only receive messages during startup)
- Hibernate EventStream processes to cleanup memory between sending

### DIFF
--- a/apps/api_web/lib/api_web/event_stream.ex
+++ b/apps/api_web/lib/api_web/event_stream.ex
@@ -7,16 +7,24 @@ defmodule ApiWeb.EventStream do
   import Plug.Conn
   alias __MODULE__.Supervisor
 
-  @type state :: {Plug.Conn.t(), pid}
+  @enforce_keys [:conn, :pid, :timeout]
+  defstruct @enforce_keys ++ [:timer]
 
-  @spec call(Plug.Conn.t(), module, map) :: Plug.Conn.t()
+  @typep state :: %__MODULE__{
+           conn: Plug.Conn.t(),
+           pid: pid,
+           timeout: non_neg_integer,
+           timer: reference | nil
+         }
+
+  @spec call(Plug.Conn.t(), module, map) :: no_return
   def call(conn, module, _params) do
     state = initialize(conn, module)
-    event_stream_loop(&event_stream_loop/2, state)
+    hibernate_loop(state)
   end
 
   @spec initialize(Plug.Conn.t(), module) :: state
-  def initialize(conn, module) do
+  def initialize(conn, module, timeout \\ 30_000) do
     {:ok, pid} = Supervisor.server_child(conn, module)
 
     conn =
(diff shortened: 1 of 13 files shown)


apps/events/lib/events/gather.ex
```diff
@@ -4,19 +4,20 @@ defmodule Events.Gather do
...
   def new(keys, callback) when is_list(keys) and is_function(callback, 1) do
-    %__MODULE__{keys: Enum.sort(keys), received: %{}, callback: callback}
+    %__MODULE__{keys: MapSet.new(keys), callback: callback}
   end
 
   def update(%__MODULE__{keys: keys, received: received, callback: callback} = state, key, value) do
     received = Map.put(received, key, value)
 
-    if received |> Map.keys() |> Enum.sort() == keys do
+    if received |> Map.keys() |> MapSet.new() == keys do
```
losvedir: I'm not sure enough about MapSet to know if `==` works for equality. Should this be `MapSet.equal?/2`?

Also, is the only purpose of `keys` to keep track of if everything has been received yet? Would it be any faster if `update/3` here did a `MapSet.delete` whenever a new key came in, and then the conditional was to see if it was empty?

paulswartz: In general no, `==` is not good enough because `MapSet` has two versions. However, it's only an issue if we're getting a MapSet from a previous version of Elixir which we aren't.

Interesting idea on checking the other way: I'll do a quick benchmark to find out.

paulswartz: They're basically the same. `gather` is the current code, `reverse` is removing items from keys. The inputs are the number of keys we're checking for.

```
##### With input five #####
Name              ips        average  deviation         median         99th %
gather       579.97 K        1.72 μs  ±1500.58%        1.99 μs        2.99 μs
reverse      575.13 K        1.74 μs  ±1461.76%        1.99 μs        2.99 μs

Comparison:
gather       579.97 K
reverse      575.13 K - 1.01x slower +0.0145 μs

##### With input one #####
Name              ips        average  deviation         median         99th %
reverse        1.20 M      831.65 ns  ±3784.37%         990 ns         990 ns
gather         1.18 M      849.04 ns  ±3797.90%         990 ns         990 ns

Comparison:
reverse        1.20 M
gather         1.18 M - 1.02x slower +17.39 ns

##### With input ten #####
Name              ips        average  deviation         median         99th %
reverse      322.86 K        3.10 μs  ±1149.70%        2.99 μs        4.99 μs
gather       322.45 K        3.10 μs   ±915.33%        2.99 μs        5.99 μs

Comparison:
reverse      322.86 K
gather       322.45 K - 1.00x slower +0.00397 μs
```

apps/state/lib/state/server.ex
```diff
@@ -118,7 +119,7 @@ defmodule State.Server do
       Indicies in the `:mnesia` table where state is stored
       """
       @spec indicies :: [atom]
-      def indicies, do: unquote(opts[:indicies])
+      def indicies, do: unquote(indicies)
```
losvedir: I believe the correct spelling is "indices". But I see this is a configurable option and a public function so changing it might have far reaching implications that might not be worth it. Could always live with it a la HTTP Referer.

paulswartz: I'll put in a separate ticket to address.

apps/api_web/lib/api_web/event_stream.ex
```diff
@@ -7,16 +7,23 @@ defmodule ApiWeb.EventStream do
   import Plug.Conn
   alias __MODULE__.Supervisor
 
-  @type state :: {Plug.Conn.t(), pid}
+  defstruct [:conn, :timer, :timeout, :pid]
```
losvedir: For struct keys that shouldn't be `nil`, like `conn`, `timeout`, and `pid` according to the typespec below, yet nevertheless don't have good defaults, in our apps we use `@enforce_keys`, for an extra level of protection when constructing the struct.

```ex
@enforce_keys [:conn, :timeout, :pid]
defstruct @enforce_keys ++ [:timer]
```

Just an idea, but our team likes it.

apps/api_web/lib/api_web/event_stream.ex
```diff
@@ -26,43 +33,55 @@ defmodule ApiWeb.EventStream do
...
+    ensure_timer(%__MODULE__{conn: conn, pid: pid, timeout: timeout})
   end
 
-  @spec event_stream_loop(loop, state) :: Plug.Conn.t() when loop: (loop, state -> Plug.Conn.t())
-  def event_stream_loop(next_call, {conn, pid} = state) do
-    with {:ok, new_conn} <- receive_result(state) do
-      next_call.(next_call, {new_conn, pid})
-    else
-      _ ->
-        Supervisor.server_unsubscribe(pid)
```
losvedir: It doesn't look like `Supervisor.server_unsubscribe/1` is ever called in the new code. Is it unnecessary with this new approach?

paulswartz: Generally yes (the only way out terminates the process, which automatically unsubscribes), but it can't hurt to keep it in.

apps/events/lib/events/gather.ex
losvedir: Ah, neat! Thanks for checking.

apps/api_web/lib/api_web/event_stream.ex
paulswartz: :+1: added!

PR #71, opened 2019-06-13
Support for fetching facility name for older API versions

lboyarsky:
Ticket: [Facility name is not fetched for older API versions](https://app.asana.com/0/810933294009540/1126993432964133/f)

```bash
$ curl -g "http://localhost:4000/stops/place-brntn?fields[facility]=name,type&include=facilities" |jsonpp
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  7207  100  7207    0     0   1997      0  0:00:03  0:00:03 --:--:--  1996
{
  "data": {
    "attributes": {
      "address": "197 Ivory St, Braintree, MA 02184",
      "description": null,
      "latitude": 42.207854,
      "location_type": 1,
      "longitude": -71.001138,
      "name": "Braintree",
      "platform_code": null,
      "platform_name": null,
      "wheelchair_boarding": 1
    },
    "id": "place-brntn",
    "links": {
      "self": "/stops/place-brntn"
    },
    "relationships": {
      "child_stops": {},
      "facilities": {
        "data": [
          {
            "id": "333",
            "type": "facility"
          },
    
       [...]

        ],
        "links": {
          "related": "/facilities/?filter[stop]=place-brntn"
        }
      },
      "parent_station": {
        "data": null
      },
      "recommended_transfers": {},
      "zone": {
        "data": {
          "id": "CR-zone-2",
          "type": "zone"
        }
      }
    },
    "type": "stop"
  },
  "included": [
    {
      "attributes": {
        "name": "Braintree Escalator 333 (Busway to Commuter Rail platform, parking garage)",
        "type": "ESCALATOR"
      },
      "id": "333",
      "links": {
        "self": "/facilities/333"
      },
      "relationships": {
        "stop": {
          "data": {
            "id": "place-brntn",
            "type": "stop"
          }
        }
      },
      "type": "facility"
    },

[...]

  ],
  "jsonapi": {
    "version": "1.0"
  }
}
```

### DIFF
--- a/apps/api_web/lib/api_web/api_controller_helpers.ex
+++ b/apps/api_web/lib/api_web/api_controller_helpers.ex
@@ -158,13 +158,12 @@ defmodule ApiWeb.ApiControllerHelpers do
   defp do_filter_valid_field_attributes(conn, {type, fields}) do
     view_module = view_module_for_type(type)
 
-    case view_module.attribute_set(conn) do
-      attribute_set when map_size(attribute_set) > 0 ->
-        fields
-        |> String.split(",")
-        |> Enum.filter(&MapSet.member?(attribute_set, &1))
-        |> Enum.map(&String.to_existing_atom/1)
-    end
+    attr_filter = fn attr -> conn |> view_module.attribute_set |> MapSet.member?(attr) end
+
+    fields
+    |> String.split(",")
+    |> Enum.filter(attr_filter)
+    |> Enum.map(&String.to_existing_atom/1)
   end
 
   defp view_module_for_type(type) do
--- a/apps/api_web/lib/api_web/api_controller_helpers.ex
+++ b/apps/api_web/lib/api_web/api_controller_helpers.ex
@@ -158,21 +158,14 @@ defmodule ApiWeb.ApiControllerHelpers do
   defp do_filter_valid_field_attributes(conn, {type, fields}) do
     view_module = view_module_for_type(type)
 
-    case view_module.__attributes() do
-      [_ | _] = attributes ->
-        attribute_set = MapSet.new(attributes, &Atom.to_string/1)
-
+    case view_module.attribute_set(conn) do
+      attribute_set when map_size(attribute_set) > 0 ->
         fields
         |> String.split(",")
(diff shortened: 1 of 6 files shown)


apps/api_web/lib/api_web/api_controller_helpers.ex
```diff
@@ -161,11 +164,16 @@ defmodule ApiWeb.ApiControllerHelpers do
 
         fields
         |> String.split(",")
-        |> Enum.filter(&MapSet.member?(attribute_set, &1))
+        |> Enum.filter(&(MapSet.member?(attribute_set, &1) || older_param(conn, type, &1)))
         |> Enum.map(&String.to_existing_atom/1)
     end
   end
 
+  def older_param(%{assigns: %{api_version: ver}}, "facility", "name") when ver < "2019-07-01",
```
paulswartz: Rather than having this as an addition to `ApiControllerHelpers`, can you defer to the `view_module`? Something like:
```elixir
attribute_set = view_module.attribute_set(conn)

# in api_view_helpers
def attribute_set(_conn) do
  MapSet.new(__MODULE__.__attributes(), &Atom.to_string/1)
end
defoverridable [attribute_set: 1]

# in FacilityView
def attribute_set(conn) do
  ...behavior...
end
```

lboyarsky: Yes, it looks much better now, thanks. Refactored.

apps/api_web/lib/api_web/views/facility_view.ex
```diff
@@ -73,4 +73,10 @@ defmodule ApiWeb.FacilityView do
       }
     end
   end
+
+  def attribute_set(%{assigns: %{api_version: ver}} = conn) when ver < "2019-07-01" do
+    conn |> super() |> MapSet.put("name")
```
paulswartz: - Can we make the new version the first case, and the older versions the fallback?
- I was thinking you should drop `long_name`, but I think that'll break the website?

lboyarsky: Fixed. Yes, website switched to using `long_name` and they're still using the same (older) API version that they used before, so removing `long_name` would break it for them indeed.

PR #74, opened 2019-06-21
Fix distance always a valid sort

handorff:
`valid_order_by_key?` is only called with key `:distance` if `opts_map` is missing one or both of `:latitude` or `:longitude`, so it should always be invalid rather than always being valid.

### DIFF
--- a/apps/state/test/state_test.exs
+++ b/apps/state/test/state_test.exs
@@ -65,6 +65,30 @@ defmodule StateTest do
              ) == [three, one, two]
     end
 
+    test "sorting by distance returns error when missing lat/lng" do
+      items = [
+        %{latitude: 1.0, longitude: 2.0},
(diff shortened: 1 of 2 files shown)


apps/state/test/state_test.exs
```diff
@@ -65,6 +65,31 @@ defmodule StateTest do
              ) == [three, one, two]
     end
 
+    test "sorting by distance returns error when missing lat/lng" do
+      items = [
+        one = %{latitude: 1.0, longitude: 2.0},
+        two = %{latitude: 1.0, longitude: 1.0},
+        three = %{latitude: 1.0, longitude: 3.0}
```
paulswartz: This probably won't validate because you aren't using `one` `two` or `three`. You also probably only one or two items.

PR #75, opened 2019-06-25
Add sort=time to sort by whichever time is present

handorff:


apps/state/lib/state.ex
```diff
@@ -197,6 +204,24 @@ defmodule State do
...
 
+  defp time(%{arrival_time: nil, departure_time: time}) do
+    time
+  end
+
+  defp time(%{arrival_time: time}) do
+    time
+  end
+
+  defp time(_), do: nil
```
paulswartz: Do you need this case, since you're checking that one of the times is present?

apps/state/lib/state.ex
```diff
@@ -197,6 +204,24 @@ defmodule State do
...
+
+  def sort_by_time(results, :asc) do
+    Enum.sort_by(results, &time(&1), &<=/2)
+  end
+
+  def sort_by_time(results, :desc) do
+    Enum.sort_by(results, &time(&1), &>=/2)
```
paulswartz: `&time/1`

PR #79, opened 2019-07-01
Error on invalid sort which is not an existing atom

handorff:
Opening a PR to get feedback on this, since it's a bit hacky.

The atom `:invalid` is not an attribute of any model and is always an invalid sort, so trying to sort by any key which is not an existing atom will return: `{"errors":[{"code":"bad_request","detail":"Invalid sort key.","source":{"parameter":"sort"},"status":"400"}],"jsonapi":{"version":"1.0"}}`.

The alternative would be to change `filter_opts` to return something along the lines of `{:ok, _}` or `{:error, _}`, then modify ~15 calls to `filter_opts` to handle the error case. This would allow us to specify which sort key is invalid, though we don't actually do this now (as you can see from the response above). On the other hand, it would split the handling of bad sort keys across two different places depending on whether the bad key is an existing atom or not, which seems less readable.

### DIFF
--- a/apps/api_web/lib/api_web/controllers/alert_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/alert_controller.ex
@@ -122,7 +122,7 @@ defmodule ApiWeb.AlertController do
       |> apply_filters()
       |> case do
         list when is_list(list) ->
-          State.all(list, Params.filter_opts(params, @pagination_opts))
+          State.all(list, Params.filter_opts(params, @pagination_opts, conn))
 
         {:error, _} = error ->
           error
--- a/apps/api_web/lib/api_web/params.ex
+++ b/apps/api_web/lib/api_web/params.ex
@@ -24,34 +24,34 @@ defmodule ApiWeb.Params do
 
   ## Examples
 
-      iex> ApiWeb.Params.filter_opts(%{"page" => %{"offset" => 0}}, [:offset])
+      iex> ApiWeb.Params.filter_opts(%{"page" => %{"offset" => 0}}, [:offset], build_conn())
       [offset: 0]
 
-      iex> ApiWeb.Params.filter_opts(%{"page" => %{"limit" => 10}}, [:limit])
+      iex> ApiWeb.Params.filter_opts(%{"page" => %{"limit" => 10}}, [:limit], build_conn())
       [limit: 10]
 
-      iex> ApiWeb.Params.filter_opts(%{"sort" => "name"}, [:order_by])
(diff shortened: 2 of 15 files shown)


apps/api_web/lib/api_web/controllers/alert_controller.ex
```diff
@@ -122,7 +122,7 @@ defmodule ApiWeb.AlertController do
       |> apply_filters()
       |> case do
         list when is_list(list) ->
-          State.all(list, Params.filter_opts(params, @pagination_opts))
+          State.all(list, Params.filter_opts(params, @pagination_opts, conn))
```
paulswartz: `params` are a part of `conn`: do you need both?

apps/api_web/test/api_web/params_test.exs
```diff
@@ -3,26 +3,44 @@ defmodule ApiWeb.ParamsTest do
...
            ]
   end
 
+  test "filter_opts returns invalid sort on versions after 2019-07-01", %{conn: conn} do
+    conn = assign(conn, :api_version, "2019-02-12")
+    params = %{"sort" => "notavalidsort"}
+
+    assert Params.filter_opts(params, [:order_by], conn) == []
+
+    conn = assign(conn, :api_version, "2019-07-01")
```
paulswartz: If you do this one first, you don't need to set the version. We automatically set the version to the highest one in `ConnCase`.

apps/api_web/lib/api_web/params.ex
```diff
@@ -91,10 +91,17 @@ defmodule ApiWeb.Params do
 
     [{:order_by, order_by} | acc]
   rescue
-    ArgumentError -> acc
+    ArgumentError ->
+      case conn.assigns.api_version do
+        version when version >= "2019-07-01" ->
```
paulswartz: You can do this with an `if`:
```elixir
if conn.assigns.api_version >= "2019-07-01" do
  ...
else
  ...
end
```

apps/api_web/lib/api_web/controllers/alert_controller.ex
handorff: Are they the same params? I see a lot of examples of functions which take `params` as well as `conn`; e.g. `index_data/2` in each controller.

Separately, is there any cost to passing the whole `conn` around? I'm only using the `api_version`, so I could also rewrite to just pass the version to `Params.filter_opts` if that would be preferable.

paulswartz: It's all in the same process, so I don't think there's any issue with passing the conn around. We can leave it as a separate parameter for now.

PR #81, opened 2019-07-05
Return the intersection rather than the union of different facilities filters

handorff:


apps/state/lib/state/facility.ex
```diff
@@ -89,11 +89,25 @@ defmodule State.Facility do
...
-    Enum.uniq_by(facilities, & &1.id)
+    [first_result] = Enum.take(search_results, 1)
+
+    search_results
+    |> Stream.drop(1)
+    |> Enum.reduce(first_result, fn results, acc ->
+      acc_set = MapSet.new(acc)
+      Enum.filter(results, fn facility -> facility in acc_set end)
+    end)
+    |> Enum.uniq_by(& &1.id)
```
paulswartz: Since you're already creating a MapSet, maybe we can be more efficient?
```elixir
Enum.reduce(search_results, :no_results, fn
  results, %MapSet{} = acc -> MapSet.intersection(acc, MapSet.new(results))
  results, :no_results -> MapSet.new(results)
end)
```

apps/state/test/state/facility_test.exs
```diff
@@ -94,7 +94,7 @@ defmodule State.FacilityTest do
         |> State.Facility.filter_by()
         |> Enum.sort_by(& &1.id)
 
-      assert sorted_results == [by_id("3"), by_id("6"), by_id("701")]
+      assert sorted_results == [by_id("3")]
```
paulswartz: Are there remaining tests about the sorting?

handorff: There aren't any tests explicitly of sorting, but that seems like it would belong in the controller testing.

There are other tests which have multiple results and sort them by ID, but I interpreted those as sorting only so that we can be guaranteed equality when comparing lists.

paulswartz: Ah, I see. I misunderstood the test. :+1:

apps/state/lib/state/facility.ex
handorff: This definitely looks better, but unfortunately causes a Dialyzer error:

```
    apps/state/lib/state/facility.ex:105: The call 'Elixir.MapSet':intersection(_acc@1::#{'__struct__':='Elixir.MapSet', _=>_},'Elixir.MapSet':t(_)) does not have an opaque term of type 'Elixir.MapSet':t(_) as 1st argument
```

This seems to be a known issue, but I don't know that there's any resolution on either the Elixir or the Dialyzer side: https://github.com/elixir-lang/elixir/issues/8463

PR #85, opened 2019-07-16
Filter /predictions by route_pattern_id

handorff:
Asana ticket: [API: query predictions by route_pattern](https://app.asana.com/0/810933294009540/1128756751923940/f)

### DIFF
--- a/apps/api_web/lib/api_web/controllers/prediction_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/prediction_controller.ex
@@ -62,7 +62,7 @@ defmodule ApiWeb.PredictionController do
     filter_param(:id, name: :trip)
 
     parameter("filter[route_pattern]", :query, :string, """
-    Filter by `route_pattern_id`: returns predictions associated with a trip with the given `route_pattern_id`. Multiple `route_pattern_id` #{
+    Filter by `/included/{index}/relationships/route_pattern/data/id` of a trip. Multiple `route_pattern_id` #{
       comma_separated_list()
     }.
     """)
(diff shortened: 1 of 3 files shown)


apps/api_web/lib/api_web/controllers/prediction_controller.ex
```diff
@@ -61,6 +61,12 @@ defmodule ApiWeb.PredictionController do
     filter_param(:id, name: :route)
     filter_param(:id, name: :trip)
 
+    parameter("filter[route_pattern]", :query, :string, """
+    Filter by `route_pattern_id`: returns predictions associated with a trip with the given `route_pattern_id`. Multiple `route_pattern_id` #{
```
lboyarsky: Probably should say something like "Filter by `/included/{index}/relationships/route_pattern/data/id`" to be consistent with description for other filters, there is no `route_pattern_id` field in response. 

image

apps/api_web/lib/api_web/controllers/prediction_controller.ex
```diff
@@ -275,6 +291,16 @@ defmodule ApiWeb.PredictionController do
     |> select(:stop_id)
   end
 
+  defp all_stops_and_route_patterns(stop_ids, route_pattern_ids, matchers) do
+    trip_ids =
+      %{}
+      |> Map.put(:route_patterns, route_pattern_ids)
```
lboyarsky: Could be just `Map.new(route_patterns: route_pattern_ids)` or simply `%{route_patterns: route_pattern_ids}` I think

apps/api_web/lib/api_web/controllers/prediction_controller.ex
```diff
@@ -275,6 +291,16 @@ defmodule ApiWeb.PredictionController do
...
+      |> State.Trip.filter_by()
+      |> Enum.map(& &1.id)
```
lboyarsky: Can probably be `Stream.map`

apps/api_web/lib/api_web/controllers/prediction_controller.ex
handorff: I wasn't sure what to do here since `route_pattern_id` isn't in attributes or in relationships. (It's the associated trip's `route_pattern_id`.)

Does it make sense to say something like: Filter by `/included/{index}/relationships/trip/included/{index}/route_pattern/id`?

lboyarsky: I think the path here should correspond with actual path in the result, which would be `/included/{index}/relationships/route_pattern/data/id`

image

PR #86, opened 2019-07-17
Create table mapping stops to route patterns at that stop

handorff:
Asana ticket: [generate table mapping stop ID -> route_patterns](https://app.asana.com/0/810933294009540/1130249504961680/f)

This PR adds `State.RoutePatternsAtStop`, which is similar to `State.RoutesAtStop` except that it caches the `route_pattern_ids` which go to each stop instead of the `route_ids`. The code is almost exactly the same (to the point where credo fails due to duplicated code).

If we were maintaining both the route and route_pattern cache long-term, we would definitely want to factor out most of these modules. However, I think that we should only maintain `State.RoutePatternsAtStop`, and re-implement the functions which query `State.RoutesAtStop` by finding the route patterns for the provided stops/directions and then fetching the routes for those route patterns.

We can't do this quite yet, since ADDED trips don't have `route_pattern_ids` (though they do have `route_ids`) so it's possible that we would be missing some routes. I propose that we do the following:
1. Merge this PR, so that we can...
2. Add the ability to filter `/route_patterns` by `stop_id`.
3. Generate `route_pattern_ids` for ADDED trips / any other trips which are missing them.
4. Reimplement `State.RoutesAtStop` to query the `RoutePatternsAtStop` ETS table as described above, and delete the separate table for `RoutesAtStop`.

Any objections to or foreseeable issues with this plan? I could also refactor this PR (if we don't want to temporarily duplicate code), or we could first do the work to generate `route_pattern_ids` for ADDED trips (if we don't want to temporarily maintain two mostly-redundant ETS tables).


apps/state/lib/state/route_patterns_at_stop.ex
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  defp do_gather_route({route_id, trips}) do
+    trips
+    |> Stream.reject(&ignore_trip_for_route?/1)
+    |> Enum.group_by(&{&1.direction_id, &1.service_id})
+    |> Stream.flat_map(&do_gather_stops_on_trips/1)
+    |> Stream.uniq()
+    |> Stream.map(fn {stop_id, direction_id, service_id} ->
+      {stop_id, direction_id, service_id, route_id}
```
lboyarsky: Do we need `direction_id` and `service_id`? I would assume that we don't based on what I read in the ticket. The table would be much smaller w/o them.

PR #88, opened 2019-07-18
Switching to latest Elixir and Erlang versions

lboyarsky:
Ticket: [update API to Erlang 22, Elixir 1.9](https://app.asana.com/0/810933294009540/1129956922358338/f)

This seems to be working fine, I haven't found any issues. If there is a way to compare the performance between new and older versions (some splunk report for dev env?), it would be great to do it given that I had to disable native compilation option for csv parsing. 

I also suggest we switch to native releases in separate ticket given that:
- should there be any performance degradation we won't understand whether it comes from versions update or from switching to the native releases;
- I'm not sure how to test those releases.


apps/parse/lib/simple_csv.ex
```diff
@@ -3,8 +3,6 @@ defmodule SimpleCSV do
   A simpler, but higher-performance, CSV parser.  Expects that that lines being
   parsed don't contain any commas or quotes inside quoted fields.
   """
-  @compile :native
```
lboyarsky: Apparently native compilation is not supported anymore: https://groups.google.com/forum/#!topic/erlang-programming/z7rAKtAaClc

apps/api_web/mix.exs
```diff
@@ -72,8 +72,8 @@ defmodule ApiWeb.Mixfile do
       {:benchwarmer, "~> 0.0", only: [:dev, :test]},
       {:dialyxir, ">= 0.0.0", only: [:dev, :test]},
       {:logster, "~> 1.0"},
-      {:phoenix_swagger, "~> 0.8.1"},
-      {:ex_json_schema, "< 0.6.0"},
+      {:phoenix_swagger, github: "mbta/phoenix_swagger", branch: "lev-elixir19"},
```
lboyarsky: Should change branch to specific commit ref once https://github.com/mbta/phoenix_swagger/pull/2 lands

lboyarsky:
@paulswartz looks like codecov is not reporting the status again.

PR #92, opened 2019-07-25
Filter /route_patterns by stop

handorff:


apps/state/lib/state/route_pattern.ex
```diff
@@ -38,6 +41,18 @@ defmodule State.RoutePattern do
     by_route_ids(route_ids)
   end
 
+  def filter_by(%{stop_ids: stop_ids, direction_id: direction_id}) do
+    stop_ids
+    |> RoutesPatternsAtStop.route_patterns_by_stops_and_direction([{:direction_id, direction_id}])
```
paulswartz: You can combine these:
```elixir
def filter_by(%{stop_ids: stop_ids} = filters) do
  opts = case filters do
    %{direction_id: direction_id} -> [direction_id: direction_id]
    _ -> []
  end
  stop_ids
  |> RoutesPatternsAtStop.route_patterns_by_stops_and_direction(opts)
  |> by_ids
end
```

PR #93, opened 2019-07-31
Set signing_salt and secret_key_base with env vars

handorff:
Separately, we need to actually set the environment variables as part of prod deploy.

### DIFF
--- a/apps/api_web/config/config.exs
+++ b/apps/api_web/config/config.exs
@@ -9,11 +9,11 @@ use Mix.Config
 config :api_web, ApiWeb.Endpoint,
   url: [host: "localhost"],
   root: Path.dirname(__DIR__),
-  secret_key_base: "${SECRET_KEY_BASE}",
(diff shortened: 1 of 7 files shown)


apps/api_web/config/config.exs
```diff
@@ -9,10 +9,12 @@ use Mix.Config
 config :api_web, ApiWeb.Endpoint,
   url: [host: "localhost"],
   root: Path.dirname(__DIR__),
-  secret_key_base: "v1EHfW07QPr8ai7bi0hooadtBorROPNjhSWx7CGv7AiCOhEyGoeT1jagMTNCE3PU",
+  secret_key_base: System.get_env("SECRET_KEY_BASE"),
   render_errors: [accepts: ~w(json html)],
   http: [compress: true, protocol_options: [idle_timeout: 86_400_000]]
 
+config :api_web, :signing_salt, System.get_env("SIGNING_SALT")
```
paulswartz: This will only get picked up at build time, so it would end up being the same values for Dev and Prod. Instead, you can override these in `config/prod.exs` with "${SECRET_KEY_BASE}" and they'll be loaded at startup.

PR #96, opened 2019-08-02
Server query improvement

paulswartz:
This adds a new type of query to `State.Server`, where we can build more efficient queries by using more features of ETS/Mnesia. There's a lot of code unfortunately, but each commit is self-contained.

### DIFF
--- a/apps/state/lib/state/server/query.ex
+++ b/apps/state/lib/state/server/query.ex
@@ -1,0 +1,38 @@
+defmodule State.Server.Query do
+  @moduledoc """
+  Queries a Server given a map of values.
+
+  Given a Server module and a map of key -> [value], return the values from
+  the Server where each of the keys has one of the values from the given
+  list. A key without a provided list can have any value.
+
+  An example query for schedules would look like:
+
+  %{
+     trip_id: ["1", "2"],
+     stop_id: ["3", "4"]
+  }
+
+  And the results would be the schedules on trip 1 or 2, stopping at stop 3 or 4.
+  """
+
+  @type q :: map
+  @type recordable :: struct
+  @type index :: atom
+
+  alias State.Server
+
+  @spec query(module, q) :: [recordable] when q: map, recordable: struct
+  def query(module, %{} = q) when is_atom(module) do
+    do_query(module, q)
+  end
+
+  defp do_query(module, q) when map_size(q) == 1 do
+    [{index, values}] = Map.to_list(q)
+    Server.by_index(values, module, {index, module.key_index}, [])
+  end
+
+  defp do_query(module, _q) do
+    module.all()
+  end
+end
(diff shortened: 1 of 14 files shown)


apps/state/lib/state/server/query.ex
```diff
@@ -0,0 +1,182 @@
...
+        selectors =
+          for record <- records_from_struct_and_values(struct, index, index_values) do
+            {
+              record,
+              guards,
+              [:"$_"]
+            }
+          end
+
+        %{result | selectors: [selectors | result.selectors]}
```
handorff: Should this be `[selectors ++ result.selectors]`?

apps/state/lib/state/server/query.ex
```diff
@@ -0,0 +1,182 @@
...
+
+  alias __MODULE__.Result
+
+  defp do_query(module, qs) do
+    qs
+    |> Enum.reduce(%Result{module: module}, &accumulate/2)
+    |> finalize()
+  end
+
+  def accumulate(q, %Result{module: module} = result) when map_size(q) > 0 do
```
handorff: Is this function returning the union of queries when it should be returning the intersection?
e.g. if `q = %{a: [1, 2], b: [3, 4]}` we want to find matches where `(a in [1, 2]) AND (b in [3, 4])`. It seems like we're always appending each individual result to the list of matches or selectors.

paulswartz: It's the union of intersections, so this function handles one of the intersections, and adds it to the growing union. For example, in State.Stop:
```elixir
  def location_type_0_ids_by_parent_ids(ids) do
    [
      %{
        id: ids,
        location_type: [0]
      },
      %{
        parent_station: ids,
        location_type: [0]
      }
    ]
    |> State.Stop.Cache.query()
    |> Enum.map(& &1.id)
  end
```

This returns `(id IN ids AND location_type = 0) OR (parent_station IN ids AND location_type = 0)`.

apps/state/lib/state/server/query.ex
```diff
@@ -0,0 +1,182 @@
...
+      {true, :a}
+
+      iex> first_index([:a, :b], %{b: 2})
+      {true, :b}
+
+      iex> first_index([:a, :b], %{c: 3})
+      {false, :c}
+  """
+  @spec first_index([index, ...], q) :: {boolean, index}
+  def first_index(indices, q) do
```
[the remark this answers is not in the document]
paulswartz: They should already be ordered: earlier we'd gone through and tried to put the more useful indicies earlier in the list, and the primary key (or closest) at the front.

apps/state/lib/state/server/query.ex
paulswartz: No, they get flattened in `finalize`:

```elixir
  def finalize(%Result{module: module, selectors: selectors, matches: matches}) do
    selected = Server.select_with_selectors(module, List.flatten(selectors))
    flat_results([selected | matches])
  end
```

paulswartz:
This doesn't improve the performance in practice, and queries aren't really our bottleneck at the moment. #111 is still useful, so I'll clean that up.

PR #97, opened 2019-08-05
fix: allow content-type as a header

paulswartz:
This seems to come with `fetch()` requests from JavaScript apps, and it's
fine to include.


paulswartz:
`corsica` (our upstream library) already has tests for this, so having our own is overkill.

PR #99, opened 2019-08-06
Show route patterns from multi route trips

handorff:


apps/state/lib/state/route_pattern.ex
```diff
@@ -32,45 +33,49 @@ defmodule State.RoutePattern do
...
-    all()
+  defp ids_from_routes(%{route_ids: route_ids} = filters) do
+    opts =
+      case filters do
+        %{direction_id: direction_id} -> %{routes: route_ids, direction_id: direction_id}
+        _ -> %{routes: route_ids}
+      end
+
+    opts
+    |> Trip.filter_by()
```
paulswartz: can you pass `filters` directly to `Trip.filter_by`?

handorff: I believe `State.Trip` expects the key `routes` whereas `State.RoutePatterns` expects `route_ids`.

PR #100, opened 2019-08-09
Show shapes from multi route trips

handorff:


apps/state/lib/state/trip/added.ex
```diff
@@ -59,11 +59,12 @@ defmodule State.Trip.Added do
...
-      Enum.find_value(
-        State.Shape.select_routes([prediction.route_id], prediction.direction_id),
-        &last_stop_on_shape(&1, prediction, stop)
-      )
+    shapes =
+      [prediction.route_id]
+      |> State.Shape.select_routes(prediction.direction_id)
+      |> Enum.filter(&(&1.route_id == prediction.route_id))
+
+    last_stop_id = Enum.find_value(shapes, &last_stop_on_shape(&1, prediction, stop))
```
lboyarsky: It seems that we don't need this newly introduced `shapes` variable, we can calculate `last_stop_id` in a single pass, combining the conditions from `filter` and `find_value`, can't we?

apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
```diff
@@ -270,6 +270,7 @@ defmodule StateMediator.Integration.GtfsTest do
       assert [_] =
                ["CR-Lowell"]
                |> State.Shape.select_routes(1)
+               |> Enum.filter(&(&1.route_id == "CR-Lowell"))
```
lboyarsky: Why would it return shapes with other `route_id`'s? I'd expect this filter to be a noop, is it not the case? May be I'm missing something

paulswartz: There are CR-Haverhill/CR-Lowell multi-route trips, so this includes them both.

@handorff Maybe we should include that as a part of the tests for those multi-route trips?

apps/state/lib/state/shape.ex
```diff
@@ -24,18 +24,19 @@ defmodule State.Shape do
...
+      case direction_id do
+        id when id in [0, 1] -> %{routes: route_ids, direction_id: id}
+        _ -> %{routes: route_ids}
       end
 
-    do_select(matchers)
+    opts
+    |> State.Trip.filter_by()
+    |> Enum.map(& &1.shape_id)
+    |> Enum.uniq()
```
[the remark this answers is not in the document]
paulswartz: If you're looking them up this way, do we still need the index on `route_id`?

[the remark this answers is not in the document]
handorff: I don't think `State.Shape.by_ids` can take a `MapSet` so I'd need to convert back anyways.

I'll remove the `route_id` index.

PR #107, opened 2019-09-04
Change shapes to use full route_pattern name on latest API version

handorff:
- Store the full route_pattern name as the shape name, not just the destination.
- When clients on older versions query shape names, modify the names to remove the origins before sending the response.

### DIFF
--- a/apps/api_web/lib/api_web/views/shape_view.ex
+++ b/apps/api_web/lib/api_web/views/shape_view.ex
@@ -26,4 +26,15 @@ defmodule ApiWeb.ShapeView do
     |> StopsOnRoute.by_route_id(shape_ids: [id])
     |> Stop.by_ids()
   end
+
+  def name(shape, conn) do
+    if conn.assigns.api_version < "2019-07-01" do
+      case String.split(shape.name, " - ", parts: 2) do
+        [_, name] -> name
+        [name] -> name
+      end
+    else
+      shape.name
+    end
+  end
 end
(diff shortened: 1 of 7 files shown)


apps/state/lib/state/shape.ex
```diff
@@ -24,29 +24,53 @@ defmodule State.Shape do
...
-    |> Enum.sort_by(&{-&1.priority, &1.name})
+    shapes =
+      opts
+      |> State.Trip.filter_by()
+      |> Enum.map(& &1.shape_id)
+      |> Enum.uniq()
+      |> by_ids()
+      |> Enum.sort_by(&{-&1.priority, &1.name})
+
+    if api_version < "2019-07-01" do
```
paulswartz: The `state` module doesn't know anything about the API version, and it's better to keep it that way.

Instead, you can override `def name(shape, conn)` in ShapeView, and get the API version there.

apps/api_web/lib/api_web/views/shape_view.ex
```diff
@@ -26,4 +26,15 @@ defmodule ApiWeb.ShapeView do
     |> StopsOnRoute.by_route_id(shape_ids: [id])
     |> Stop.by_ids()
   end
+
+  def name(shape, conn) do
```
paulswartz: You can do this with a pattern match:
```
def name(%{name: name}, %{assigns: %{api_version: version}}) when version >= "2019-07-01" do
  name
end
def name(%{name: name}, _conn) do
  # split
end
```

PR #108, opened 2019-09-05
Switch to using enhanced VehiclePositions JSON

handorff:
Replaces `VehiclePositions.pb` with `VehiclePositions_enhanced.json` and updates parser accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ In addition to the Elixir config files, the V3 API allows runtime configuration 
 | `MBTA_GTFS_URL` | `https://cdn.mbta.com/MBTA_GTFS.zip` | URL for the GTFS .zip file. |
 | `ALERTS_URL` | `https://cdn.mbta.com/realtime/Alerts_enhanced.json` | URL for the Alerts. Can be either a JSON or Protobuf file.
 | `MBTA_TRIP_SOURCE` | `https://cdn.mbta.com/realtime/TripUpdates_enhanced.json` | URL for the TripUpdates. Can be either a JSON or Protobuf file. |
-| `MBTA_VEHICLE_SOURCE ` | `https://cdn.mbta.com/realtime/VehiclePositions.pb` | URL for the VehiclePositions. Must be a Protobuf file. |
+| `MBTA_VEHICLE_SOURCE ` | `https://cdn.mbta.com/realtime/VehiclePositions_enhanced.json` | URL for the VehiclePositions. Must be a JSON file. |
 | `IBM_PARKING_SOURCE` | undefined | URL for parking information. |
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -30,7 +30,7 @@ In addition to the Elixir config files, the V3 API allows runtime configuration
 | `MBTA_GTFS_URL` | `https://cdn.mbta.com/MBTA_GTFS.zip` | URL for the GTFS .zip file. |
 | `ALERTS_URL` | `https://cdn.mbta.com/realtime/Alerts_enhanced.json` | URL for the Alerts. Can be either a JSON or Protobuf file.
 | `MBTA_TRIP_SOURCE` | `https://cdn.mbta.com/realtime/TripUpdates_enhanced.json` | URL for the TripUpdates. Can be either a JSON or Protobuf file. |
-| `MBTA_VEHICLE_SOURCE ` | `https://cdn.mbta.com/realtime/VehiclePositions.pb` | URL for the VehiclePositions. Must be a Protobuf file. |
+| `MBTA_VEHICLE_SOURCE ` | `https://cdn.mbta.com/realtime/VehiclePositions_enhanced.json` | URL for the VehiclePositions. Must be a JSON file. |
```
paulswartz: Can we support both? Otherwise it'll be hard/impossible to deploy since we can't do a configuration change and a code change at the same time.

paulswartz: What we've done in the past is pattern match the incoming data:
```
def parse(body)
def parse("{" <> _ = body), do: parse_json(body)
def parse(body), do: parse_protobuf(body)
```

apps/parse/lib/parse/vehicle_positions.ex
```diff
@@ -1,38 +1,54 @@
...
-  def parse(blob) do
-    blob
-    |> FeedMessage.decode()
-    |> (fn message -> message.entity end).()
-    |> Stream.map(fn entity -> entity.vehicle end)
-    |> Stream.map(&parse_vehicle_update/1)
+  @impl true
+  def parse(body) do
+    body
+    |> Jason.decode!()
```
paulswartz: If you give Jason the `strings: :copy` option (see https://hexdocs.pm/jason/Jason.html#decode/2-options), then you don't need to use `optional_field_copy` elsewhere in the module.

apps/parse/test/parse/vehicle_positions_json_test.ex
```diff
@@ -0,0 +1,94 @@
+defmodule Parse.VehiclePositionsJsonTest do
```
paulswartz: This isn't picked up as a test: it should be renamed to `.exs`.

PR #115, opened 2019-09-23
fix: add additional buffers for Nginx

paulswartz:
We added these to the website to handle incoming requests; I hope that
they'll address the small number of 504 Gateway timeouts we're seeing.


paulswartz:
We ended up pulling Nginx out of the set of servers.

PR #117, opened 2019-09-25
fix: use dumb-init for running the server

paulswartz:
From https://github.com/Yelp/dumb-init:

> dumb-init is a simple process supervisor and init system designed to run as
> PID 1 inside minimal container environments (such as Docker).

It properly forwards signals to the child process. I think this may be
related to issues we're seeing when we're terminating instances in Elastic
Beanstalk.

### DIFF
--- a/rel/Dockerfile
+++ b/rel/Dockerfile
@@ -2,7 +2,7 @@
 FROM debian:stretch
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-		libssl1.1 libsctp1 curl \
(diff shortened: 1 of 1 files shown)


paulswartz:
Tested yesterday/today, and it seems to have addressed the 504 Gateway Timeout issues we were seeing in the morning as instances were terminated.

lboyarsky:
BTW https://github.com/Yelp/dumb-init suggests to use it in `ENTRYPOINT` command rather than in `CMD`. Other option they mention is to use it in shebang.

PR #120, opened 2019-09-27
add overrides for new Foxboro service

paulswartz:
Pulled out of #114, but without the integration test changes. These should pass the existing tests, but have the right output with the new GTFS data.


paulswartz:
The one weird thing is that for the current schedule, it adds Dedham Corp Center to Fairmount inbound: https://green.dev.mbtace.com/schedules/CR-Fairmount/timetable?direction_id=1

PR #122, opened 2019-10-01
Update subway route pattern names in the tests

lboyarsky:
**Asana Ticket:** [give Subway route patterns an origin in addition to the destination](https://app.asana.com/0/810933294009540/1140697116556421/f)

Comes w/ https://github.com/mbta/gtfs_creator/pull/550

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -55,10 +55,9 @@ config :state, :shape,
     # Order the Red Line Ashmont first, and change the northbound names to
     # the names of the branch.
     "931_0009" => 2,
-    "931_0010" => {2, "Ashmont"},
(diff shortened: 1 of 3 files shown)


paulswartz:
There are also some overrides in `apps/state/config/config.exs` which should be updated.

lboyarsky:
Updated. Am I right that we don't use shape `934_0001` any more (I removed it from overrides)?

lboyarsky:
@paulswartz I think it's good to merge? Integration tests won't pass until corresponding gtfs_creator change is merged, and I cannot merge it until we have this API change merged so I can revert `API_USER` and `API_BRANCH` to defaults in the makefile there.

PR #124, opened 2019-10-08
Vehicle: label query checks all labels in consist

handorff:
- Parse labels from consist and store them as a list with the vehicle.
- Replace label `build_filters` with `do_post_search_filter` which matches individual labels in the consist as well as the label on the (entire) vehicle.

### DIFF
--- a/apps/model/lib/model/vehicle.ex
+++ b/apps/model/lib/model/vehicle.ex
@@ -17,7 +17,8 @@ defmodule Model.Vehicle do
     :current_status,
     :current_stop_sequence,
     :updated_at,
-    :effective_route_id
+    :effective_route_id,
+    :consist
   ]
 
   alias Model.WGS84
@@ -59,8 +60,9 @@ defmodule Model.Vehicle do
   * `:effective_route_id` - The `Model.Route.id` of the `Model.Route.t` that the vehicle is _currently_ on.  When
       `trip_id` has only one route, then `effective_route_id` and `route_id` will always match.  If `trip_id` has
       multiple routes, then `effective_route_id` can be any of the routes of `trip_id`.
-  * `:label` - User visible label, such as the one of on the signage on the vehicle.  See
+  * `:label` - User visible label, such as the one on the signage on the vehicle.  See
       [GTFS-realtime VehicleDescriptor label](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-vehicledescriptor).
+  * `:consist` - Set of user visible labels (such as the one on the signage on the vehicle) on individual cars. Only present for light and heavy rail.
   * `:updated_at` - Time at which vehicle information was last updated.
   * `:latitude` - Latitude of the vehicle's current position.  See
       [GTFS-realtime Position latitude](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-position).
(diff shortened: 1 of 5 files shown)


apps/parse/lib/parse/vehicle_positions_json.ex
```diff
@@ -47,6 +48,14 @@ defmodule Parse.VehiclePositionsJson do
     []
   end
 
+  defp parse_consist([_ | _] = consist) do
+    consist
+    |> Enum.map(fn %{"label" => car_label} -> car_label end)
+    |> MapSet.new()
```
paulswartz: For using this data in the future, we probably want to keep it as a list: the order of the labels is the order of the cars.

apps/state/lib/state/vehicle.ex
```diff
@@ -66,12 +66,41 @@ defmodule State.Vehicle do
...
+      all()
+      |> Enum.filter(fn vehicle ->
+        label_matches?.(vehicle, labels) or consist_matches?.(vehicle, labels)
+      end)
+      |> Enum.map(& &1.id)
+
+    for matcher <- matchers, vehicle_id <- vehicle_ids do
+      Map.put(matcher, :id, vehicle_id)
+    end
+  end
```
paulswartz: If you want to go this route, it would be better to do this as a post-filter. What you're doing here is querying all the vehicles, filtering them, then using the IDs to query the vehicles. But you already had the vehicles after the second step, so the extra ID filter is wasted work.

Other options would be to do either something like what we do with Alerts (separate tables for querying parts of the data) or Trips (extra fields to handle the fact that a vehicle can have multiple labels.

apps/model/lib/model/vehicle.ex
```diff
@@ -88,7 +90,8 @@ defmodule Model.Vehicle do
           route_id: Model.Route.id() | nil,
           speed: speed | nil,
           stop_id: Model.Stop.id() | nil,
-          trip_id: Model.Trip.id() | nil
+          trip_id: Model.Trip.id() | nil,
+          consist: MapSet.t(String.t()) | nil
```
paulswartz: `[String.t()] | nil`

apps/parse/lib/parse/vehicle_positions_json.ex
```diff
@@ -47,6 +48,13 @@ defmodule Parse.VehiclePositionsJson do
     []
   end
 
+  defp parse_consist([_ | _] = consist) do
+    consist
+    |> Enum.map(fn %{"label" => car_label} -> car_label end)
```
paulswartz: I'm surprised this doesn't get caught by Credo: shouldn't use a pipe for a single function call.

apps/state/lib/state/vehicle.ex
```diff
@@ -101,7 +99,32 @@ defmodule State.Vehicle do
...
+      case consist do
+        nil ->
+          false
+
+        _ ->
+          consist
+          |> MapSet.new()
+          |> MapSet.intersection(labels)
+          |> MapSet.size()
+          |> Kernel.!==(0)
```
paulswartz: `not MapSet.disjoint?(labels, MapSet.new(consist))`

apps/state/lib/state/vehicle.ex
```diff
@@ -101,7 +99,32 @@ defmodule State.Vehicle do
...
+      end
+    end
+
+    label_matches? = fn %Model.Vehicle{label: label}, labels ->
+      label in labels
+    end
+
+    Enum.filter(vehicles, fn vehicle ->
+      label_matches?.(vehicle, labels) or consist_matches?.(vehicle, labels)
+    end)
```
paulswartz: Since you're building the match functions here anyways, you don't need to pass in `labels`. It's available from the higher scope.

PR #125, opened 2019-10-09
Add start_date and end_date to /status

handorff:
Move `feed_version` into `feed` as `version` and add `start_date` and `end_date`.

### DIFF
--- a/apps/state/lib/state/metadata.ex
+++ b/apps/state/lib/state/metadata.ex
@@ -46,10 +46,10 @@ defmodule State.Metadata do
   end
 
   @doc """
-  Sets latest version of the Feed.
+  Sets latest version, start_date and end_date of the Feed.
   """
-  def feed_updated(version) do
-    :ets.insert(table_name(), {State.Feed, version})
+  def feed_updated(version, start_date, end_date) do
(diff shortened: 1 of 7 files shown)


apps/state/test/state/metadata_test.exs
```diff
@@ -109,13 +129,24 @@ defmodule State.MetadataTest do
   @doc """
   Caches a value.
   """
+  def cache(key, {version, start_date, end_date}) do
```
paulswartz: Can you insert the `{version, start_date, end_date}` tuple as the value?

apps/state/lib/state/metadata.ex
```diff
@@ -46,10 +46,10 @@ defmodule State.Metadata do
   end
 
   @doc """
-  Sets latest version of the Feed.
+  Sets latest version, start_date and end_date of the Feed.
   """
-  def feed_updated(version) do
-    :ets.insert(table_name(), {State.Feed, version})
+  def feed_updated(version, start_date, end_date) do
```
paulswartz: If you pass the tuple in here, you don't need to do the unwrapping in `feed_metadata/0`.

PR #127, opened 2019-10-21
fix: treat empty consist as nil

paulswartz:
As we get the consist information as a separate message from the rest of the
vehicle information, it's possible to have a vehicle without a consist. Treat
that empty list the same as if we were not provided a consist at all.


apps/parse/test/parse/vehicle_positions_json_test.exs
```diff
@@ -130,5 +130,32 @@ defmodule Parse.VehiclePositionsJsonTest do
       [%{consist: consist}] = parse_entity(entity)
       assert consist == ["1877", "1876", "1854", "1855", "1833", "1832"]
     end
+
+    test "handles an empty cconsist" do
```
handorff: typo

PR #138, opened 2019-10-28
add ability to clone an existing key

paulswartz:
In order to more easily rotate API keys for internal applications, this adds an administrative feature to create a clone of an existing key. The cloned keys shares the same attributes (rate limit, description, CORS) as the original key.


paulswartz:
I like that idea; I'll work on it today.

PR #158, opened 2019-11-22
fix(Schedule): don't filter out ignored trips when querying stops/trips

paulswartz:
Since some of the trips may be ignored in the general case, we still want to
include them when we're asking for trips and stops together. For example,
trips which are in `multi_route_trips.txt` should also appear in the schedule queries.

Compare: `/schedules/?filter[route]=746&filter[direction_id]=0&filter[stop]=place-sstat` on Green (has data, good) vs. Dev (no data, bad)


apps/state/lib/state/routes_patterns_at_stop.ex
```diff
@@ -48,11 +48,12 @@ defmodule State.RoutesPatternsAtStop do
 
   def routes_by_stops_and_direction(stop_ids, opts \\ []) when is_list(stop_ids) do
     direction_id = Keyword.get(opts, :direction_id, :_)
+    ignore? = if Keyword.get(opts, :ignore?, true) == false, do: :_, else: true
```
handorff: Why the `== false` in addition to the default value? In case someone passes in a non-boolean value for `ignore?`?

paulswartz: The default behavior is the same as passing `ignore?: true` as an option. The new feature is to support `ignore: false` which includes both sets of stop IDs.

handorff: I just meant the default value in `Keyword.get/3`.

e.g. `ignore? = if Keyword.get(opts, :ignore?, true), do: true, else: :_`

paulswartz: Ah, gotcha: fixed.

PR #164, opened 2019-12-11
Add rating start/end dates, description to API

handorff:
Add new `rating_start_date`, `rating_end_date` and `rating_description` attributes from GTFS `calendar_attributes.txt` to the associated service in the API.


apps/api_web/lib/api_web/controllers/service_controller.ex
```diff
@@ -163,6 +163,29 @@ defmodule ApiWeb.ServiceController do
               enum: Enum.to_list(0..5),
               example: 1
             )
+
+            rating_start_date(
+              :string,
+              "Earliest date which is a part of the rating (season) which contains this service. Format is ISO8601.",
+              "x-nullable": true,
```
paulswartz: Is this actually nullable? I thought they all had start dates?

apps/parse/lib/parse/calendar_attributes.ex
```diff
@@ -28,13 +34,26 @@ defmodule Parse.CalendarAttributes do
...
   defp copy_int(s), do: String.to_integer(s)
+
+  defp copy_date(nil), do: nil
+  defp copy_date(""), do: nil
+
+  defp copy_date(s) do
+    s
+    |> Timex.parse!("{YYYY}{0M}{0D}")
+    |> NaiveDateTime.to_date()
+  end
```
paulswartz: Can we do this without adding a new dependency on Timex?
```elixir
defp copy_date(<< year_bin::binary-4, month_bin::binary-2, day_bin::binary-2 >>) do
  Date.new(String.to_integer(year_bin), String.to_integer(month_bin), String.to_integer(day_bin))
end
```

apps/api_web/lib/api_web/controllers/service_controller.ex
handorff: Currently Foxboro services don't, but we could add one if it made sense.

Separately, it looks like services created by disruptions aren't getting rating columns filled in. I'll file a ticket to fix that.

apps/parse/lib/parse/calendar_attributes.ex
handorff: We're currently using Timex to parse similarly formatted dates in at least `calendar.txt`, `calendar_dates.txt` and `feed_info.txt` -- should I change all of them?

paulswartz: No, nevermind; I thought we'd done more work to remove Timex but must have misremembered.

apps/api_web/lib/api_web/controllers/service_controller.ex
paulswartz: Fair enough, re: Foxboro. I retract my comment.

PR #182, opened 2020-01-31
test: ensure that RoutePatternsAtStop is updated

paulswartz:
Since filtering by trip/stop uses RoutePatternsAtStop, we need to ensure that
it's been updated before querying in the test cases.


paulswartz:
Intermittent. If the `RoutePatternsAtStop` updated automatically fast enough, then they'll pass, which tends to work on a local machine. On CI however, the slowdown can cause failures like https://semaphoreci.com/mbta/api/branches/master/builds/190.

PR #185, opened 2020-02-04
feat: deploy via ECS/Dockerrun

paulswartz:
Switches the API deployment to match Arrow and other EB applications: build a Docker image, push it to ECS, and create a Dockerrun.aws.json file pointing at ECS.


Dockerfile
```diff
@@ -23,5 +8,36 @@ RUN mix local.hex --force && \
...
+ADD mix.* /root/
+
+RUN mix do deps.get --only prod, phx.swagger.generate, compile, phx.digest
+
+ADD rel/ rel/
+
+RUN mix distillery.release --verbose
+
+# The one the elixir image was built with
+FROM debian:stretch
```
losvedir: I think [`elixir:1.9.1`](https://github.com/c0b/docker-elixir/commit/8c92148d07fc89a57706dcb22c926a96e73a4194#diff-51c95e8cd4f39fd0a0428238d2574e74) pulls from just `erlang:22` (in contrast to what used to be in this file, `erlang:22.0`), which nowadays is built on [`debian:buster`](https://github.com/erlang/docker-erlang-otp/blob/636231974186645b6c96c6c888e61fd086394c3a/22/Dockerfile).

I'm not sure what happens if the two build stages are inconsistent. I vaguely remember copying over `stretch` when building Arrow before getting a crash and realizing it had changed.

build.sh
```diff
@@ -1,25 +1,25 @@
...
-git checkout -- bin
+test -d $githash && rm -r $githash
+mkdir $githash
+sed -e "s/DOCKER_REPO/$repo/g" -e "s/GITHASH/$githash/g" ./semaphore/Dockerrun.aws.json > $githash/Dockerrun.aws.json
+cp -r rel/.ebextensions $githash
+pushd $githash
+zip -r ../$BUILD_ARTIFACT .
 popd
+
+rm -r $githash
```
losvedir: I was confused at first since the script here seemed to end prematurely, but then after poking around in semaphore I see that there's actually a fair bit of bash scripting there.

Does it make sense as part of this deployment overhaul to pull in some of that code to version control, either as part of this script, or another script `deploy.sh`? (Arrow, e.g., has two scripts, `build_push.sh`, which is the same as the one you added here, and `deploy.sh`, which is basically this file plus a lot of the stuff that is in semaphore right now.)

Dockerfile
paulswartz: thanks, good catch!

build.sh
paulswartz: In general, yes. However, in this case I wanted to be able to do this branch without needing to make changes to the Semaphore build/deploy configuration.

Dockerfile
paulswartz: Actually, it looks like this is Stretch (and the builds are failing with Buster).

- `elixir:1.9.1` was built against `erlang:22.1.3`: https://hub.docker.com/layers/elixir/library/elixir/1.9.1/images/sha256-ab9c60fe6cdafc15fead89cbf99c2950aaec6fb94d8b0694d867eca372e6dcc3
- `erlang:22.1.3` was built against `buildpack-deps:stretch`: https://github.com/erlang/docker-erlang-otp/commit/b5f9e48145117d414c1475216a3e0e3bc550c664#diff-b6d767d2f8ed5d21a44b0e5886680cb9

losvedir: Ah, interesting. So I guess even though the dockerfile at the time simply specified `erlang 22` (as far as I can tell from the GitHub history), it has been forever snapshotted on the docker hub at `22.1.3`?

How are you getting to `erlang:22.1.3` from that link, btw? I see the ENV var which has that `OTP_VERSION` but I'm not seeing a jump to a corresponding docker image. I'm assuming it has to do with the `ADD file` at the top, but I'm coming up short trying to trace that hash.

paulswartz: I assumed it based on the OTP_VERSION: I couldn't find a way back from that hash.

paulswartz: Also based on the builds crashing when I tried to use `buster` :)

losvedir: Haha, fair enough! Just wondering if I was missing something.

PR #186, opened 2020-02-11
refactor: switch email service from Mailgun to AWS SES

fjlanasa:
Asana ticket: [🍎 switch API e-mails from Mailgun to Amazon SES](https://app.asana.com/0/584764604969369/1160343341012041/f)

Just changed the `ApiAccounts.Mailer` adapter to `Bamboo.SesAdapter` which required installing the `bamboo_ses` adapter

TODO:
- [ ] ensure the necessary AWS users get the correct access


apps/api_accounts/mix.exs
```diff
@@ -56,7 +56,8 @@ defmodule ApiAccounts.Mixfile do
      {:comeonin, "~> 5.1"},
      {:bcrypt_elixir, "~> 2.0"},
      {:uuid, "~> 1.1"},
-     {:bamboo, github: "mbta/bamboo", ref: "6ae35de"}]
+     {:bamboo, github: "mbta/bamboo", ref: "6ae35de", override: true},
```
fjlanasa: @paulswartz I had to set `override: true` here because this conflicts with the dependencies of `bamboo_ses`. Does this seem like the right approach?

apps/api_web/lib/api_web/controllers/exceptions.ex
```diff
@@ -1,3 +1,5 @@
 defimpl Plug.Exception, for: ApiAccounts.NoResultsError do
   def status(_expection), do: 404
+
+  def actions(_exception), do: []
```
fjlanasa: Plug was bumped from `1.83` to `1.9`, which required implementing this function

apps/api_accounts/mix.exs
paulswartz: what's in our fork of `bamboo`? I don't recall what we changed.

fjlanasa: https://github.com/thoughtbot/bamboo/compare/master...mbta:master

paulswartz: That seems like we can move back to whatever the hex-released version of Bamboo is.

fjlanasa: wil do

PR #192, opened 2020-02-25
improvements to /schedule/ API performance

paulswartz:


apps/state/lib/state/schedule.ex
```diff
@@ -77,8 +79,9 @@ defmodule State.Schedule do
   """
   @spec filter_by(filter_opts) :: [Schedule.t()]
   def filter_by(filters) do
+    filters = convert_filters(filters)
```
lemald: Is there some performance advantage to taking this out of the pipeline?

apps/api_web/mix.exs
```diff
@@ -60,7 +60,7 @@ defmodule ApiWeb.Mixfile do
     [
       {:phoenix, "< 1.4.15"},
       {:phoenix_html, "~> 2.11"},
-      {:ja_serializer, github: "mbta/ja_serializer", branch: "master"},
+      {:ja_serializer, github: "paulswartz/ja_serializer", branch: "mbta"},
```
lemald: Just to clarify, I assume you're undoing this once the branch in `ja_serializer` gets merged?

paulswartz: correct

apps/state/lib/state/schedule.ex
lemald: Ah okay, now I see why you did this.

paulswartz: I changed the name to make this more clear. What's changing is that we're now passing the converted filters into the downstream functions, instead of the original filters. Since `convert_filters` turns a route and a date into a list of trips, we can remove the `date` filter from downstream checks to avoid the extra date check on [277](https://github.com/mbta/api/pull/192/files/b9e39a9705c85db092cae64c5f3274d233ebf6ce#diff-b43743fe0d7e9f0401b48e3202eb903dR277).

PR #194, opened 2020-02-26
feat: Handle new CR hash-based route patterns

losvedir:
The new GTFS files will now have route patterns for the commuter rail generated based on the stops they visit and whether they're flag stops or leave early stops. Consequently, it's now possible for multiple trips with the same *shape* to have different *route patterns*, when previously there was a 1-to-1 correspondence between shape and route pattern.


losvedir:
Yep, my `gtfs_creator` branch passes when using this branch of the API locally.

PR #204, opened 2020-03-13
Move shape prioritization to use starts_with

jfabi:
In support of https://github.com/mbta/gtfs_creator/pull/763.

Checks for `shape_id`s _starting with_ the IDs listed in apps/state/config/config.exs, rather than requiring an exact match. This is so that new shapes that are generated from Disruptions will be properly prioritized.

(It's been a while since I've written any Elixir, so hopefully this all does what I think it does, particularly with the testing.)

### DIFF
--- a/apps/state/lib/state/shape.ex
+++ b/apps/state/lib/state/shape.ex
@@ -203,12 +203,12 @@ defmodule State.Shape do
 
   defp override_priorities_from_configuration(shapes) do
     config = Application.fetch_env!(:state, :shape)
-    overrides = config[:overrides]
+    prefix_overrides = config[:prefix_overrides]
     suffix_overrides = config[:suffix_overrides]
 
     shapes
     |> Enum.map(fn %{id: id} = shape ->
-      case find_override(id, overrides, suffix_overrides) do
+      case find_override(id, prefix_overrides, suffix_overrides) do
         nil ->
           shape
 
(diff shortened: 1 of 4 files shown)


apps/state/lib/state/shape.ex
```diff
@@ -222,16 +222,18 @@ defmodule State.Shape do
...
+      if String.starts_with?(id, shape_id) do
+        priority
+      else
+        Enum.find_value(suffix_overrides, fn {suffix, priority} ->
+          if String.ends_with?(id, suffix) do
+            priority
+          end
+        end)
+      end
+    end)
```
paulswartz: This is going to loop over all the suffix overrides for each prefix override. You could do something like:
```elixir
Enum.find_value(
  [{&String.starts_with?/2, prefix_overrides}, 
   {&String.ends_with?/2, suffix_overrides}], 
  fn {check_fn, overrides} ->
    Enum.find_value(overrides, fn {check_value, priority} ->
      if check_fn.(id, check_value), do: priority
    end)
  end)
```
which will check all the prefixes, then all the suffixes.

apps/state/lib/state/shape.ex
```diff
@@ -222,16 +222,14 @@ defmodule State.Shape do
...
-    if priority = Map.get(overrides, id) do
-      priority
-    else
-      Enum.find_value(suffix_overrides, fn {suffix, priority} ->
-        if String.ends_with?(id, suffix) do
-          priority
-        end
+  defp find_override(id, prefix_overrides, suffix_overrides) do
+    [{&String.starts_with?/2, prefix_overrides}, {&String.ends_with?/2, suffix_overrides}]
+    |> Enum.find_value(fn {check_fn, overrides} ->
```
paulswartz: We try to avoid using a single pipe, per the Elixir core teams style guide.

jfabi:
Currently getting an error `no function clause matching in String.starts_with?/2` after splitting apart part of it to pass Credo and Dialyzer.

See https://semaphoreci.com/mbta/api/branches/jf-gtfs-green-d/builds/8.

apps/state/lib/state/shape.ex
```diff
@@ -222,16 +222,19 @@ defmodule State.Shape do
...
+  defp find_override(id, prefix_overrides, suffix_overrides) do
+    Enum.find_value(
+      [{&String.starts_with?/2, prefix_overrides}, {&String.ends_with?/2, suffix_overrides}],
+      fn {check_fn, overrides} ->
+        find_override_priority(check_fn, overrides, id)
+      end
+    )
+  end
+
+  defp find_override_priority(overrides, check_fn, id) do
```
paulswartz: You flipped the arguments. You call it as "(check_fn, overrides, id)", but it's defined as "(overrides, check_fn, id)".

paulswartz:
```suggestion
  defp find_override_priority(check_fn, overrides, id) when is_function(check_fn, 2) do
```

jfabi: Give that a shot. Thank you...!

PR #212, opened 2020-03-25
speed up Trip.Added

paulswartz:
It appears that the issue the website is seeing with trips being returned as `null` is a race condition between when the predictions are available in `State.Prediction` and when the new trip is created in `State.Trip.Added`.
This PR makes three changes to address those issues:

1) adds a `trip_match?` flag to `Model.Prediction` which `State.Trip.Added` can use to only check predictions which we know didn't match a trip in GTFS.
2) Uses the `route_pattern_id` if it exists on the Prediction to shortcut all the complicated shape matching.
3) When including trips, adds a fallback to calculate the Added trip based on the appropriate predictions for that specific trip.

On dev-green, these changes take the `State.Trip.Added` processing time from 500-1500ms down to 10-25ms, which alone should cut down on the amount of time a trip doesn't exist. If a request does come in that window, it should be created for the request which needs it.

### DIFF
--- a/apps/state/test/state/prediction_test.exs
+++ b/apps/state/test/state/prediction_test.exs
@@ -48,7 +48,7 @@ defmodule State.PredictionTest do
 
     new_state([@prediction])
 
-    assert by_route_id("alternate") == [%{@prediction | route_id: "alternate"}]
+    assert [%{route_id: "alternate"}] = by_route_id("alternate")
   end
 
   describe "filter_by_route_type/2" do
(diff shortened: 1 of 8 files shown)


apps/state/test/state/prediction_test.exs
```diff
@@ -86,6 +86,18 @@ defmodule State.PredictionTest do
 
       assert [%{direction_id: 1}] = pre_insert_hook(prediction)
     end
+
+    test "sets trip_match? to true if there's a trip in GTFS" do
```
lemald: Might also be worth testing that `trip_match?` doesn't get set if there's no trip?

paulswartz: 9463e09

PR #214, opened 2020-04-04
chore: update to Elixir 1.10 & Erlang 22.3

paulswartz:
I was having trouble installing Erlang 22.0.7 on my machine after updating to OS X 10.15.4 and figured I'd take the opportunity to update.

It's currently baking on prod to ensure nothing unusual is happening, but everything was okay when I tested on dev-green yesterday.


Dockerfile
```diff
@@ -19,11 +21,10 @@ ADD rel/ rel/
 RUN mix distillery.release --verbose
 
 # The one the elixir image was built with
-FROM debian:stretch
+FROM alpine:3.11
 
-RUN apt-get update && apt-get install -y --no-install-recommends \
-		libssl1.1 libsctp1 curl dumb-init \
-	&& rm -rf /var/lib/apt/lists/*
+RUN apk add --update libssl1.1 curl bash dumb-init \
```
losvedir, line 26: No more need for `libsctp1`? How did you determine that? @ianwestcott and I had a discussion once about these, and best we could come up with was just some trial and error with different combinations of them to see if the app barfs.

apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
```diff
@@ -416,7 +416,7 @@ defmodule StateMediator.Integration.GtfsTest do
     test "have valid route, trip and stop IDs" do
       for %{route_id: route_id} <- all_routes(),
           prediction <-
-            State.Prediction.select_grouped([State.Prediction], [
+            State.Prediction.select([
```
losvedir, line 419: How was this test passing before it was removed previously?

Dockerfile
paulswartz: Pretty much. I started with only libssl1.1 and dumb-init since I know those are required, and added curl and bash back because the app complained.

apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
paulswartz: We aren't running it in CI :/ It only runs if we provide a source for predictions in the integration tests, and we aren't doing that at the moment. Once it lands I can add it back.

Dockerfile
ianwestcott: This question took me down a fun little rabbit hole which led here: https://bugs.erlang.org/browse/ERL-262?focusedCommentId=11711&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel

TL;DR: The dependency on `libsctp1` was bogus and was caused by a bug in OTP 19.1 that was later fixed. It's possible that we (or even the Debian packages?) still assume it's required, but Alpine doesn't.

PR #217, opened 2020-04-15
feat: remove fields apart from polyline from shape in new version

lemald:
Asana ticket: [🍎 deprecate redundant shape fields](https://app.asana.com/0/295455480314405/1152826632386678/f)


apps/api_web/config/config.exs
```diff
@@ -38,9 +38,10 @@ config :api_web, :versions,
     "2018-07-23",
     "2019-02-12",
     "2019-04-05",
-    "2019-07-01"
+    "2019-07-01",
+    "2020-04-21"
   ],
-  default: "2019-07-01"
+  default: "2020-04-21"
```
paulswartz, line 44: Can we make the new version `2020-05-01`, and not change the default yet? Since anonymous API usage uses the default version, I'd like to give people a little more time to update.

apps/api_web/lib/api_web/views/shape_view.ex
```diff
@@ -2,31 +2,51 @@ defmodule ApiWeb.ShapeView do
...
-  def stops(%{id: id, route_id: route_id}, _conn) do
-    route_id
-    |> StopsOnRoute.by_route_id(shape_ids: [id])
-    |> Stop.by_ids()
+  @impl true
+  def attributes(shape, %{assigns: %{api_version: version}} = conn)
+      when version >= "2020-04-21" do
+    shape
+    |> super(conn)
+    |> Map.take([:polyline])
```
paulswartz, line 21: You can simplify this:
```
%{
  polyline: shape.polyline
}
```

lemald: Without even calling `super`? Or are you just referring to the `Map.take/2` bit?

paulswartz: All `super` really does for us here is `Map.take(struct, keys)`, so if you're building the map yourself it's easier to just do it.

apps/api_web/config/config.exs
lemald: I assume we should also make a followup ticket to bump the default version once 5/1 rolls around?

paulswartz: Once this is on production, people can update their applications without waiting for the default to change.

lemald: Don't we still want to make it the default eventually, though, especially if we want to eventually discourage use of `shapes` for things that are now better done with `route_patterns`?

paulswartz: Sorry, that's what the :+1: was for. We'll land this to give people time to update, and then when the date comes, we'll update the default.

PR #218, opened 2020-04-15
fix: route_controller date filter

fjlanasa:
Asana ticket: [🍎🐛 API filter by date doesn't work without route_id](https://app.asana.com/0/295455480314405/1158758724948945/f)

Updated `route_controller` to allow filtering by only `date` or `date` and `type`, in addition to `stop` and `date`, which is currently enabled.

### DIFF
--- a/apps/api_web/lib/api_web/controllers/route_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/route_controller.ex
@@ -139,14 +139,14 @@ defmodule ApiWeb.RouteController do
   defp do_filter(%{stops: _stops, type: types} = filters) do
     filters
     |> routes_at_stops()
-    |> routes_by_id_and_type(types)
+    |> Enum.flat_map(fn id -> for type <- types, do: %{id: id, type: type} end)
+    |> Route.select()
   end
 
-  defp do_filter(%{service_ids: [], type: _}), do: []
+  defp do_filter(%{service_ids: []}), do: []
 
-  defp do_filter(%{service_ids: [service_id | _], type: types}) do
-    service_id |> RoutesByService.for_service_id() |> routes_by_id_and_type(types)
-  end
+  defp do_filter(%{service_ids: [service_id | _], type: types}),
+    do: RoutesByService.for_service_id_and_types(service_id, types) |> Route.by_ids()
 
   defp do_filter(%{stops: _stops} = filters) do
     filters
@@ -158,20 +158,12 @@ defmodule ApiWeb.RouteController do
     Route.by_types(type)
   end
 
-  defp do_filter(%{service_ids: []}), do: []
-
   defp do_filter(%{service_ids: [service_id | _]}) do
     service_id |> RoutesByService.for_service_id() |> Route.by_ids()
   end
 
   defp do_filter(_filters) do
     Route.all()
-  end
-
-  defp routes_by_id_and_type(route_ids, types) do
-    route_ids
-    |> Enum.flat_map(fn id -> for type <- types, do: %{id: id, type: type} end)
-    |> Route.select()
   end
 
   defp routes_at_stops(%{stops: stops} = filters) do
--- a/apps/api_web/lib/api_web/controllers/route_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/route_controller.ex
@@ -146,7 +146,7 @@ defmodule ApiWeb.RouteController do
   defp do_filter(%{service_ids: []}), do: []
 
   defp do_filter(%{service_ids: [service_id | _], type: types}),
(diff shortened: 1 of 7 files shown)


apps/state/lib/state/routes_by_service.ex
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  @impl GenServer
+  def handle_info(:update, state) do
+    state = update_state(state)
+    {:noreply, state, :hibernate}
+  end
+
+  def update_state(state) do
+    trips = State.Trip.all() ++ State.Trip.Added.all()
```
paulswartz, line 77: Added trips aren't active on a service, so you don't need to include them.

apps/api_web/lib/api_web/controllers/route_controller.ex
```diff
@@ -139,8 +139,13 @@ defmodule ApiWeb.RouteController do
...
     |> routes_at_stops()
-    |> Enum.flat_map(fn id -> for type <- types, do: %{id: id, type: type} end)
-    |> Route.select()
+    |> routes_by_id_and_type(types)
+  end
+
+  defp do_filter(%{service_ids: [], type: _}), do: []
+
+  defp do_filter(%{service_ids: [service_id | _], type: types}) do
+    service_id |> RoutesByService.for_service_id() |> routes_by_id_and_type(types)
```
paulswartz, line 148: Since the `type` is a route type, could you also incorporate that into what `RoutesByService` lets you query? Something like `for_service_id_and_type()`?

apps/state/lib/state/routes_by_service.ex
```diff
@@ -0,0 +1,97 @@
...
+
+  def for_service_id(service_id) do
+    results = :ets.lookup(@table, service_id)
+
+    if Enum.empty?(results) do
+      []
+    else
+      [{_, routes} | _] = results
+      routes
+    end
```
paulswartz, line 29: This is probably better as a `case`:

```elixir
case :ets.lookup(@table, service_id) do
  [{_, routes} | _] ->
    routes
  [] ->
    []
end
```

fjlanasa:
> Is it possible to query by all three things? For example, `?stop=place-alfcl&type=3&date=2020-04-18` to get buses from Alewife on Saturday?

Yep, I added a test case for that.

apps/api_web/lib/api_web/controllers/route_controller.ex
```diff
@@ -143,6 +143,11 @@ defmodule ApiWeb.RouteController do
     |> Route.select()
   end
 
+  defp do_filter(%{service_ids: []}), do: []
+
+  defp do_filter(%{service_ids: [service_id | _], type: types}),
+    do: service_id |> RoutesByService.for_service_id_and_types(types) |> Route.by_ids()
```
paulswartz, line 149: Why only filter by the first service ID, and not all service IDs that are active?

apps/state/lib/state/routes_by_service.ex
```diff
@@ -0,0 +1,114 @@
...
+  def for_service_id(service_id) do
+    @table |> :ets.match_object({{service_id, :_}, :_}) |> do_get_routes()
+  end
+
+  def for_service_id_and_types(service_id, route_types) do
+    @table
+    |> :ets.select(
+      for(
+        type <- route_types,
+        do: {{{:"$0", :"$1"}, :"$2"}, [{:==, :"$0", service_id}, {:==, :"$1", type}], [:"$_"]}
```
paulswartz, line 30: You don't need the guards here: `{{service_id, type}, :_}, [], [:"$_"]}` should work fine.

apps/state/lib/state/routes_by_service.ex
```diff
@@ -0,0 +1,114 @@
...
+
+    items =
+      trips
+      |> Enum.group_by(fn x ->
+        {x.service_id, Map.get(routes, x.route_id, %Model.Route{}).type}
+      end)
+      |> Enum.map(fn {x, y} -> {x, Enum.map(y, fn x -> x.route_id end)} end)
+
+    @table |> :ets.delete_all_objects()
+    @table |> :ets.insert(items)
```
paulswartz, line 102: We generally avoid using `|>` when there's only a single one: `:ets.delete_all_objects(@table)` and `:ets.insert(@table, items)` are fine.

apps/api_web/lib/api_web/controllers/route_controller.ex
fjlanasa: eh i misread the `do_format_filter` that converts `date` to `service_ids`. I originally thought that it ultimately only returned 1 `service_id`. I'll change this to use all and add a test case.

fjlanasa: Should be resolved now

PR #223, opened 2020-04-21
Add `stop_headsign` to Schedule endpoint

lemald:
Asana ticket: [🍎 add headsign to the schedules API](https://app.asana.com/0/295455480314405/1159497818907390/f)

### DIFF
--- a/apps/model/lib/model/schedule.ex
+++ b/apps/model/lib/model/schedule.ex
@@ -16,6 +16,7 @@ defmodule Model.Schedule do
     :arrival_time,
     :departure_time,
     :stop_sequence,
+    :stop_headsign,
     :pickup_type,
     :drop_off_type,
     :position,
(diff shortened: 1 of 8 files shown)


apps/model/lib/model/schedule.ex
```diff
@@ -89,6 +91,7 @@ defmodule Model.Schedule do
           service_id: Model.Service.id(),
           stop_id: Model.Stop.id(),
           stop_sequence: non_neg_integer,
+          stop_headsign: String.t(),
```
paulswartz, line 94: `String.t() | nil`

apps/parse/lib/parse/stop_times.ex
```diff
@@ -35,6 +35,12 @@ defmodule Parse.StopTimes do
       arrival_time: convert_time(row["arrival_time"], row["drop_off_type"]),
       departure_time: convert_time(row["departure_time"], row["pickup_type"]),
       stop_sequence: String.to_integer(row["stop_sequence"]),
+      stop_headsign:
+        if row["stop_headsign"] == "" do
+          nil
+        else
+          copy(row["stop_headsign"])
+        end,
```
paulswartz, line 43: I think there's an `optional_copy` helper?

lemald: Looks like there's one defined locally in `vehicle_positions.ex`, but I can easily move that into the `Parse.Helpers` module.

PR #226, opened 2020-04-29
chore: bump default API version to 2020-05-01

lemald:
In #217 we introduced the new 5/1 API version. Since it's almost May, we should get ready to make that the default.


lemald:
Rebasing to latest master, hopefully for the last time (relative code coverage decreased but that's probably just an artifact of being behind master).

PR #238, opened 2020-05-19
docs: Improve swagger docs for AlertResource

digitalcora:
* Several fields were nullable but not documented as such.

* The docs now explicitly state that, although all identifying fields on `InformedEntity` are nullable, at least one of them will be non-null.

* The previous minimal description of `InformedEntity` is replaced with the more detailed docs from its typespec, which had gone unused in the Swagger docs until now.

This also includes a small fix for a dev-only (?) build issue with the docs, which is explained in its own commit.

| Before | After |
|---|---|
|||

### DIFF
--- a/apps/api_web/lib/api_web/controllers/alert_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/alert_controller.ex
@@ -306,7 +306,7 @@ defmodule ApiWeb.AlertController do
         end,
       InformedEntity:
         swagger_schema do
-          description("Object representing a particular part of the system affected by an alert")
+          description(typedoc(:informed_entity))
 
           properties do
             activities(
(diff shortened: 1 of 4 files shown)


digitalcora:
I can't quite make sense of the coverage change in `diff_server.ex` — nothing I changed _should_ have impacted it, and the unit tests for the module don't appear to explicitly test the code path that lost coverage.

apps/api_web/lib/api_web/controllers/alert_controller.ex
```diff
@@ -315,14 +315,34 @@ defmodule ApiWeb.AlertController do
...
+              "`direction_id` of the affected Trip.\n\n" <> direction_id_description(),
+              example: 0
+            )
+
+            facility(nullable(%Schema{type: :string}, true), "`id` of the affected Facility.",
+              example: "405"
+            )
+
+            route(nullable(%Schema{type: :string}, true), "`id` of the affected Route.",
+              example: "CR_Worcester"
```
lemald, line 329: Extremely minor nitpick about something carried over from the old version, but if we want the example to be an actual route it should be `CR-Worcester` with a hyphen rather than an underscore.

digitalcora: Good catch! I fixed it 😊

PR #244, opened 2020-05-27
feat: add occupancy_status field to Vehicle

fjlanasa:
Asana ticket: [👩‍👩‍👧‍👧 support crowding fields on Vehicle: `occupancy_status` enum ](https://app.asana.com/0/584764604969369/1171321386032805)

Changes:
* Add field `occupancy_status` to `Models.Vehicle`, using the values listed in the [GTFS realtime reference](http://gtfs.org/reference/realtime/v2/#enum-occupancystatus)
* Added logic to parse this field from a `VehiclePostion` entity (both JSON and protobuf)
* Added logic to `VehicleView` to return this value through the `VehicleController`
* Added documentation, just using the descriptions in the official reference

### DIFF
--- a/apps/api_web/lib/api_web/controllers/vehicle_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/vehicle_controller.ex
@@ -244,6 +244,26 @@ defmodule ApiWeb.VehicleController do
               "Speed that the vehicle is traveling in meters per second. See [GTFS-realtime Position speed](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-position).",
               example: 16
             )
+
+            occupancy_status(
+              :string,
+              """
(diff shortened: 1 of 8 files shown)


apps/api_web/lib/api_web/controllers/vehicle_controller.ex
```diff
@@ -244,6 +244,25 @@ defmodule ApiWeb.VehicleController do
               "Speed that the vehicle is traveling in meters per second. See [GTFS-realtime Position speed](https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#message-position).",
               example: 16
             )
+
+            occupancy_status(
+              :string,
```
paulswartz, line 249: Can we mark this as `nullable` too?

fjlanasa: yup, actually the failing test is due to me not marking it as nullable

fjlanasa: sorry, squashed and force pushed before i saw this, so this change is in the initial commit

PR #250, opened 2020-06-11
fix: ignore Green-E shuttle bus shapes

paulswartz:


apps/state/config/config.exs
```diff
@@ -89,6 +89,9 @@ config :state, :shape,
     "858t0001" => -1,
     # Green-D (Newton Highlands)
     "858t0002" => -1,
+    # Green-E (shuttle bus)
+    "6020021" => -1,
+    "6020022" => -1,
```
lemald, line 94: So if I'm understanding this correctly, `prefix_overrides` just overrides the priority of certain shapes, and this is making sure that the shuttle shapes are deprioritized relative to the normal ones?

jfabi:
This seems perfectly fine to me, though an alternate solution could be to switch the shapes in the data to our preexisting `input/replacement_bus` shapes (by adding onto the exist Lechmere shuttle patch in the `gtfs_creator` branch).

apps/state/config/config.exs
paulswartz: Right. https://github.com/mbta/api/blob/7ce0886a643bc09f79388779142c40349b730f28/apps/state/config/config.exs#L150-L152 has a general rule like this for shuttle shapes that we build ourselves, but this one comes out of HASTUS so we need to do a specific override.

paulswartz:
We can do that later if we need/want to: the shape overrides only matter if that shape ID exists.

PR #256, opened 2020-06-24
Return 406 when attempting to stream a 'show' endpoint

losvedir:
Previously, when passing an `accept` header of `text/event-stream` to a "show" endpoint (e.g. `/vehicles/[id]`) the API would return the resource once and close the connection. This first commit of this PR changes it to return a `406` response instead.

The second commit updates the swagger documentation to list `406` as a possible response. I wasn't sure if we intend to enumerate _all_ possible responses there, or just highlight the main ones, in which case we might not need or want to include it. However, when trying it locally, I ran into a lot of swagger errors:

```
Resolver error at paths./vehicles/{id}.get.responses.406.schema.$ref
Could not resolve reference because of: Could not resolve pointer: /definitions/NotAcceptable does not exist in document
```

I guessed at `:NotAcceptable` based on the nearby `:NotFound` but it didn't seem to like it. I wasn't able to find where these were defined for swagger, though. Is there something I can update to include the "Not Acceptable" response?

### DIFF
--- a/apps/api_web/lib/api_web/api_controller_helpers.ex
+++ b/apps/api_web/lib/api_web/api_controller_helpers.ex
@@ -52,6 +52,21 @@ defmodule ApiWeb.ApiControllerHelpers do
   end
 
   def show(module, conn, params) do
+    conn
+    |> get_format()
+    |> show_for_format(module, conn, params)
+  end
+
+  def show_for_format("event-stream", _module, conn, params) do
+    render_json_api(
+      conn,
+      params,
+      {:error, :not_acceptable,
+       "Streaming not supported for an individual resource. Instead list resources and filter by ID."}
+    )
+  end
+
+  def show_for_format(_format, module, conn, params) do
     data =
       case ApiWeb.Params.validate_show_params(params, conn) do
         :ok ->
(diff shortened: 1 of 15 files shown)


apps/api_web/lib/api_web/api_controller_helpers.ex
```diff
@@ -52,6 +52,21 @@ defmodule ApiWeb.ApiControllerHelpers do
...
+    |> get_format()
+    |> show_for_format(module, conn, params)
+  end
+
+  def show_for_format("event-stream", _module, conn, params) do
+    render_json_api(
+      conn,
+      params,
+      {:error, :not_acceptable,
+       "Streaming not supported for 'show' endpoints. Use 'index' and filter by ID."}
```
paulswartz, line 65: Do we describe them in the documentation as `show` endpoints?

apps/api_web/lib/api_web/api_controller_helpers.ex
```diff
@@ -62,7 +62,7 @@ defmodule ApiWeb.ApiControllerHelpers do
       conn,
       params,
       {:error, :not_acceptable,
-       "Streaming not supported for 'show' endpoints. Use 'index' and filter by ID."}
+       "Streaming not supported for an individual resource. Instead list resources and filter by ID."}
```
paulswartz, line 65: :+1:

PR #267, opened 2020-07-30
Add new API version with back-compat departure time filtering

digitalcora:
Prior to mbta/rtr#818, as a consequence of how RTR made predictions, API consumers would never see a departure time that was in the past by more than a few seconds. Here we introduce a new API version and replicate the no-past-departure-times behavior in previous versions for backwards compatibility.


apps/api_web/config/config.exs
```diff
@@ -39,7 +39,8 @@ config :api_web, :versions,
     "2019-02-12",
     "2019-04-05",
     "2019-07-01",
-    "2020-05-01"
+    "2020-05-01",
+    "2020-08-13"
```
digitalcora, line 43: The date is somewhat arbitrary — going by past releases, we use the date at which the new version becomes the default. So I made it be A) a couple weeks in the future, and B) on a Thursday, since that's when our sprint ends. I'm not sure if we have a specific policy around choosing these dates.

digitalcora:
Closing; we've decided that since past departure times are already being published for bus and CR, this is not a backwards-incompatible change.

PR #271, opened 2020-08-13
fix: nil crash in State.Trip.Added

digitalcora:
The stop fetched in `prediction_based_on_shape` could be nil, but later code didn't account for that possibility. Tests didn't catch this due to the data for the "unknown stop" test not including any shapes.

### DIFF
--- a/apps/state/lib/state/trip/added.ex
+++ b/apps/state/lib/state/trip/added.ex
@@ -114,6 +114,8 @@ defmodule State.Trip.Added do
     end
   end
 
+  defp last_stop_id_on_shape(_, _, nil), do: nil
+
   defp last_stop_id_on_shape(%{priority: p} = shape, prediction, stop) when p >= 0 do
     shape_stops =
       State.StopsOnRoute.by_route_id(
--- a/apps/state/lib/state/trip/added.ex
+++ b/apps/state/lib/state/trip/added.ex
@@ -52,10 +52,10 @@ defmodule State.Trip.Added do
   end
 
   @spec prediction_to_trip({Trip.id(), Prediction.t()}) :: [Trip.t()]
-  defp prediction_to_trip({trip_id, %{route_pattern_id: route_pattern_id} = prediction})
-       when is_binary(route_pattern_id) do
-    with %{representative_trip_id: rep_id} <- State.RoutePattern.by_id(route_pattern_id),
-         [trip | _] <- State.Trip.by_id(rep_id) do
+  defp prediction_to_trip({trip_id, prediction}) do
+    with %{route_pattern_id: route_pattern_id} when is_binary(route_pattern_id) <- prediction,
+         %{representative_trip_id: rep_trip_id} <- State.RoutePattern.by_id(route_pattern_id),
+         [trip | _] <- State.Trip.by_id(rep_trip_id) do
       [
         %{
           trip
@@ -68,15 +68,11 @@ defmodule State.Trip.Added do
       ]
     else
       _ ->
-        prediction_based_on_shape(prediction)
+        prediction_to_trip_via_shape(prediction)
     end
   end
 
-  defp prediction_to_trip({_trip_id, %{route_pattern_id: nil} = prediction}) do
-    prediction_based_on_shape(prediction)
-  end
-
-  defp prediction_based_on_shape(prediction) do
+  defp prediction_to_trip_via_shape(prediction) do
     stop =
       case State.Stop.by_id(prediction.stop_id) do
         %{parent_station: nil} = stop -> stop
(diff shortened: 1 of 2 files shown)


apps/state/lib/state/trip/added.ex
```diff
@@ -52,10 +52,10 @@ defmodule State.Trip.Added do
   end
 
   @spec prediction_to_trip({Trip.id(), Prediction.t()}) :: [Trip.t()]
-  defp prediction_to_trip({trip_id, %{route_pattern_id: route_pattern_id} = prediction})
-       when is_binary(route_pattern_id) do
-    with %{representative_trip_id: rep_id} <- State.RoutePattern.by_id(route_pattern_id),
-         [trip | _] <- State.Trip.by_id(rep_id) do
+  defp prediction_to_trip({trip_id, prediction}) do
+    with %{route_pattern_id: route_pattern_id} when not is_nil(route_pattern_id) <- prediction,
```
lemald, line 56: Why the change from `is_binary(route_pattern_id)` to `not is_nil(route_pattern_id)`?

paulswartz:
Any idea which predictions/trips/patterns are causing this issue?

apps/state/lib/state/trip/added.ex
digitalcora: Ah, that was part of an intermediate refactoring step and I didn't notice it was kind of a pointless change now. It should be functionally equivalent, since the value can only be either a string or `nil`, but I'll change it back.

digitalcora:
I've looked in the logs around where this error appears, but there don't seem to be any details that would help in answering the question. Logically the issue occurs under these conditions:

1. A prediction appears in the Prediction state with `trip_match?: false`
2. It has the highest stop sequence among predictions that share its `trip_id` (it's for the "last stop")
3. It doesn't have a `route_pattern_id`, or else its route pattern and/or representative trip can't be found
4. It has a `stop_id`, but no stop with that ID is present in the Stop state

The last item seems strange to me — that a prediction could have a stop ID that matches no known stop. However, we did explicitly account for the possibility:

```elixir
stop =
  case State.Stop.by_id(prediction.stop_id) do
    %{parent_station: nil} = stop -> stop
    %{parent_station: id} -> State.Stop.by_id(id)
    _other -> nil # <== it's this branch
  end
```

It's just that the code right after this didn't handle the case where `stop` is `nil` (ain't dynamic typing grand?).

@paulswartz You might know best whether this is something we expect to happen occasionally, or an exceptional situation we should investigate further. If the latter, we could add some logging here?

paulswartz:
From the logs, it appears to be happening frequently. Can you run the API locally with the debugging?

digitalcora:
Ah, yes, I didn't realize it was happening frequently enough for that to be an option. I ran it locally for a while, and finally caught this one prediction that fit:

```elixir
%Model.Prediction{
  arrival_time: #DateTime<2020-08-14 13:36:37-04:00 EDT America/New_York>,
  departure_time: nil,
  direction_id: 1,
  route_id: "Green-D",
  route_pattern_id: nil,
  schedule_relationship: :added,
  status: nil,
  stop_id: "70209",
  stop_sequence: 650,
  trip_id: "ADDED-1580507311",
  trip_match?: false,
  vehicle_id: "G-10088"
}
```

`70209` appears to be the shuttle drop-off at Lechmere. I'd have expected this stop to exist in our data, but strangely it doesn't — the value `70209` doesn't appear anywhere in `stops.txt` or `stop_times.txt`, despite being referenced in some places like the shuttle definitions. @jfabi do you know why that might be?

In any case, maybe we should log a warning here?

lemald:
Oh, interesting! I'm not sure where you're seeing 70209 associated with the shuttle drop-off at Lechmere, but RTR considers that to be the stop ID for Lechmere (drop-off side on the Green Line). Interestingly the alert for the Lechmere reconstruction only affects Green-E, despite the fact that other branches are sometimes extended to Lechmere, so whenever an operator happens to enter an AVI code that would indicate going to Lechmere, we hit this problem.

lemald:
Possible hack-y solution: update RTR so that the various `RIVERSIDE;LECHMERE` etc. patterns actually go to the North Station turnback, since we know that for now that's as far as they're actually going. We'll need to completely redo a lot of GL patterns once GLX opens anyway.

digitalcora:
It is very possible I misinterpreted what I was reading in `gtfs_creator` — I'm sure you're correct that it's the actual rail stop, and in that case it would make much more sense that it doesn't currently exist in the feed. We should probably add another task to determine how to handle this situation.

In terms of this PR, do you think we should merge as-is, or add logging?

digitalcora:
I will say, from running the app locally it seems like there's one time where it _is_ normal to encounter this scenario a lot, and that's at startup. If we start receiving predictions earlier than we finish loading stops, every prediction will hit this branch for a short time. So any logging we added would be a bit noisy around deploys...

PR #277, opened 2020-09-08
Add back-compat for Nubian and Commuter Rail stop ID changes

digitalcora:
Some things to note:

* This is targeted to `cfg-new-api-release` and references to the new API version use the placeholder `2020-XX-XX` (considered to be "greater than" any `2020-` version or earlier). Once we decide on a cutover date for these changes, that branch can be rewritten to use the actual version and merged into `master`.

* While implementing this I found our previous back-compat for stop ID changes only applied to the `/predictions` and `/schedules` endpoints. For consistency I incorporated these existing mappings into the new `LegacyStops` mechanism that's applied to all endpoints. Technically this is a backwards-incompatible change, but only in the sense that queries which had previously been broken by a stop ID change will start working again. Retro-backwards-compatible?


apps/api_web/lib/api_web/legacy_stops.ex
```diff
@@ -0,0 +1,263 @@
...
+      "Readville" => ["DB-0095", "FB-0095-04", "FB-0095-05", "NEC-2192-02", "NEC-2192-03"],
+      "River Works / GE Employees Only" => ["ER-0099", "ER-0099-01", "ER-0099-02"],
+      "Rockport" => ["GB-0353-S"],
+      "Roslindale Village" => ["NB-0064-S"],
+      "Route 128" => ["NEC-2173", "NEC-2173-01", "NEC-2173-02"],
+      "Rowley" => ["ER-0312-S"],
+      "Ruggles" => ["NEC-2265"],
+      "Ruggles-01" => ["NEC-2265-01"],
+      "Ruggles-02" => ["NEC-2265-02"],
+      "Ruggles-03" => ["NEC-2265-03"],
```
digitalcora, line 185: I noticed that, while our previous back-compat code expanded `Back Bay`, `North Station`, and `South Station` to their corresponding platformed stops like `Back Bay-01`, we never did the same for `Ruggles`. Did these platformed stops always exist, unlike the others?

apps/api_web/lib/api_web/controllers/stop_controller.ex
```diff
@@ -240,7 +243,11 @@ defmodule ApiWeb.StopController do
   def show_data(conn, %{"id" => id} = params) do
     case Params.validate_includes(params, @show_includes, conn) do
       {:ok, _includes} ->
-        Stop.by_id(id)
+        case LegacyStops.expand([id], conn.assigns.api_version) do
+          [single] -> Stop.by_id(single)
+          [_, _] = pair -> Enum.find_value(pair, &Stop.by_id/1)
+          _ -> nil
+        end
```
digitalcora, line 250: I'm beginning to think this isn't a great way to handle this. It only works in the specific scenario where a stop ID is renamed exactly once, and never split into multiple stops — if any splitting or additional renames happen later, it spontaneously starts returning 404 for a legacy ID that used to work fine. This is a fairly opaque behavior compared to just always returning 404 with no backwards compatibility, and feels likely to bite us later.

What we _could_ do is, in `LegacyStops`, encode the knowledge of whether a "new" stop ID replaces the old one, or should be used in addition to it (i.e. platforms). Then `expand` could have an "only renames" option which, if enabled, ignores IDs that are additions. In a situation like:

```
alpha => beta    => gamma
      \> beta-01 => gamma-01
      \> beta-02 => gamma-02
```

Expanding `[alpha]` normally would result in _all_ of the IDs above, but expanding it with "only renames" would just return `[alpha, beta, gamma]`, which would give correct behavior for this endpoint.

losvedir: I agree that this probably isn't the behavior we want, for the reasons you mention. 

The rename vs replace is an interesting idea. And actually, it probably matters for the `place-dudly` to `place-nubn` change, since that's a parent station change, and it feels bad to break that stop for people. But I'm ambivalent about the necessity of handling in the general case vs just a one-off case for that parent station.

apps/api_web/lib/api_web/legacy_stops.ex
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  @doc """
+  Transforms a list of stop IDs as per the given API version and mappings. See the moduledoc for
+  the structure and meaning of mappings.
+  """
+  @spec expand([Stop.id()], String.t(), %{
+          optional(String.t()) => %{optional(Stop.id()) => [Stop.id()]}
+        }) :: [Stop.id()]
+  def expand(stop_ids, api_version, all_mappings \\ @mappings) do
```
losvedir, line 250: This works, but one thing to keep in mind is that about a 1/3 of all API requests are going to be hitting this code path (~100/sec), so it's probably worth spending a bit of time thinking about optimization.

One thought right off the bat I had was whether `@mappings` needs to be a map, since the keys aren't guaranteed to be sorted (they are for small maps, but I forget exactly how it works in Erlang and what causes the switch), which is why you're doing the `Enum.sort_by` here. If it was instead a list already, we could skip the `sort` step.

More generally, it's possible to move some of this work to compile time. In particular, we actually know at compile time for a given API version and stop_id what the replacement stops should be. So with a bit of macro magic and computation we could generate a bunch of function heads like:

```ex
def backwards_compatible_stop("South Station", "2020-01-01"), do: ["South Station-01"]
```

Can you maybe poke at that a little bit and see what the implementation complexity is, to see if it's worth it?

digitalcora: Good point about the request volume. I'll try benchmarking the current approach vs. `master` to see if there's any concerning differences.

paulswartz:
One option here (that we've done in the past) is deploy the branch to production for a few hours/days to see what the performance is like (comparing using the API dashboard in Splunk). Doing that without merging the branch allows for a much easier rollback.

______________________________________________________________________
This email/electronic message, including any attached files, is being sent by the MBTA. It is solely intended for the recipient(s) and may contain information that is proprietary, confidential, legally privileged, and/or exempt from disclosure pursuant to state and federal law. If you have received this message in error or are not the intended recipient(s), please notify the sender immediately by reply, and delete all copies of this email/electronic message and any attached files from your computer. If you are the intended recipient(s), you may use the information contained in this email/electronic message and any attached files only as authorized by the MBTA. Any unauthorized use, dissemination, or disclosure of this email/electronic message and/or its attached files is strictly prohibited.

apps/api_web/lib/api_web/controllers/stop_controller.ex
digitalcora: > it probably matters for the `place-dudly` to `place-nubn` change, since that's a parent station change, and it feels bad to break that stop for people.

That specific change does work fine with this code, since there are no "splits" and we've only renamed the stop once (so far?). Unfortunately it doesn't work with some other stop renames we already have, for example `South Station` to `NEC-2287`, since a "platform split" also occurred so the expanded set of stop IDs is more than `[old, new]`.

apps/api_web/lib/api_web/legacy_stops.ex
digitalcora: From benchmarking it appears this branch adds 1.5µs per request in the best case and 30µs in the worst case (lots of mappings being applied). Our average request time on production is about 3ms, so this branch would increase that by 1% at worst.

I think this is probably acceptable, though I did do a bit of work to see if we could essentially compile the mappings up-front as you suggested. This only gives a 10µs penalty vs. master in the worst case, but also complicates testing of `LegacyStops` since you can't easily dependency-inject the mappings. Overall I'm not sure it's worth it for a gain of 20µs on a 3ms request.

jfabi:
@digitalcora I've gone through a sample of maybe half the stations in the mapping, and they all look good to me, so that portion gets my approval.

Another idea on the implementation of the mapping though: Could it be possible to just define the parent station's `stop_id` for each old child stop, and have that auto-expand to include all the children (assuming `vehicle_type=2`, and not including itself, of course)? That might make this more resilient to the possibility of adding new tracks in the further future. Not sure how this all jives with your thinking yesterday/today.

digitalcora:
> Could it be possible to just define the parent station's `stop_id` for each old child stop, and have that auto-expand to include all the children

This sounds intriguing but I'm not 100% sure what you mean. Could you give an example of what the static "mappings" data structure would look like with this approach?

jfabi:
> > Could it be possible to just define the parent station's `stop_id` for each old child stop, and have that auto-expand to include all the children
> 
> This sounds intriguing but I'm not 100% sure what you mean. Could you give an example of what the static "mappings" data structure would look like with this approach?

Maybe it'd have to live elsewhere, but the mapping would look something like
```
%{
      "Abington" => "place-PB-0194",
      "Anderson/ Woburn" => "place-NHRML-0127",
}
```
and then somewhere else you'd have some code (running once upon start up, if possible, if performance is an issue?) to look at each of those new stops (which are parent stations) and extract children which are Commuter Rail stops, attaching those to the `old_stop`.

digitalcora:
Hm, interesting. This would mean we wouldn't have to make as many changes when new platforms are later added to an already multi-platform stop, but I _think_ the current approach can also handle this — and it does so without having to look anything up in the state (or know that it should look up Commuter Rail stops specifically). Am I missing something, and there's a back-compat scenario the current approach cannot handle at all? (I wouldn't be surprised; this is hairy stuff)

jfabi:
> Hm, interesting. This would mean we wouldn't have to make as many changes when new platforms are later added to an already multi-platform stop, but I _think_ the current approach can also handle this — and it does so without having to look anything up in the state (or know that it should look up Commuter Rail stops specifically). Am I missing something, and there's a back-compat scenario the current approach cannot handle at all? (I wouldn't be surprised; this is hairy stuff)

The current approach could handle this just fine by simply adding new `stop_id`s here in the future—the difference is whether or not an additional https://github.com/mbta/api change would have to be made in tandem, somewhat documented in https://github.com/mbta/wiki/blob/master/transit_tech/Procedures/GTFS/New_Stop_Platform.md right now. I don't feel too strongly here.

digitalcora:
In the latest commit I separated out stop ID renames from "added" stop IDs, and made the `/stops/:id` endpoint only use the renames when it searches for a stop to return.

apps/api_web/lib/api_web/legacy_stops.ex
```diff
@@ -0,0 +1,279 @@
...
+      "70150" => {nil, ["71150"]},
+      "70151" => {nil, ["71151"]},
+      "70200" => {nil, ["71199"]},
+      "70201" => {nil, ["Government Center-Brattle"]},
+      "70202" => {nil, ["Government Center-Brattle"]}
+    },
+    "2020-XX-XX" => %{
+      "Abington" => {"PB-0194-S", []},
+      "Anderson/ Woburn" => {"NHRML-0127", ["NHRML-0127-01", "NHRML-0127-02"]},
+      "Andover" => {"WR-0228-02", []},
```
digitalcora, line 72: @jfabi For cases like this, I thought it was appropriate for it to be a "rename" since the old stop ID is in fact going away and there is only one new stop ID that replaces it. The only functional difference is that it allows e.g. `/stops/Andover` to keep working in older versions. Does this seem right to you?

jfabi: Yep, this sounds right to me. 🎉

losvedir (approved):
Looks good to me, since merging the PR puts it into the "new-api-version" branch, which will then be ready to test on dev-green.

We can also put it on prod at some point for a bit to verify the performance is acceptable, but we'll do that as a follow-up, since the performance is fine for the Keolis testing right now.

PR #289, opened 2020-09-22
fix: exclude Lechmere shuttle stops from Green-E

lemald:
A couple of things to note:
1. From my debugging it looked like `place-lech` was being included as well, which didn't trigger the validation error but is obviously incorrect for now.
2. I also fixed a typo elsewhere in the map (a stray character in the `Green-D` key above).


jfabi (approved):
Looks good to me, and seems like this would address the validation error we've been seeing!

PR #291, opened 2020-09-30
chore: bump dependencies

lemald:


losvedir (approved):
Looks good to me. Seems to be mostly the usual patch version updates we've seen in all the other apps.

PR #298, opened 2020-10-13
Switch from distillery to native mix release

losvedir:
This updates the API to build using a native `mix release` rather than with distillery. It brings the build process in line with our other more recent apps, and seems to be the preferred way going forward. It will also play nicely with the Config Provider being worked on in another branch to move some environment variables to AWS Secrets Manager. While distillery has the concept of config providers, I don't know if they work exactly the same, and since we already proved out this process in other apps with native releases, I figured we might as well switch here, too.

In favor of the new `config/releases.exs` (soon to be `config/runtime.exs`, but the API isn't on Elixir 1.11 yet), I moved the runtime configuration to that file, rather than the `REPLACE_OS_VARS=true` approach of interpolating in the `"${VAR}"` format. However, from experimentation it seems in umbrella apps it doesn't look for a `config/releases.exs` in each app, so I had to move them all to the root directory.

I wasn't sure how "vestigial" the various `rel` configs were, and if the `startup` script is still necessary, but for now I tried to maintain the same behavior in the new `mix release` way.

There's also a small bonus commit to remove the `HTTP_ACCEPTS` environment variable, since that was only for migrating to event streaming, which has now been done.


rel/env.sh.eex
```diff
@@ -0,0 +1,19 @@
...
+#   *)
+#     ;;
+# esac
+
+# Set the release to work across nodes. If using the long name format like
+# the one below (my_app@127.0.0.1), you need to also uncomment the
+# RELEASE_DISTRIBUTION variable below. Must be "sname", "name" or "none".
+
+export RELEASE_DISTRIBUTION=sname
+export RELEASE_NODE=<%= System.get_env("NODE_NAME") %>
```
paulswartz, line 19: With this, you can probably get rid of the `startup` script.

config/releases.exs
```diff
@@ -0,0 +1,31 @@
+import Config
+
+config :ex_aws,
+  dynamodb: [
+    # make sure port is first, otherwise it won't be converted to an integer
+    port: System.get_env("DYNAMO_PORT"),
```
paulswartz, line 6: I think you can make this `String.to_integer(System.get_env("DYNAMO_PORT"))` now, since it's running as an Elixir script.

apps/api_web/mix.exs
```diff
@@ -3,29 +3,30 @@ defmodule ApiWeb.Mixfile do
...
+    [
+      app: :api_web,
+      aliases: aliases(),
+      build_embedded: Mix.env() == :prod,
+      build_path: "../../_build",
+      compilers: [:phoenix] ++ Mix.compilers() ++ [:phoenix_swagger],
+      config_path: "../../config/config.exs",
+      deps: deps(),
+      deps_path: "../../deps",
+      elixir: "~> 1.9",
```
paulswartz, line 15: Since the API is now on 1.10, we can bump this as well.

paulswartz (commented):
I assume this has also been tested on dev green?

losvedir:
Addressed the comments. And, yes, I deployed this branch to dev-green (with the commits on top of our other testing stuff). I've now put it on api-dev, rebased back onto master.

It deployed cleanly and the logs look good. One thing I wasn't entirely sure about was the `RELEASE_NODE` env var from the startup script before and `env.sh.eex` now, and what it's for. I think it's used in the logs? I notice `api-i-[...]@[...]` but not sure if I should be checking for anything else.

paulswartz (approved):
:cake: love to delete a bunch of code!

PR #300, opened 2020-10-16
fix: parse "Woodland" as `park-woodl-garage`

paulswartz:
From Michael G Chiang <mgchiang@us.ibm.com>:

> We found a change has occurred when the Woodland server was updated.
> From now on, the park-woodl-garage garage will be coming in as 'Woodland'.

### DIFF
--- a/apps/parse/config/config.exs
+++ b/apps/parse/config/config.exs
@@ -6,7 +6,7 @@ config :parse, Facility.Parking,
   garages: %{
     "Alewife" => "park-alfcl-garage",
     "Braintree" => "park-brntn-garage",
-    # Woodland
+    "Woodland" => "park-woodl-garage",
     "DataPark" => "park-woodl-garage",
     "MBTA BEVERLY" => "park-ER-0183-garage",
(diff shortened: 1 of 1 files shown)


apps/parse/config/config.exs
```diff
@@ -6,7 +6,7 @@ config :parse, Facility.Parking,
   garages: %{
     "Alewife" => "park-alfcl-garage",
     "Braintree" => "park-brntn-garage",
-    # Woodland
+    "Woodland" => "park-woodl-garage",
```
lemald, line 9: Does it make sense to still leave the `"DataPark" => "park-woodl-garage"` map entry below just in case they ever fall back to using it or something? Or would it be better to just remove it? I suppose we have map entries for both `MBTA SALEM` and `MBTA SALEM ` (with the trailing space)...

paulswartz: I thought about that, and figured that it's better to be safer and keep both versions around.

lemald (approved):
🅿️

PR #305, opened 2020-10-20
Prepare changes for new release

digitalcora:
This includes all the changes that will go into the next release, without adding a new version for them (so the back-compat code, using `2020-XX-XX` as a placeholder, is always enabled). A future commit will add the new version once we decide on the date it should become the default.


losvedir (approved):
All the individual PRs that went into this branch were approved so I think this one should be good to go as well. I deployed the branch to prod and it's been running without issue for the last hour. 👍

PR #308, opened 2020-10-29
feat: add route_type filter to schedule controller

fjlanasa:
Asana ticket: [🍎 Provide a route_type filter for /schedules](https://app.asana.com/0/295455480314405/1198173974731337/f)

Adding a `route_type` filter to the schedule controller. Like the `route_type` filter in the predictions controller, I am requiring that one other filter be included. The implementation is likewise pretty similar.

### DIFF
--- a/apps/api_web/lib/api_web/controllers/schedule_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/schedule_controller.ex
@@ -16,8 +16,7 @@ defmodule ApiWeb.ScheduleController do
   plug(ApiWeb.Plugs.ValidateDate)
   plug(:date)
 
-  @filters ~w(date direction_id max_time min_time route stop stop_sequence
-              trip)s
+  @filters ~w(date direction_id max_time min_time route stop stop_sequence route_type trip)s
   @pagination_opts ~w(offset limit order_by)a
   @includes ~w(stop trip prediction route)
 
@@ -59,6 +58,7 @@ defmodule ApiWeb.ScheduleController do
     include_parameters(@includes)
     filter_param(:date, description: "Filter schedule by date that they are active.")
     filter_param(:direction_id)
+    filter_param(:route_type, desc: "Must be used in conjunction with another filter.")
 
     filter_param(
       :time,
@@ -94,11 +94,15 @@ defmodule ApiWeb.ScheduleController do
   def index_data(conn, params) do
     with {:ok, filtered} <- Params.filter_params(params, @filters, conn),
          {:ok, _includes} <- Params.validate_includes(params, @includes, conn) do
+      # must include one additional filter besides `route_type` and `date`, is automatically included
       case format_filters(filtered, conn) do
+        %{route_type: _} = filters when map_size(filters) == 2 ->
+          {:error, :only_route_type}
+
         filters when map_size(filters) > 1 ->
-          # greater than 1 because `date` is automatically included
           filters
           |> Schedule.filter_by()
+          |> Schedule.filter_by_route_type(Map.get(filters, :route_type))
           |> populate_extra_times(conn)
           |> State.all(Params.filter_opts(params, @pagination_opts, conn))
 
@@ -185,6 +189,10 @@ defmodule ApiWeb.ScheduleController do
           []
         end
     end
+  end
+
+  defp do_format_filter({"route_type", route_types}) do
+    %{route_type: Params.route_types(%{"route_type" => route_types})}
   end
 
   defp do_format_filter({key, time}) when key in ["min_time", "max_time"] do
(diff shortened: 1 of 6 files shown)


apps/api_web/lib/api_web/controllers/schedule_controller.ex
```diff
@@ -16,7 +16,7 @@ defmodule ApiWeb.ScheduleController do
   plug(ApiWeb.Plugs.ValidateDate)
   plug(:date)
 
-  @filters ~w(date direction_id max_time min_time route stop stop_sequence
+  @filters ~w(date direction_id max_time min_time route stop stop_sequence route_type
               trip)s
```
digitalcora, line 20: I think these all fit on one line, if we're changing this anyway ⤴

apps/api_web/lib/api_web/controllers/schedule_controller.ex
```diff
@@ -95,10 +95,13 @@ defmodule ApiWeb.ScheduleController do
     with {:ok, filtered} <- Params.filter_params(params, @filters, conn),
          {:ok, _includes} <- Params.validate_includes(params, @includes, conn) do
       case format_filters(filtered, conn) do
+        %{route_type: _} = filters when map_size(filters) == 2 ->
+          {:error, :only_route_type}
         filters when map_size(filters) > 1 ->
           # greater than 1 because `date` is automatically included
```
digitalcora, line 102: We might want to shift this up above the `case` and reword since it now describes how multiple clauses work (the `2` in `map_size(filters) == 2` is there for the same reason as the `1` in `map_size(filters) > 1`).

apps/api_web/test/api_web/controllers/schedule_controller_test.exs
```diff
@@ -156,11 +156,78 @@ defmodule ApiWeb.SchedulerControllerTest do
...
+      State.Schedule.new_state([schedule1, schedule2])
+
+      for {params, expected} <- [
+            {%{"stop" => "stop"}, [schedule1, schedule2]},
+            {%{"stop" => "stop", "route_type" => "2"}, [schedule2]},
+            {%{"stop" => "stop", "route_type" => "1,2"}, [schedule1, schedule2]}
+          ] do
+        conn = get(conn, "/schedules", params)
+        assert conn.assigns.data == expected
+      end
```
digitalcora, line 228: Any strong opinions on using a loop like this in tests vs. writing out each assertion? I favor the latter since if a test fails it can be hard to tell which iteration of a loop it failed on. For example, here if the failure message tells you the `expected` was `[schedule1, schedule2]`, you don't know if it was on the first or the third iteration. The equivalent "unrolled" version doesn't seem _too_ verbose to me:

```elixir
conn = get(conn, "/schedules", %{"stop" => "stop"})
assert conn.assigns.data == [schedule1, schedule2]

conn = get(conn, "/schedules", %{"stop" => "stop", "route_type" => "2"})
assert conn.assigns.data == [schedule2]

conn = get(conn, "/schedules", %{"stop" => "stop", "route_type" => "1,2"})
assert conn.assigns.data == [schedule1, schedule2]
```

digitalcora (commented):
We should add documentation for the new filter param. Judging by how it works with predictions, all we need is to add `filter_param(:route_type)` within the `swagger_path :index` block in the schedule controller.

While we're here, we could also document that both here and with predictions, `route_type` must be used together with another filter. I think this could be done easily enough by updating the appropriate clause of `filter_param` in `SwaggerHelpers` to accept an option the same way the `:direction_id` clause does.

apps/api_web/test/api_web/controllers/schedule_controller_test.exs
fjlanasa: no strong opinions. i think the above looks fine too, so i'll do that since it produces a more useful message.

apps/api_web/lib/api_web/swagger_helpers.ex
```diff
@@ -82,7 +82,7 @@ defmodule ApiWeb.SwaggerHelpers do
       :query,
       :string,
       """
-      Filter by route_type: https://developers.google.com/transit/gtfs/reference/routes-file.
+      Filter by route_type: https://developers.google.com/transit/gtfs/reference/routes-file. Must be used in conjunction with another filter.
```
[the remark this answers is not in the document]
fjlanasa: ah, of course. pushed up a fix for that.

apps/api_web/lib/api_web/controllers/schedule_controller.ex
fjlanasa: I'm not really coming up a way to shift these clauses around that results in any simpler code. At the end of the day I'll still have to handle `%{route_type: _} = filters when map_size(filters) == 2` differently than just `map_size(filters) > 1`.

fjlanasa: ah, ok. makes sense.

digitalcora (approved):
🉑✨

PR #316, opened 2020-11-25
fix: deprioritize shape 811_0010 for Green-B

fjlanasa:
Adjustment for Winter 2021 rating

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -37,9 +37,9 @@ config :state, :shape,
     "811_0007" => -1,
     # Green-B (North Station)
     "811_0008" => -1,
-    # Green-B (North Station),
+    # Green-B (North Station)
     "811_0009" => -1,
-    # Green-B (North Station),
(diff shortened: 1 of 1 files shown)


apps/state/config/config.exs
```diff
@@ -37,8 +37,10 @@ config :state, :shape,
     "811_0007" => -1,
     # Green-B (North Station)
     "811_0008" => -1,
-    # Green-B (North Station)
+    # Green-B (North Station),
     "811_0009" => -1,
+    # Green-B (North Station),
```
jfabi, line 42: The comma in the comment isn't necessary, per the convention in the rest of this file.

jfabi (commented):
I left a pedantic comment about comments, but otherwise this looks fine!

jfabi (approved):
Looks good to me, assuming Semaphore is also fine with it—you can update the `gtfs_creator` branch after this gets merged in.

PR #318, opened 2020-12-07
test: more flexible Kingston/Plymouth validation

digitalcora:
**Relevant to:** [🚧 Plymouth closure Dec 12–14](https://app.asana.com/0/584764604969369/1199237221152083)

To guard against regression of a bug which caused Plymouth to not be present in the `CR-Kingston` stop list on weekends, this test asserted that it should be there on every day of service. However, this prevents us from implementing a disruption where the Kingston line actually will not stop at Plymouth for a few days, due to the station being closed.

This changes the test to check the stop list by service (instead of by date, which was translated to a service internally), and to only check services with a `schedule_typicality` less than 4. This allows services that don't stop at Plymouth (or Kingston) to pass validation as long as they're considered "atypical".

### DIFF
--- a/apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
+++ b/apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
@@ -138,7 +138,7 @@ defmodule StateMediator.Integration.GtfsTest do
       assert invalid_routes == []
     end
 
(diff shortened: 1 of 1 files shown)


losvedir (approved):
Oh, of course! Great idea and solution! 👍

PR #321, opened 2020-12-18
chore: finalize new API version 2021-01-09

digitalcora:
This simply finds-and-replaces all instances of `2020-XX-XX` to `2021-01-09`, and adds the latter as an official version. I'll create a separate branch to bump the default version, and we can merge and deploy that one on January 9th; this one can be merged at any time.


lemald (approved):
👍

PR #324, opened 2020-12-28
chore: Update hackney to fix checkout_timeout issue

losvedir:
`mix deps.update hackney` to pick up the new bug fix in 1.17 which resolves the periodic `checkout_timeout` errors.


digitalcora (approved):
🐴🚕

PR #325, opened 2020-12-28
fix: filtering route patterns by parent station ID

digitalcora:
**Asana:** [🐛 API route_patterns / stop filter](https://app.asana.com/0/584764604969369/1199540919582901)

We had originally intended that filtering route patterns by a parent station ID would return all route patterns serving its child stops, hence the existence of this `route_patterns_by_family_stops` function that was never called until now. Other endpoints with stop ID filters already work this way, except for `/facilities`, but that is currently irrelevant since we don't associate facilities with child stops.


losvedir (approved):
Looks good to me, other than the failing test from the newlines. 👍

PR #327, opened 2020-12-30
feat: mark instance unhealthy if vehicle or prediction data is stale

lemald:
Ticket: [[#869] 💳 mark instances unhealthy when they're unable to update vehicle locations / predictions](https://app.asana.com/0/584764604969369/1199189033001773/f)

Note that I would have liked to get the `timestamps_fetcher` and `current_time_fetcher` values via `Appication.compile_env` and just have different values for them in `config.exs` and `test.exs`, but unfortunately I need to change `current_time_fetcher` from one test to another, so `Application.get_env` it is. I'm open to suggestions for better ideas, though.

### DIFF
--- a/apps/health/lib/health/checkers/real_time.ex
+++ b/apps/health/lib/health/checkers/real_time.ex
@@ -1,0 +1,23 @@
+defmodule Health.Checkers.RealTime do
+  @moduledoc """
+  Health check which makes sure that real-time data (vehicle positions and
+  predictions) aren't stale.
+  """
+  @stale_data_seconds 15 * 60
+  @current_time_fetcher &DateTime.utc_now/0
+
+  def current do
+    updated_timestamps = State.Metadata.updated_timestamps()
+
+    [prediction: updated_timestamps.prediction, vehicle: updated_timestamps.vehicle]
+  end
+
+  def healthy?(current_time_fetcher \\ @current_time_fetcher) do
+    updated_timestamps = State.Metadata.updated_timestamps()
+    current_time = current_time_fetcher.()
+
+    Enum.all?([:prediction, :vehicle], fn feed ->
+      DateTime.diff(current_time, updated_timestamps[feed]) < @stale_data_seconds
+    end)
+  end
+end
--- a/apps/health/config/config.exs
+++ b/apps/health/config/config.exs
@@ -5,5 +5,6 @@ use Mix.Config
 config :health,
   checkers: [
     Health.Checkers.State,
-    Health.Checkers.RunQueue
+    Health.Checkers.RunQueue,
+    Health.Checkers.RealTime
   ]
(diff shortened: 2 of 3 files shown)


digitalcora:
> I'm open to suggestions for better ideas

Rather than environment configuration (and thus needing to have the tests shuffle the environment around), we could use dependency injection. Example:

```elixir
  # real_time.ex

  @updated_timestamps_fn &State.Metadata.updated_timestamps/0

  def current(updated_timestamps_fn \\ @updated_timestamps_fn) do
    %{prediction: prediction, vehicle: vehicle} = updated_timestamps_fn.()
    [prediction: updated_timestamps.prediction, vehicle: updated_timestamps.vehicle]
  end
```

```elixir
    # real_time_test.exs

    defp fake_updated_timestamps do
      # ...
    end

    test "returns current vehicle and prediction timestamps" do
      assert current(&fake_updated_timestamps/0) == [
               prediction: DateTime.from_naive!(~N[2020-12-30 15:00:00], "Etc/UTC"),
               vehicle: DateTime.from_naive!(~N[2020-12-30 15:05:00], "Etc/UTC")
             ]
    end
```

The value of `@updated_timestamps_fn` could of course also be taken from environment configuration, but I don't think there's much benefit to this if the value doesn't differ per environment (like, dev vs. prod).

lemald:
@digitalcora I totally forgot that the simplest approach available was even an option here. 😝  I just pushed the changes to take that approach instead.

apps/health/lib/health/checkers/real_time.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Health.Checkers.RealTime do
+  @moduledoc """
+  Health check which makes sure that real-time data (vehicle positions and
+  predictions) aren't stale.
+  """
+  @stale_data_seconds 15 * 60
+  @timestamps_fetcher &State.Metadata.updated_timestamps/0
+  @current_time_fetcher &Timex.now/0
```
paulswartz, line 8: `DateTime.utc_now/0`

apps/health/lib/health/checkers/real_time.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Health.Checkers.RealTime do
+  @moduledoc """
+  Health check which makes sure that real-time data (vehicle positions and
+  predictions) aren't stale.
+  """
+  @stale_data_seconds 15 * 60
+  @timestamps_fetcher &State.Metadata.updated_timestamps/0
```
digitalcora, line 7: Come to think of it, do we need to inject this? We could call `State.Metadata.state_updated` from the tests, kind of like how we use other state modules' `new_state` functions.

lemald: Hmm, personally I kind of like tests that don't depend too much on the state of other processes, but it does look like there's a bit of a convention of using the approach you describe with the various `State` modules. I'll give that a shot.

digitalcora (approved):
🩺✅

PR #329, opened 2021-01-15
chore: Adjustments for CR schedule effective 1/23

jfabi:
**Asana Ticket:** [❄️ Winter CR schedule beginning 1/23](https://app.asana.com/0/584764604969369/1199734802667758/f)

This is a companion pull request to [...] in https://github.com/mbta/gtfs_creator, which actually adds the new Commuter Rail schedule. The new schedule is a lot like our existing reduced schedule, and not very much like our previous fall regular schedule, so several validations and other knobs need fine tuning. Descriptions are included inline.

Please note that this branch is currently failing in CI. This is because it's looking at the gtfs_creator `master` branch, which has different data than what's going in. Usually what we do in these cases is merge in the API branch while it's still failing, but only right before merging and deploying in the GTFS branch.


apps/state/config/config.exs
```diff
@@ -12,6 +12,8 @@ config :state, :route_pattern,
     "Green-D-1-1" => true,
     "Green-D-3-1" => true,
     # don't ignore Rockport Branch shuttles
+    "Shuttle-BeverlyRockport-0-0" => false,
+    "Shuttle-BeverlyRockport-0-1" => false,
```
jfabi, line 16: This is so that the Beverly–Rockport shuttles get displayed on the dotcom timetable for `CR-Newburyport`.

apps/state/config/config.exs
```diff
@@ -195,6 +197,12 @@ config :state, :stops_on_route,
     ],
     {"CR-Newburyport", 0} => [
       [
+        "Beverly",
```
jfabi, line 200: These blocks in `stop_order_overrides` help ensure that the API returns the stops on `CR-Newburyport` in the order expected on the MBTA.com timetable. Before this change. Weekends (with the entire Rockport branch shuttled) in the new schedule, the Rockport branch stops would show up in a weird place on the timetable sometimes.

apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
```diff
@@ -163,33 +163,6 @@ defmodule StateMediator.Integration.GtfsTest do
       assert invalid_services == []
     end
 
-    test "keeps Winsor Gardens and Plimptonville in the right order" do
```
jfabi, line 166: Plimptonville no longer has service and will no longer, even when we begin to add back service in the future.

apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
```diff
@@ -308,15 +281,6 @@ defmodule StateMediator.Integration.GtfsTest do
     end
   end
 
-  describe "schedules" do
-    test "CR-Fairmount has alternate route trips" do
```
jfabi, line 312: The new schedule no longer has an Franklin Line trips which operate via the Fairmount Line.

apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
```diff
@@ -262,10 +235,10 @@ defmodule StateMediator.Integration.GtfsTest do
     test "Newburyport/Rockport has 2 non-ignored shapes each direction" do
       [shapes_0, shapes_1] = shapes_in_both_directions("CR-Newburyport")
 
-      assert [%{name: "North Station - Manchester"}, %{name: "North Station - Newburyport"}] =
+      assert [%{name: "North Station - Newburyport"}, %{name: "North Station - Manchester"}] =
```
jfabi, line 238: Because of the change in weekend service, the new schedule has more train trips to/from Newburyport than to/from Manchester.

digitalcora:
This all looks good to me, especially the explanatory comments — could we incorporate these into the relevant commit messages, so they'll be easily accessible when looking at history, blames, etc.?

jfabi:
> This all looks good to me, especially the explanatory comments — could we incorporate these into the relevant commit messages, so they'll be easily accessible when looking at history, blames, etc.?

That's a neat idea. Squashed for merger with the comments added to the commit message.

losvedir (approved):
👍

PR #330, opened 2021-01-19
chore: Update build/release scripts for fargate

losvedir:
Small tweaks to prepare for importing the API into terraform and deploying to Fargate via ECS.

Fargate abstracts away the underlying EC2 instance from us, so the `curl` command in the mix release startup fails. Looking at the test logs, it shows e.g. `api@ip-10-10-10-10`, (i.e. some IP address corresponding to our fargate instance), so that should still be enough to distinguish instances' requests from each other in the logs.

The other change relates to the way we model ECS in terraform; we have to specify an initial docker image, and then our deploys are free to change that without having to update our terraform config. However, for that initial image, it's our terraform convention to point to a tag `:latest` so spinning up a new AWS environment will start with the latest image.

### DIFF
--- a/rel/env.sh.eex
+++ b/rel/env.sh.eex
@@ -16,7 +16,6 @@
 # RELEASE_DISTRIBUTION variable below. Must be "sname", "name" or "none".
 
 DIRNAME=$(dirname $0)
-INSTANCE_ID=$(curl --max-time 5 -s http://169.254.169.254/latest/meta-data/instance-id)
 
 export RELEASE_DISTRIBUTION=sname
-export RELEASE_NODE=api-${INSTANCE_ID:-unknown}
+export RELEASE_NODE=api
--- a/semaphore/build_push.sh
+++ b/semaphore/build_push.sh
@@ -14,6 +14,8 @@ fi
 githash=$(git rev-parse --short HEAD)
 docker build --pull -t $APP:latest .
 docker tag $APP:latest $DOCKER_REPO:git-$githash
+docker tag $APP:latest $DOCKER_REPO:latest
 
 # push images to ECS image repo
 docker push $DOCKER_REPO:git-$githash
+docker push $DOCKER_REPO:latest


digitalcora:
> that should still be enough to distinguish instances' requests from each other in the logs

I recall we have some things on the Splunk dashboard that show graphs for individual instances based on the ID, so those might need to be updated for the new format.

> for that initial image, it's our terraform convention to point to a tag `:latest` so spinning up a new AWS environment will start with the latest image

With this change, it seems like we will update `latest` whenever a build-push is done, for any environment. Wouldn't this result in `latest` sometimes being e.g. some experimental commit we deployed to dev-green but may never merge, and thus a poor choice for creating a new environment?

losvedir:
> I recall we have some things on the Splunk dashboard that show graphs for individual instances based on the ID, so those might need to be updated for the new format.

Good catch. I see one, anyway, that counts how many nodes there are running. Looks like it doesn't quite work out of the box on my branch's splunk index, but should be an easy enough tweak once we switch.

> With this change, it seems like we will update latest whenever a build-push is done, for any environment. Wouldn't this result in latest sometimes being e.g. some experimental commit we deployed to dev-green but may never merge, and thus a poor choice for creating a new environment?

That's a good question. You can read Ian's reasoning in the devops repo. The easiest thing would have been to just tag a known commit as `:base` or something, but it could grow stale (maybe app wouldn't boot up) or be deleted, since our ECR docker image repository doesn't retain all images and tags indefinitely. `:latest` seems fine - we won't spin up a new environment very often, just when we do, we want some reasonably recent image to ensure the thing starts up, so we can then use the normal deploy process to update with a new docker image. An alternative would be to only push when merging to dev, maybe with a separate script only invoked by the Dev server in semaphore, but to me it's not really worth it. The main point, again, is just to have a reasonably recent commit to ensure that the app starts up (e.g., contains the relevant environment variables).

ianwestcott:
@digitalcora at the risk of self-promotion, I wrote about the reasoning behind this approach (and the many alternatives I evaluated along the way) here: https://ianw.org/posts/ecs-terraform-continuous-delivery.html

digitalcora:
I see, I see. I wonder if `latest` is the best name for this, since in my experience (on Docker Hub at least) that's a convention to indicate the latest _released version_ of an image, which in our case more closely corresponds to a production deploy. But it's a minor quibble, if anything.

digitalcora (approved):
🖼🐳

PR #333, opened 2021-02-06
fix: Do not ignore new Fitchburg Line shuttles

jfabi:
_This is a companion pull request to https://github.com/mbta/gtfs_creator/pull/1096._

#### Summary of changes

**Asana Ticket:** [🚧 Littleton–Alewife weekday shuttles](https://app.asana.com/0/584764604969369/1199686637343555/f)

Introduces new API exceptions to ensure that upcoming Fitchburg Line shuttles properly appear on the dotcom timetable page, and so that the stops appear in the correct order.

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -19,7 +19,12 @@ config :state, :route_pattern,
     "Shuttle-ManchesterRockport-0-0" => false,
     "Shuttle-ManchesterRockport-0-1" => false,
     "Shuttle-RockportWestGloucester-0-0" => false,
-    "Shuttle-RockportWestGloucester-0-1" => false
+    "Shuttle-RockportWestGloucester-0-1" => false,
+    # don't ignore Fitchburg Line shuttles to/from Alewife
+    "Shuttle-AlewifeLittletonExpress-0-0" => false,
+    "Shuttle-AlewifeLittletonExpress-0-1" => false,
+    "Shuttle-AlewifeLittletonLocal-0-0" => false,
+    "Shuttle-AlewifeLittletonLocal-0-1" => false
   }
 
 config :state, :shape,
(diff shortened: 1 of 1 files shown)


apps/state/config/config.exs
```diff
@@ -19,7 +19,12 @@ config :state, :route_pattern,
     "Shuttle-ManchesterRockport-0-0" => false,
     "Shuttle-ManchesterRockport-0-1" => false,
     "Shuttle-RockportWestGloucester-0-0" => false,
-    "Shuttle-RockportWestGloucester-0-1" => false
+    "Shuttle-RockportWestGloucester-0-1" => false,
+    # don't ignore Fitchburg Line shuttles to/from Alewife
```
digitalcora, line 23: I'm not totally familiar with the meaning of "ignore" here... ignore in what context?

jfabi: (1) `prefix_overrides` (see Line 31), (2) `suffix_overrides` (further below), and (3) multi-route trips are all combined to help define which trips should be deprioritized (negative integers) or highly-prioritized (highly-positive integers) for the purpose of selecting a primary pattern, and for returning a set of stops, for a route-direction.

In the case of shuttles, we deprioritize all shuttles in `suffix_overrides` (Line 179).

In https://github.com/mbta/api/blob/master/apps/state/lib/state/helpers.ex, we introduce a way to ignore those (in the case of shuttles) deprioritizations a the route pattern level by way of `ignore_override_prefixes`. `false` means that we ignore the deprioritization, which sounds backwards here out of the context of `ignore_trip_route_pattern` in the other file.

digitalcora: My head hurts from the double-negatives, but I think I understand now. Thanks!

Ideally I think the function of `ignore_trip_route_pattern?` would be inverted so the meaning is positive (returns true if the trip **should** be considered part of the route). Then rather than having a `false` here meaning "**don't** _not_ consider this to be on the route", you'd have a `true` meaning "**do** consider this to be on the route".

If you think this would be an improvement, I'd be all for tossing it into this PR as a separate commit, in the spirit of leaving things better than we found them. Understand if you have more important things to get to, though!

digitalcora (approved):
Approving for merge as-is; if you want to do the suggested refactor, I'll re-review! 🔄 ✅

apps/state/config/config.exs
jfabi: I think it's a good idea (thinking about it makes my head hurt too), but maybe warrants a little more thinking (ie, do we want to try to simplify things further, and/or explain them within the config file)?

I decided to file this as an inbound task, [Clean up API shape/pattern exception logic](https://app.asana.com/0/881264583703207/1199912774871224/f), to groom further—feel free to add any additional thoughts to it.

PR #337, opened 2021-02-08
feat: health check for the maximum number of open Ports

paulswartz:
Asana: https://app.asana.com/0/116562505129567/1199678649889558

Normally, we have less than 100 open Ports at any one time. When we get an
unhealthy instance, we tend to see many more than that. This health check
will mark that instance as unhealthy so that AWS will replace the instance
for us.


lemald (approved):
Looks like Credo has one minor gripe, but looks good aside from that!

PR #338, opened 2021-02-10
Close event streams cleanly on app shutdown 

digitalcora:
This comprises two major changes:

### `fix: close event streams on app shutdown`

We handle event streams using a Plug that (under normal circumstances) never returns — the Phoenix process spawned to handle the connection would thus remain alive until terminated as part of app shutdown. The HTTP connection would not be closed cleanly in this case, resulting in clients getting errors.

To address this, we add a "Canary" process after the Endpoint in the supervision tree, whose sole purpose is to be terminated before the Endpoint so it can give the EventStream processes advance notice of the shutdown. They respond to this message by ending their infinite loops and returning the final conn from the `call` function, allowing Phoenix to close the connection.

### `chore: handle load balancer connection draining`

When an API instance is shut down on AWS, the following sequence of events occurs:

1. ALB stops routing incoming connections to the instance ("draining")
2. ALB waits for a configurable amount of time
3. ALB forcibly closes any remaining open connections to the instance
4. ECS sends SIGTERM to the application

This creates a problem where the app doesn't know it's being shut down until it's already too late to cleanly close any open connections. This isn't an issue for most apps since HTTP connections generally have a lifetime shorter than the ALB draining time, but it is an issue for event-stream connections, which remain open indefinitely (until either the client closes them or an error occurs). When ALB closes these, it creates unwanted errors for clients.

To work around this, we add an `ALBMonitor` process to the app, which periodically checks the instance's status in the ALB API. If the status is "draining", we proactively shut down the VM so the app can clean up its own connections before ALB does. Phoenix performs its own connection draining on shutdown, and in the commit prior to this we added "shutdown awareness" to event streams.

### Notes

This currently only works on `dev-blue`, the only environment that has been Terraformed so far. The others run on different infrastructure that would require a different implementation of the `ALBMonitor`, and we're not bothering with that since they will be Terraformed eventually. In these environments, the `ECS_CONTAINER_METADATA_URI` variable is not present, so the `ALBMonitor` will quietly do nothing.


apps/api_web/lib/api_web/event_stream/diff_server.ex
```diff
@@ -130,6 +132,11 @@ defmodule ApiWeb.EventStream.DiffServer do
     end
   end
 
+  @impl GenServer
+  def terminate(_reason, %{refs: refs}) do
+    for parent <- Map.keys(refs), do: send(parent, :exiting)
```
paulswartz, line 137: Rather than trapping exits and manually sending a message, what about having the parents monitoring this process? Then BEAM will send it a message automatically when this process stops.

apps/alb_monitor/lib/alb_monitor/monitor.ex
```diff
@@ -0,0 +1,94 @@
...
+        log_warn("get_instance_ip failed: #{inspect(other)}")
+        nil
+    end
+  end
+
+  defp log_warn(message) when is_binary(message), do: Logger.warn("alb_monitor " <> message)
+  defp log_info(message) when is_binary(message), do: Logger.info("alb_monitor " <> message)
+
+  defp schedule_check(%State{check_interval: check_interval}) do
+    Process.send_after(self(), :check, check_interval)
```
paulswartz, line 92: Rather than manually doing this, it might be easier to use [:timer.send_interval](https://erlang.org/doc/man/timer.html#send_interval-2).

apps/api_web/test/api_web/event_stream_test.exs
```diff
@@ -93,14 +93,24 @@ defmodule ApiWeb.EventStreamTest do
...
+      assert {:close, conn} = receive_result(state)
       assert chunks(conn) =~ "filter[] is required"
       on_exit(fn -> assert_stopped(state.pid) end)
     end
+
+    test "closes the connection when notified the diff server is exiting", %{conn: conn} do
+      state = initialize(conn, @module)
+      assert_receive_data()
+
+      send(self(), :exiting)
```
paulswartz, line 105: Rather than faking this, can you stop the DiffServer and see that the result is correct?

paulswartz (commented):
Did you look at using the `drainer` configuration from the Cowboy2Adapter? https://hexdocs.pm/phoenix/Phoenix.Endpoint.Cowboy2Adapter.html#module-endpoint-configuration

apps/alb_monitor/mix.exs
```diff
@@ -0,0 +1,60 @@
...
+
+  def project do
+    [app: :alb_monitor,
+     aliases: aliases(),
+     build_embedded: Mix.env == :prod,
+     build_path: "../../_build",
+     config_path: "../../config/config.exs",
+     deps: deps(),
+     deps_path: "../../deps",
+     elixir: "~> 1.3",
```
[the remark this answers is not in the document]
digitalcora: I copied it from some other `mix.exs` in the repo, but maybe when creating new apps we should use whatever Elixir version the overall repo is actually using at the moment!

apps/alb_monitor/lib/alb_monitor/monitor.ex
```diff
@@ -0,0 +1,94 @@
...
+
+  require Logger
+
+  @ex_aws Application.compile_env!(:alb_monitor, :ex_aws)
+  @http Application.compile_env!(:alb_monitor, :http)
+
+  defmodule State do
+    @moduledoc "State of the GenServer."
+    defstruct check_interval: 5_000,
+              fetch_env_fn: &System.fetch_env/1,
```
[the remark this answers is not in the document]
digitalcora: I think I see what you mean, but how would tests fake and manipulate the environment with this approach?

apps/alb_monitor/mix.exs
```diff
@@ -0,0 +1,60 @@
...
+  #   {:myapp, in_umbrella: true}
+  #
+  # Type "mix help deps" for more examples and options
+  defp deps do
+    [
+      {:ex_aws, "~> 2.0"},
+      {:ex_aws_elastic_load_balancing, "~> 2.0"},
+      {:httpoison, "~> 1.7"},
+      {:mox, "~> 1.0", only: :test},
+      {:poison, "~> 3.1"},
```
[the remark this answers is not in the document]
digitalcora: I noticed we already had Poison installed from elsewhere in the umbrella and wanted to avoid using two different JSON libraries, but on closer inspection we are _also_ already using Jason. Might as well switch this, then.

apps/alb_monitor/lib/alb_monitor/monitor.ex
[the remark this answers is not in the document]
paulswartz: You could either update the environment, or have it overridable:

```elixir
defmodule State
  defstruct arn: Application.get_env!(:alb_monitor, :arn)
end

def start_link(overrides // %{}) do
  state = struct!(State, overrides)
  GenServer.start_link(__MODULE__, state)
end

# in test
start_link(arn: "test_arn")
```

apps/api_web/lib/api_web.ex
```diff
@@ -19,8 +19,8 @@ defmodule ApiWeb do
       # Start the endpoint when the application starts
       worker(ApiWeb.RateLimiter, []),
       worker(RequestTrack, [[name: ApiWeb.RequestTrack]]),
-      supervisor(ApiWeb.EventStream.Supervisor, []),
-      supervisor(ApiWeb.Endpoint, [])
+      supervisor(ApiWeb.Endpoint, []),
+      supervisor(ApiWeb.EventStream.Supervisor, [])
```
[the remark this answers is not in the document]
digitalcora: In my testing, doing it this way seemed necessary to get the connections closed cleanly — if the endpoint shuts down first, the event-stream processes are terminated without knowing they're supposed to finish up, since the DiffServer is what tells them that, but _it_ doesn't know it's getting shut down until after the Endpoint is shut down and it's too late. But I hadn't thought about the dependency when things are starting up, and you're definitely right.

Possibly the solution involves what you mentioned above:

> Did you look at using the `drainer` configuration from the Cowboy2Adapter?

I hadn't seen this before now, and it seems like it might help but is also very under-documented... I can't find any guidance on how to use it or what it can do (a not-yet-published [version of the doc](https://github.com/phoenixframework/phoenix/blob/c8883af5582a38496e4b7e45e05d3a4d759a6caa/lib/phoenix/endpoint/cowboy2_adapter.ex#L17-L21) at least clarifies what the _default_ value does, but to me this only raises more questions). Have you used it before?

apps/api_web/lib/api_web/event_stream/diff_server.ex
digitalcora: Good idea, though wouldn't it contradict your point that the DiffServer supervisor should be before the Endpoint in the supervision tree? If that were the case, the Endpoint would be shut down before the DiffServers, so the connection processes wouldn't be around anymore to get the message that their DiffServer is shutting down.

apps/api_web/test/api_web/event_stream_test.exs
digitalcora: Good idea, I'll give it a try!

apps/alb_monitor/lib/alb_monitor/monitor.ex
digitalcora: I'm a bit wary of `send_interval` since it doesn't consider the time taken to handle the messages, creating the possibility of filling up the mailbox or sending excessive requests to an external service (imagine if the ALB API starts timing out or responding very slowly). With this approach it's not possible for a `check` to be less than 5 seconds after the previous one, no matter how long it took to perform the check, so it seems safer. Does that make sense?

apps/alb_monitor/lib/alb_monitor/monitor.ex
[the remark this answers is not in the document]
digitalcora:
```elixir
defstruct arn: Application.fetch_env!(:alb_monitor, :arn)
```

I'm not sure this would work, since it would be getting the value at compile time, but the value wouldn't be present until runtime. It would have to be like:

```elixir
defstruct arn_fn: fn -> Application.fetch_env!(:alb_monitor, :arn) end
```

And then you're pretty much back to the `fetch_env_fn` approach, but with more `fn`s. Though there might be value in having all the runtime environment variables nicely listed in `releases.exs`; in that case it could work like this?

```elixir
defstruct env_fn: &Application.fetch_env!(:alb_monitor, &1)

# in releases.exs
config :alb_monitor, target_group_arn: System.get_env("ALB_TARGET_GROUP_ARN"), ...
```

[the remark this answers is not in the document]
paulswartz: I thought they happened at runtime, but if not, you can have it as a default in `start_link`:
```elixir
def start_link(overrides // %{}) do
  overrides = Map.put_new(overrides, :arn, Application.fetch_env!(:alb_monitor, :arn))
  ...
end

File: apps/api_web/lib/api_web.ex (hunk 7 above)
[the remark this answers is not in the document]
Comment (paulswartz) [reply apps/api_web/lib/api_web.ex:23]: I haven't used it before. The code in https://github.com/elixir-plug/plug_cowboy/blob/v2.4.1/lib/plug/cowboy/drainer.ex  is interesting, though. It's a separate process, running after the Cowboy listeners, which waits for them to stop having any open connections.

For us, it sounds like what we want is for `DiffServer` to remain before `Endpoint`, and for a new server to be after `Endpoint` and be responsible for shutting down the long-running connections, similar to what `Plug.Cowboy.Drainer` is doing here.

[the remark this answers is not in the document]
Comment (digitalcora) [reply apps/api_web/lib/api_web.ex:23]: Yes, I'm not sure if the `drainer` config can help us here; it seems like the mentioned "drainer process" is [always `Plug.Cowboy.Drainer`](https://github.com/phoenixframework/phoenix/blob/c8883af5582a38496e4b7e45e05d3a4d759a6caa/lib/phoenix/endpoint/cowboy2_adapter.ex#L71), and the config might be better named `drainer_opts`.

> For us, it sounds like what we want is for `DiffServer` to remain before `Endpoint`, and for a new server to be after `Endpoint` and be responsible for shutting down the long-running connections

Hmmmmm... I'm trying to think of how we do this without races. So far I'm thinking something like this (calling the new process `Canary`):

```
ApiWeb.Supervisor => EventStream.Supervisor => ConnectionRegistry
                  => Endpoint                        |
                  => EventStream.Canary -------------/
```

New event-stream connections register themselves in a registry owned by the EventStream supervisor (since it has to be running before the Endpoint). When the canary is shut down, it sends all running event-stream connections a message using the registry. It then shuts down immediately, but because of the Cowboy drainer, the connections will have a few seconds to respond to the message and finish up before the Endpoint shuts down.

The race I can see here is that there's a tiny window _after_ the Canary has fetched the list of event-stream connections, but _before_ the Endpoint starts draining, where a new event-stream connection could be established. It wouldn't know about the shutdown, and so would still experience the problem we're trying to fix. Maybe this would be uncommon in practice, but it bugs me that it would be possible. (it also bugs me how convoluted this is, but maybe it's just a convoluted problem) What do you think?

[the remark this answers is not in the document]
Comment (paulswartz) [reply apps/api_web/lib/api_web.ex:23]: > The race I can see here is that there's a tiny window _after_ the Canary has fetched the list of event-stream connections, but _before_ the Endpoint starts draining, where a new event-stream connection could be opened.

Hasn't the load balancer already dropped us at this point? I don't think it's possible for there to be a new connection by the time we're disconnecting existing connections.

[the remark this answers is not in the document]
Comment (digitalcora) [reply apps/api_web/lib/api_web.ex:23]: Oh, you're right, it would have. So this race might exist in a non-AWS context, but fortunately we don't have to worry about that. 😅 If you think the approach is good otherwise, I will give it a shot.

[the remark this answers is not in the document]
Comment (paulswartz) [reply apps/api_web/lib/api_web.ex:23]: Yeah, I'd go for that rather than re-ordering the supervision tree.

File: apps/alb_monitor/lib/alb_monitor/monitor.ex (hunk 5 above)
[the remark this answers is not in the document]
Comment (digitalcora) [reply apps/alb_monitor/lib/alb_monitor/monitor.ex]: Tried doing it this way (more or less) in my latest commit. I'm in favor of putting the env vars in `releases.exs`, and it did allow extraction of the `no_ecs_metadata_uri` case to a separate function clause, which is nice.

**Comment by digitalcora** [comment]
@paulswartz Incorporated the new approach we talked about! Though still have to verify it works on dev-blue.

File: apps/alb_monitor/lib/alb_monitor/monitor.ex [hunk 8]
```diff
@@ -0,0 +1,100 @@
...
+    def default do
+      %__MODULE__{
+        ecs_metadata_uri: Application.get_env(:alb_monitor, :ecs_metadata_uri),
+        target_group_arn: Application.get_env(:alb_monitor, :target_group_arn)
+      }
+    end
+  end
+
+  @spec start_link(%State{}) :: GenServer.on_start()
+  def start_link, do: State.default() |> start_link()
```
paulswartz, line 30: I'm surprised Credo doesn't complain about the single pipe.
```elixir
def start_link(initial_state // State.default()) do
  GenServer.start_link(__MODULE__, initial_state)
end
```

apps/alb_monitor/lib/alb_monitor/monitor.ex
```diff
@@ -0,0 +1,100 @@
...
+      ip_address
+    else
+      unmatched ->
+        log_warn("get_instance_ip failed: #{inspect(unmatched)}")
+        nil
+    end
+  end
+
+  defp log_warn(message) when is_binary(message), do: Logger.warn("alb_monitor " <> message)
+  defp log_info(message) when is_binary(message), do: Logger.info("alb_monitor " <> message)
```
paulswartz, line 95: nit: you don't need to create a new binary: `Logger.warn(["alb_monitor ", message])`

apps/api_web/lib/api_web/event_stream/canary.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  @impl true
+  def init(notify_fn) do
+    Process.flag(:trap_exit, true)
+    {:ok, notify_fn}
+  end
+
+  @impl true
+  def terminate(:shutdown, notify_fn), do: notify_fn.(:shutdown)
+  def terminate({:shutdown, _reason}, notify_fn), do: notify_fn.(:shutdown)
```
paulswartz, line 24: since you only call `notify_fn/1` with `:shutdown` does it need to take an argument?

apps/api_web/lib/api_web/event_stream/supervisor.ex
```diff
@@ -2,52 +2,68 @@ defmodule ApiWeb.EventStream.Supervisor do
...
+    :ok = Registry.unregister(SubscriberRegistry, :subscribers)
+    :ok = DiffServer.unsubscribe(pid)
+  end
 
-      {:error, _} = error ->
-        error
-    end
+  @doc "Sends the given message to all processes subscribed to a DiffServer."
+  @spec subscriber_dispatch(any) :: :ok
+  def subscriber_dispatch(message) do
```
paulswartz, line 45: Do you need to talk to the clients of the `DiffServer`s? If you shut the DiffServer's normally, their clients should receive a `:DOWN` message that they can handle by gracefully exiting.

apps/alb_monitor/lib/alb_monitor/monitor.ex
digitalcora: For some reason I was pretty strongly convinced this would call `State.default` at compile time and thus defeat the point, but a quick `IO.puts` experiment proves it doesn't. And, actually, now that I check the Elixir guides, there is this sentence:

> Any expression is allowed to serve as a default value, but it won’t be evaluated during the function definition.

I guess it basically compiles to exactly what I wrote out here. Neat!

apps/api_web/lib/api_web/event_stream/supervisor.ex
digitalcora: It currently works the other way around: DiffServers monitor their clients (so it can remove the client from its list and terminate if it has no clients anymore).

I don't think there would be anything stopping the clients from also monitoring their DiffServer (?), in which case we wouldn't need the registry of clients... but then we're also leaning a bit more on implementation details. If the goal is to close the HTTP connections, why do that by terminating a different process you know the connection processes are monitoring (a mostly irrelevant detail), when you could notify them directly?

We probably should have connection processes monitor their DiffServer, though; if it goes down unexpectedly, for some reason other than a shutdown, we don't want those connections to stay open but stop delivering new data.

apps/api_web/lib/api_web/event_stream/canary.ex
digitalcora: I guess it's sort of a separation-of-concerns thing. The normal value for this function is `subscriber_dispatch`, which just forwards the provided message to all subscribers, not caring about what it is; the `Supervisor` doesn't need to know it's specifically being used to message a shutdown. I might instead call this variable `dispatch_fn` to make it clear it's expected to be a generic thing.

apps/api_web/lib/api_web/event_stream/supervisor.ex
digitalcora: ...as I think about this more it really feels like a "conversation" that should be happening exclusively between the Canary and the HTTP processes. The one exists to notify the others of a shutdown; there's no reason for even the DiffServer Supervisor to be involved. I'll try a more isolated approach.

I did try using the "terminate the diffservers" approach for comparison: it's in `cfg-close-event-streams-by-terminating`

digitalcora: Eh, it turns out this was harder than it looked. This has far exceeded its timebox already so I'm going to leave it where it is for now — unless you're strongly of the opinion that the `by-terminating` approach is better and we should use that, in which case I will swap it in.

paulswartz: What do you think is better about this approach, compared to the `by-terminating` approach? To me, the `by-terminating` approach looks much cleaner.

digitalcora: This sums it up I think:

> [the by-terminating approach is] leaning a bit more on implementation details. If the goal is to close the HTTP connections, why do that by terminating a different process you know the connection processes are monitoring (a mostly irrelevant detail), when you could notify them directly?

I would agree it's an approach that requires fewer changes and less code overall, but it's also more tied to an implementation detail. So I think it's a trade-off, not unambiguously better.

paulswartz: You need to communicate with the `DiffServer` supervisor either way: at that point the implementation of shutting down the connections is out of the hands of `Canary` and we can go with whatever is easiest (the other approach). 

If we do need to go back to "communicate with each HTTP connection", we can always update `terminate_servers` to send the shutdown messages at that point.

digitalcora: > You need to communicate with the `DiffServer` supervisor either way

This is sort of why I thought it would be better to not do that at all, by having the Canary maintain the registry and having the connection processes "subscribe" directly to it for shutdown notifications. But agree that the approach already written in `by-terminating` is fine, in the interests of not churning further on this 😄

digitalcora:
I saw a couple intermittent failures on Semaphore that seemed to indicate the 25ms check interval in the tests might be too high, so I reduced it to 20ms and did a whole bunch of reruns. It hasn't failed since, and has run 14 times in a row, so this number is probably good.

apps/api_web/lib/api_web/event_stream.ex
```diff
@@ -37,50 +38,52 @@ defmodule ApiWeb.EventStream do
...
+        continue(state, ": keep-alive\n")
 
       {:error, rendered} when is_list(rendered) ->
-        {:error, chunk(state.conn, ["event: error\ndata: ", rendered, "\n\n"])}
+        close(state, ["event: error\ndata: ", rendered, "\n\n"])
 
-      :timeout ->
-        update_state(state, chunk(state.conn, ": keep-alive\n"))
+      {:DOWN, _ref, :process, ^pid, _reason} ->
+        close(state)
```
paulswartz, line 67: This should probably check the `reason` and close normally if it's a regular shutdown reason and send an error first otherwise.

apps/api_web/lib/api_web/event_stream/supervisor.ex
```diff
@@ -2,52 +2,74 @@ defmodule ApiWeb.EventStream.Supervisor do
...
     end
   end
 
-  @doc "Unsubscribes the current pid from the given Server pid."
-  defdelegate server_unsubscribe(pid), to: ApiWeb.EventStream.DiffServer, as: :unsubscribe
+  @doc "Unsubscribes the current process from the given DiffServer."
+  @spec server_unsubscribe(pid) :: :ok
+  def server_unsubscribe(pid) do
+    :ok = DiffServer.unsubscribe(pid)
+  end
```
paulswartz, line 45: Is there a reason to re-write this, rather than delegating?

paulswartz (commented):
Have you tested this on dev-blue? I believe the ticket has a `curl` command you can use to see if the connection is terminated gracefully.

digitalcora:
> Have you tested this on dev-blue? I believe the ticket has a `curl` command you can use to see if the connection is terminated gracefully.

I've been testing on dev-blue, yes, including the current iteration of the branch. It seems to work!

apps/api_web/lib/api_web/event_stream/supervisor.ex
digitalcora: Ah, no, there isn't a reason to rewrite it anymore. This is a remnant of the previous iteration where this function started doing more things (with the connection registry).

apps/api_web/lib/api_web/event_stream.ex
digitalcora: What should the `data` of the event be in this case? I don't think there's much we could meaningfully tell a consumer just from knowing the DiffServer crashed, but I think there has to be some `data`.

digitalcora: Ah, never mind, I think we can easily render a JSON-API 500 error in a similar way as the DiffServer does errors.

apps/api_web/lib/api_web/event_stream.ex
```diff
@@ -37,50 +38,54 @@ defmodule ApiWeb.EventStream do
...
+        continue(state, ": keep-alive\n")
 
       {:error, rendered} when is_list(rendered) ->
-        {:error, chunk(state.conn, ["event: error\ndata: ", rendered, "\n\n"])}
+        close(state, ["event: error\ndata: ", rendered, "\n\n"])
 
-      :timeout ->
-        update_state(state, chunk(state.conn, ": keep-alive\n"))
+      {:DOWN, _ref, :process, ^pid, reason} ->
+        if reason in [:normal, :shutdown] or match?({:shutdown, _}, reason),
```
paulswartz, line 67: Maybe better as a new function w/ pattern matching?

digitalcora: I wanted to avoid repetition of `close(state)`, which I don't think extracting this to a function would accomplish.

digitalcora: It would also place the `close` call at a different level of abstraction than the other `continue`/`close` calls, and I think having them at the same level is better for readability.

paulswartz (approved):
🍰 Excited to stop getting the error e-mails when we do deploys!

PR #350, opened 2021-03-19
docs: bring ShapeResource up to date

digitalcora:
This bit of documentation confusingly referred to shapes having stops. The concept of "groups of trips that share a stop sequence" has since been moved from shapes to route patterns and the `representative_trip` relationship, and shapes are now just a polyline.

New version:

Screen Shot 2021-03-19 at 3 09 57 PM

### DIFF
--- a/apps/api_web/lib/api_web/controllers/shape_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/shape_controller.ex
@@ -100,22 +100,19 @@ defmodule ApiWeb.ShapeController do
       ShapeResource:
         resource do
           description("""
-          Shape representing the stops to which a particular trip can go. Trips grouped under the same route can have \
-          different shapes, and thus different stops. See \
-          [GTFS `shapes.txt`](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#shapestxt)
+          Sequence of geographic points representing a path vehicles will travel on a trip. See \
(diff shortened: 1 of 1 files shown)


apps/api_web/lib/api_web/controllers/shape_controller.ex
```diff
@@ -100,22 +100,19 @@ defmodule ApiWeb.ShapeController do
...
+            The sequence of points in \
+            [Encoded Polyline Algorithm Format](https://developers.google.com/maps/documentation/utilities/polylinealgorithm).
+            Libraries for decoding polylines are available in many languages, for example:
 
-            [Encoded Polyline Algorithm Format](https://developers.google.com/maps/documentation/utilities/polylinealgorithm)
-
-            ## Example Libraries
-
-            * [Javascript](https://www.npmjs.com/package/polyline)
-            * [Erlang](https://blog.kempkens.io/posts/encoding-and-decoding-polylines-with-erlang/)
```
bklebe, line 117: Is this Erlang post out of date?

digitalcora: I didn't think it was very relevant since it's sort of a how-to for writing your own polyline decoder, rather than a library you can install (which I think is the point here: that libraries exist and you probably _shouldn't_ build your own).

bklebe (approved):
Thanks!

PR #360, opened 2021-04-07
fix: Remove shape priority overrides for Kingston/Plymouth

jfabi:
Asana task: [[Extra] 🚧 Fix Kingston Line API shape priorities](https://app.asana.com/0/584764604969369/1200153288405579/f)

Removes shape priority overrides for Kingston and Plymouth. Plymouth Station is now closed, and so the only shapes being used on `CR-Kingston` are those for South Station <-> Kingston, which the existing overrides de-prioritize.

Also removes overrides relating to Franklin Line route patterns no longer in use (the `route_pattern_id`s we generate for CR are based on a hash of the stops; the `CR-Franklin-Foxboro-` could still be relevant if the Foxboro Pilot resumes later this year since that's a change we could still make to the `route_pattern_id` in https://github.com/mbta/gtfs_creator/blob/master/gtfs_creator2/commuter_rail/reader.py#L97).

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -5,8 +5,6 @@ use Mix.Config
 config :state, :route_pattern,
   ignore_override_prefixes: %{
     # don't ignore Foxboro via Fairmount trips
-    "CR-Franklin-3-0" => false,
-    "CR-Franklin-3-1" => false,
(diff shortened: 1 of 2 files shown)


bklebe (approved):
🚀

PR #358, opened 2021-04-02
chore: Add Worcester Line connection shuttle to timetable

jfabi:
**_Sibling of https://github.com/mbta/gtfs_creator/pull/1164, a GTFS-creator pull request that creates the relevant shuttle bus._**

#### Summary of changes

**Asana Ticket:** [[Extra] 🌷 New CR schedule for Worcester Line](https://app.asana.com/0/584764604969369/1200132943599673/f)

Ensures that the patterns/shapes for a new service, the Newton Connection RailBus (part of the Worcester Line) on route ID `Shuttle-NewtonHighlandsWellesleyFarms`, aren't excluded from stops/stations in the Worcester Line timetable (linked below). The effect is that this adds Riverside and Newton Highlands to the website timetable, just as they are on the physical and PDF timetables.

Also enforces the ordering of stations with Newton Highlands and Riverside added to the mix.

Currently deployed to dev-green: https://green.dev.mbtace.com/schedules/CR-Worcester/timetable?date=2021-04-05.

PR #371, opened 2021-05-04
chore: Add Forest Hills to Providence Line timetable

jfabi:
**Asana task:** [[Extra] 🐛🚂 Forest Hills doesn't appear on Providence Line outbound timetable](https://app.asana.com/0/584764604969369/1200278941033220/f)

Ensures that Forest Hills gets shown on the outbound timetable for `CR-Providence` on MBTA.com to match the [PDF and physical schedules](https://cdn.mbta.com/sites/default/files/route_pdfs/2021-spring/2021-04-05-spring-providence-stoughton-v2-accessible.pdf).

This branch is up live on dev-green, and you can compare https://green.dev.mbtace.com/schedules/CR-Providence/timetable?date=2021-05-19 against https://www.mbta.com/schedules/CR-Providence/timetable?date=2021-05-19 to see the difference.

Note: This pull request is very similar to https://github.com/mbta/api/pull/358, just on a different line/stops.

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -282,6 +282,13 @@ config :state, :stops_on_route,
         "place-newtn",
         "place-WML-0035"
       ]
+    ],
+    {"CR-Providence", 0} => [
+      [
+        "place-rugg",
+        "place-forhl",
+        "place-NEC-2203"
+      ]
     ]
   },
   not_on_route: %{
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -25,7 +25,9 @@ config :state, :route_pattern,
     "Shuttle-AlewifeLittletonLocal-0-1" => false,
     # don't ignore Newton Connection RailBus for Worcester Line
     "Shuttle-NewtonHighlandsWellesleyFarms-0-0" => false,
-    "Shuttle-NewtonHighlandsWellesleyFarms-0-1" => false
+    "Shuttle-NewtonHighlandsWellesleyFarms-0-1" => false,
+    # don't ignore Providence trains stopping at Forest Hills
+    "CR-Providence-d01bc229-0" => false
   }
 
 config :state, :shape,


apps/state/config/config.exs
```diff
@@ -25,7 +25,9 @@ config :state, :route_pattern,
     "Shuttle-AlewifeLittletonLocal-0-1" => false,
     # don't ignore Newton Connection RailBus for Worcester Line
     "Shuttle-NewtonHighlandsWellesleyFarms-0-0" => false,
-    "Shuttle-NewtonHighlandsWellesleyFarms-0-1" => false
+    "Shuttle-NewtonHighlandsWellesleyFarms-0-1" => false,
+    # don't ignore Providence trains stopping at Forest Hills
+    "CR-Providence-d01bc229-0" => false
```
losvedir, line 30: Hm, unfortunately this seems like it might be a little brittle, since the ID is liable to change, perhaps if a stop e.g. becomes a flag stop or a leave early. Is that likely?

When switching to these route pattern IDs, I changed this configuration list to use "prefixes", for the purposes of the `CR-Franklin-Foxboro-` at the top. But I don't think we can do `CR-Providence-` here, right? That would match too many.

losvedir: Depending how likely it is for `d01bc229` to change out from under us, we could extend the matching functionality a bit here to handle this CR format. I initially sketched out something that uses a regex, but unfortunately can't find a way to make that work cleanly, since this config is used at compile time to make macros, and you can't use a regex as a function guard. But we could have special handling for CR route patterns:

```ex
ignore_override_cr: %{
  {"CR-Providence-", "-0"} => false,
}
```

And then update the function generation code in `apps/lib/state/helpers.ex`:

```ex
for {{prefix, suffix}, ignore?} <-
      Application.get_env(:state, :route_patterns)[:ignore_override_cr] do
  def ignore_trip_route_pattern?(%Trip{route_pattern_id: unquote(prefix) <> binary-size(8) <> unquote(suffix)}), do: unquote(ignore?)
```

But I'm not sure how likely the CR route pattern is to change out from under us, if this is all worth it.

jfabi: Hmm, I'd say that it's _relatively_ unlikely that the route pattern would change, but it's certainly not impossible. If it was to happen, though, it would be during one of the (usually) semiannual major schedule changes, during which we make sure to test all new schedules on dev-green, with an eye toward peculiarities such as these late-night connecting trains. So my worries aren't too high on this.

Another solution could come from GTFS-creator: The existing prefix `CR-Franklin-Foxboro-` used to be generated by us upon detection of such a trip for this very purpose: see https://github.com/mbta/gtfs_creator/blob/136c04a28491d6f16fee69cce41fc35083388aa2/gtfs_creator2/commuter_rail/reader.py#L96. We could do a very similar thing to pick up this trip and give it a prefix of like `CR-Providence-ForestHills-`, and then intercept it here in the API's config. The con, of course, would be adding a bit more to GTFS-creator's existing bloat. 🙂

losvedir: Hmm, I see. Yeah, I don't want to mess with GTFS-creator right now for that, so I think this is fine for now.

losvedir (approved):
👍

PR #377, opened 2021-05-18
Fix delayed events when streaming alerts

digitalcora:
### Background

`State.Alert` maintains several "extra" ETS tables (beyond the Mnesia table that comes with `State.Server`), used to look up alerts by their active periods and informed entity attributes. Previously it kept these current by overriding `handle_new_state` to call the original, then update the tables. However, since the base `handle_new_state` contains the call to publish the `:new_state` event to subscribers, this resulted in a race condition:

1. A DiffServer subscribes to `/alerts?filter[route]=1`
2. The alerts JSON is updated to include a new alert affecting route `1`
3. The new alerts make their way to `State.Alert.handle_new_state`
4. `State.Alert` calls the original `handle_new_state`, which updates the Mnesia table and publishes the `:new_state` event
5. Meanwhile, the DiffServer receives the `:new_state` event and reruns its query; it gets no new results, because the `InformedEntity` table used to look up alerts by route has not yet been updated
6. Meanwhile, back in `State.Alert.handle_new_state`, the extra tables are updated, too late for the DiffServer to see the results

While it's also possible the extra tables could be updated before the event was delivered, in practice the event usually won the race. The result was `/alerts` subscriptions with filters were often "behind" by one update, creating a delay of up to one minute (since that's how often Alerts-UI uploads a new JSON file at minimum).

### Resolution

To fix this, I introduced a `post_commit_hook` to `State.Server` that allows a server to run code after a new state has been committed, but before the `:new_state` event is published. `State.Alert` now uses this hook to update its extra tables.

Note there is still a temporary inconsistency between the base table and the extra tables, the event is just not published until everything becomes consistent. Making this entire operation "transactional" would be a larger task (and maybe without much real benefit).

### Other changes

In separate commits:

* Added tests for the other `State.Server` hooks, since they didn't have any
* Updated `State.Service` to also use the new hook (this fixes a similar but mostly theoretical race condition, since the relevant "extra table" isn't used by the `/services` controller itself)

### DIFF
--- a/apps/state/lib/state/alert.ex
+++ b/apps/state/lib/state/alert.ex
@@ -66,15 +66,16 @@ defmodule State.Alert do
 
   @impl Events.Server
   def handle_event(_, _, _, state) do
+    # Re-process the current state to execute hooks
     handle_new_state(all())
     {:noreply, state}
   end
 
-  def handle_new_state(items) do
-    ret = super(items)
+  @impl State.Server
+  def post_commit_hook do
     all_alerts = all()
     for table <- @subtables, do: table.update(all_alerts)
-    ret
+    :ok
   end
 
   @impl State.Server
--- a/apps/state/lib/state/server.ex
+++ b/apps/state/lib/state/server.ex
@@ -29,9 +29,10 @@ defmodule State.Server do
 
   """
   @callback handle_new_state(binary) :: term
+  @callback post_commit_hook() :: :ok
   @callback post_load_hook([struct]) :: [struct] when struct: any
   @callback pre_insert_hook(struct) :: [struct] when struct: any
-  @optional_callbacks [post_load_hook: 1, pre_insert_hook: 1]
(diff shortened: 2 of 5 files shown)


mkennedm (approved):
Looks good.

losvedir:
Wow, good catch, and slick solution!

My only request would be to update the `State.Server` module doc with a short paragraph about the callbacks and their chronological ordering, since in the code they're ordered alphabetically. Reminds me of [ActiveRecord callbacks](https://guides.rubyonrails.org/active_record_callbacks.html#available-callbacks), so maybe just a quick list like that or something.

digitalcora:
@losvedir It's a bit more than a "quick list" but I took the liberty of improving the overall module docs as well as adding a callback list. I'm not sure if this is _too_ detailed but I couldn't think of a better place to document some of these details, and I think it's good to have them documented (so thanks for the reminder!).

PR #378, opened 2021-05-19
sample code for deploying to Elastic Beanstalk

paulswartz:
asana: https://app.asana.com/0/116562505129567/1200354670994292

### DIFF
--- a/.github/workflows/dev-green.yml
+++ b/.github/workflows/dev-green.yml
@@ -1,0 +1,72 @@
+name: Deploy to Dev-green
+
+on: workflow_dispatch
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    environment: dev-green
+    concurrency: dev-green
+    env:
+      AWS_REGION: us-east-1
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Get commit metadata
+        run: |
+          echo SHA_SHORT=$(git rev-parse --short ${{ github.sha }}) >> $GITHUB_ENV
+          echo COMMIT_MSG=$(git log --format=%B -n 1 ${{ github.sha }} | head -n 1) >> $GITHUB_ENV
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ env.AWS_REGION }}
+      - name: Login to Amazon ECR
+        uses: aws-actions/amazon-ecr-login@v1
+      - uses: mbta/actions/build-push-ecr@v1
+        id: build-push
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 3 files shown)


.github/workflows/dev-green.yml
```diff
@@ -0,0 +1,47 @@
...
+          echo COMMIT_MSG=$(git log --format=%B -n 1 ${{ github.sha }} | head -n 1) >> $GITHUB_ENV
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ env.AWS_REGION }}
+      - name: Login to Amazon ECR
+        uses: aws-actions/amazon-ecr-login@v1
+      - run: ./build.sh
```
ianwestcott, line 28: Could build.sh be rolled into its own action? E.G. `build-eb-dockerrun`, which could be run after the existing [`build-push-ecr`](https://github.com/mbta/actions/tree/main/build-push-ecr)?

.github/workflows/dev-green.yml
```diff
@@ -0,0 +1,47 @@
...
+        env:
+          APP: api
+          DOCKER_REPO: ${{ secrets.DOCKER_REPO }}
+      - name: Deploy to EB
+        uses: einaregilsson/beanstalk-deploy@v17
```
losvedir, line 33: I'm not quite sure how the version pinning works. Can `@v17` be updated out from under us? After the codecov incident, I'm sort of gunshy about using 3rd party scripts like this which look like they have access to AWS keys. I'm squinting and it looks similar to downloading the codecov script and running it, but maybe the `@v17` is akin to pinning to a checksum?

If not, I wonder if it would be worthwhile forking that action, and then using our own fork.

The same applies to `aws-actions`, really, but that feels a little more trustworthy. But I'm open to a discussion about having a rule of forking and using only our own actions. (Which would then entail a fair bit of work keeping them up to date, etc.)

.github/workflows/dev-green.yml
paulswartz: It could, but I think each project has their own which works slightly differently at this point?

.github/workflows/dev-green.yml
paulswartz: It can; that's a good point. I'll create a fork.

.github/workflows/dev-green.yml
paulswartz: Talked with @ianwestcott on Slack, and I'm going to look into adding an action to generalize this step.

paulswartz:
@ianwestcott does this look more like what you were thinking? If so, I'll pull those two composite actions into `mbta/actions`.

ianwestcott (approved):
Yes!!! This is perfect, thank you! Hopefully these actions will cover 90% of the needs for deploying our Elastic Beanstalk environments, and ideally the other 10% can be eliminated in preparation for the move to ECS.

paulswartz:
> I think this looks generally good to me. Is the omission of `dev` intentional? Is there a way to make it automatically deploy on merging a PR?

@losvedir I didn't want it to conflict with the same behavior on Semaphore. I can create one with the deploy on merge commented out, or we can disable it on Semaphore. wydt?

losvedir:
I see. I vote disable the auto-merge on semaphore, just so we can see the full process here. We can manually deploy on Semaphore if we need to for a time.

losvedir (approved):
Cool! 👍

PR #380, opened 2021-05-24
refactor: remove double-negative in stops_on_route

digitalcora:
**Asana:** https://app.asana.com/0/584764604969369/1199912774871224

Certain trips can be "ignored" by the default logic that determines which stops are "on" a route. We created a config value to override this, but due to "ignored" being `true` internally, a `false` in this config meant "don't ignore", or "don't _not_ consider these trips" — a double negative that made the value confusing to read.

This inverts the meaning of the config, such that `true` means "_do_ consider these trips", and propagates that change through the rest of the code:

* "canonical" is used as the opposite of "ignored" (this also replaces "alternate" in `StopsOnRoute` specifically, which referred to the fact that originally alternate-route trips were the only "ignored" ones)

* `ignore_trip_route_pattern?` and `ignore_trip_for_route?` are renamed `stops_on_route?` and `stops_on_route_by_shape?`, to clarify they have the same basic function except for the latter using the shape priority (and this is the unusual one, since shape priorities are deprecated)

### DIFF
--- a/apps/state/lib/state/connecting_stops.ex
+++ b/apps/state/lib/state/connecting_stops.ex
@@ -186,6 +186,6 @@ defmodule State.ConnectingStops do
 
   @spec patterns_at_stop(Stop.t()) :: MapSet.t(RoutePattern.id())
   defp patterns_at_stop(%{id: id}) do
-    [id] |> RoutesPatternsAtStop.route_patterns_by_family_stops(ignore?: false) |> MapSet.new()
+    [id] |> RoutesPatternsAtStop.route_patterns_by_family_stops(canonical?: false) |> MapSet.new()
   end
 end
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -2,34 +2,6 @@
 # and its dependencies with the aid of the Mix.Config module.
 use Mix.Config
 
-config :state, :route_pattern,
-  ignore_override_prefixes: %{
-    # don't ignore Foxboro via Fairmount trips
-    "CR-Franklin-Foxboro-" => false,
-    # ignore North Station Green-D patterns
-    "Green-D-1-1" => true,
-    "Green-D-3-1" => true,
-    # don't ignore Rockport Branch shuttles
-    "Shuttle-BeverlyRockport-0-0" => false,
-    "Shuttle-BeverlyRockport-0-1" => false,
-    "Shuttle-ManchesterGloucester-0-0" => false,
-    "Shuttle-ManchesterGloucester-0-1" => false,
-    "Shuttle-ManchesterRockport-0-0" => false,
-    "Shuttle-ManchesterRockport-0-1" => false,
-    "Shuttle-RockportWestGloucester-0-0" => false,
-    "Shuttle-RockportWestGloucester-0-1" => false,
-    # don't ignore Fitchburg Line shuttles to/from Alewife
-    "Shuttle-AlewifeLittletonExpress-0-0" => false,
-    "Shuttle-AlewifeLittletonExpress-0-1" => false,
-    "Shuttle-AlewifeLittletonLocal-0-0" => false,
-    "Shuttle-AlewifeLittletonLocal-0-1" => false,
-    # don't ignore Newton Connection RailBus for Worcester Line
-    "Shuttle-NewtonHighlandsWellesleyFarms-0-0" => false,
-    "Shuttle-NewtonHighlandsWellesleyFarms-0-1" => false,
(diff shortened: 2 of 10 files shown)


digitalcora:
cc @jfabi since it was a conversation we had that led to this task 🙂

mkennedm (approved):
Looks good, although it wouldn't hurt to have another more experienced person took a look as wll.

apps/state/lib/state/connecting_stops.ex
```diff
@@ -186,6 +186,6 @@ defmodule State.ConnectingStops do
 
   @spec patterns_at_stop(Stop.t()) :: MapSet.t(RoutePattern.id())
   defp patterns_at_stop(%{id: id}) do
-    [id] |> RoutesPatternsAtStop.route_patterns_by_family_stops(ignore?: false) |> MapSet.new()
+    [id] |> RoutesPatternsAtStop.route_patterns_by_family_stops(canonical?: false) |> MapSet.new()
```
losvedir, line 189: I haven't gotten to this code change yet, so I might answer my question later: but should this be `canonical?: true`, since before it was `ignore?: false`?

digitalcora: They mean the same thing. `ignore?: false` means "don't ignore", i.e. include trips that are `ignored` — `canonical?: false` means "non-canonical", i.e. include trips that are not `canonical`. Since `ignored == !canonical`, the value should be unchanged.

digitalcora:
> Wow, this is fascinating to see. There's a lot of logic devoted to figuring out what stops are "on" a route, which I didn't know was the case. Is it simplified by [this ticket](https://app.asana.com/0/881264583703207/1200325279789524)? I suppose we can't totally remove all this logic, since that ticket doesn't address bus routes, but I wonder how much Commuter Rail and shuttles in particular complicate this logic. And I don't think I see any bus routes in the overrides.

I think to the extent this logic is more than a simple "join" through trips and stop times (and really it's a bit weird that it isn't), it was probably for expediency of implementing features on other projects. If this wasn't a pure refactor and had a much larger timebox, there are things I'd like to revisit, like how `StopsOnRoute` and `RoutesPatternsAtStop` are conceptually the inverse of each other but have different implementations and aren't symmetric, or how the concept of "canonical" / "ignored" trips is hidden from clients and you can't control whether it's used (this is one of the issues that prevents `connecting_stops` from being useful for shuttle routes). Could be the subject of a future RFC.

digitalcora:
> Wow, this is fascinating to see. There's a lot of logic devoted to figuring out what stops are "on" a route, which I didn't know was the case.

I suspect to the extent this logic is more than a simple "join" through trips and stop times (and really it's a bit weird that it isn't), it was probably for expediency of implementing features on other projects. If this wasn't a pure refactor and had a much larger timebox, there are things I'd like to revisit, like how `StopsOnRoute` and `RoutesPatternsAtStop` are conceptually the inverse of each other but have different implementations and aren't symmetric, or how the concept of "canonical" / "ignored" trips is hidden from clients and you can't control whether it's used (this is what prevents `connecting_stops` from being useful for shuttle routes). Could be the subject of a future RFC.

> Is it simplified by [this ticket](https://app.asana.com/0/881264583703207/1200325279789524)? I suppose we can't totally remove all this logic, since that ticket doesn't address bus routes, but I wonder how much Commuter Rail and shuttles in particular complicate this logic. And I don't think I see any bus routes in the overrides.

I think "the canonical/official set of stops for this route" and "the stops that are on this route right now" are both valid things to want to know, in different scenarios, so we probably wouldn't want to _replace_ the dynamic relationship with what's described in that task. Good to think about how the API is going to expose this data, though.

paulswartz:
Most of the code here predates route patterns, so there needed to be a lot of logic to handle putting stops in the "right" order given only the data in `trips.txt` and `stop_times.txt:

- bus routes with branches, like supplemental trips
- weird commuter rail routes like Kingston Plymouth
- the Red Line

losvedir (approved):
This looks great! `canonical` is much clearer than the inverse logic with `ignore` to me. Thanks for this refactor.

One last check before merging: can you deploy to api-dev-green, if you haven't already, so we can do a quick skim of dotcom dev-green and make sure things seem to be in order?

digitalcora:
@losvedir Done. I took a look at some CR lines (including Worcester where the Newton RailBus still seem to be doing fine) and the Green-D which uses an override, and nothing stuck out, but let me know if you want to check something too.

losvedir:
Looks good to me, too. Thanks! 👍

PR #388, opened 2021-06-28
fix: don't return "empty" relationships

digitalcora:
Due to a bug in `ja_serializer`, not-included relationships would sometimes be rendered as `{}` instead of not being present at all, in violation of the JSON:API spec. This bug was fixed upstream, but we hadn't backported the fix to our own fork of the library. See: https://github.com/mbta/ja_serializer/pull/4

This also fixes a test that accidentally relied on the broken behavior.


paulswartz (approved):

digitalcora:
@paulswartz Thanks! Can we remove the Semaphore check as "required" in the repo settings? I otherwise can't merge this, since it looks like the repo is gone from Semaphore (not sure how the check showed up on this PR in the first place).

PR #390, opened 2021-06-30
fix: beanstalk-deploy inputs are kebab-case

losvedir:


digitalcora (approved):
Checked against the valid input names from the action, looks ok to me!

PR #391, opened 2021-06-30
fix: Filtering with date does not return duplicate routes

mkennedm:
[🍎🐛 /routes with date returns duplicates](https://app.asana.com/0/584764604969369/1200285233066599)

https://api-v3.mbta.com/routes?filter[type]=1&filter[date]=2021-06-30 (replace timestamp with current date) No longer has duplicate results. `apps.state.lib.state.routes_by_service.do_get_routes` now removes duplicates from the list it returns.


apps/state/lib/state/routes_by_service.ex
```diff
@@ -45,7 +45,7 @@ defmodule State.RoutesByService do
         []
 
       items ->
-        Enum.flat_map(items, fn {_key, routes} -> routes end)
+        Enum.uniq(Enum.flat_map(items, fn {_key, routes} -> routes end))
```
paulswartz, line 48: This does work, but IIRC from when we were testing it, there were also duplicates /inside/ `routes` here. Can we 1) check whether that's true and 2) if it is true, fix that issue?

mkennedm: The duplicate appears to becoming from the call to `:ets.select` on [line 34](https://github.com/mbta/api/pull/391/files#diff-877a5831dd4603e26e9a5c96801afa138e3c49738a31ab7f262e555304f7c853R34). I'm not familiar with `:ets` functions which is why this is taking me a while.

mkennedm: For more context `Enum.flat_map(service_ids, fn service_id -> Enum.map(route_types, fn type -> {{{service_id, type}, :_}, [], [:"$_"]} end) end)
[{{{"service", 3}, :_}, [], [:"$_"]}]` returns `[{{{"service", 3}, :_}, [], [:"$_"]}]`. Calling `:ets_select([{{{"service", 3}, :_}, [], [:"$_"]}])` seems to be returning `[{{"service", 3}, ["route", "route"]}]`.

mkennedm: @paulswartz Is there a way to force `:ets.select` to not produce lists with duplicates or do I have to remove the duplicates after they're created?

paulswartz: `:ets.select` is returning the data that's in the table. In this case, the data is a list with some duplicates:

|key|value|
|-|-|
|`{"service", 3}` | `["route", "route"]`

Ideally though, we wouldn't be storing a list with duplicates in the table in the first place. To fix that, you should look for where we're calling `:ets.insert` and adjusting that data so that it's not duplicated. Once that's addressed, it might also be possible to remove the `Enum.uniq` from here.

apps/state/test/state/routes_by_service_test.exs
```diff
@@ -80,6 +81,14 @@ defmodule State.RoutesByServiceTest do
       assert for_service_ids_and_types([@service.id], ["some_other_type"]) == []
       assert for_service_ids_and_types(["some_other_service"], [@route.type]) == []
     end
+
+    test "returns a list with no duplicates if multiple routes have the same type" do
```
paulswartz, line 85: 👍

Can we also add a test that there are no duplicates if there are multiple service IDs which have the same route? For example, if "service A" and "service B" both have "route 1".

apps/state/test/state/routes_by_service_test.exs
```diff
@@ -80,6 +81,22 @@ defmodule State.RoutesByServiceTest do
...
+      assert Enum.sort(
+               for_service_ids_and_types([@service.id], [
+                 @route.type
+               ])
+             ) == [@route.id]
+    end
+
+    test "returns a list with no duplicates if multiple service ids have the same route" do
+      assert Enum.sort(
```
paulswartz, line 94: If there's only one result, there's no need to sort.

apps/state/lib/state/routes_by_service.ex
```diff
@@ -104,6 +104,9 @@ defmodule State.RoutesByService do
       end)
       |> Enum.map(fn {x, y} -> {x, Enum.map(y, fn x -> x.route_id end)} end)
 
+    #IO.puts("items")
+    #IO.inspect(items)
```
paulswartz, line 108: These should be deleted. Also a tip, you can use `IO.inspect(items, label: "items")` to let Elixir annotate the output.

paulswartz (commented):
Looks like it needs formatting.

apps/state/lib/state/routes_by_service.ex
paulswartz, line 48: Can you add back the `uniq` step when creating the ETS data? Doing that makes this lookup 250x faster:
```
## State.RouteUniqBench
benchmark name       iterations   average time
flat_map dedup         10000000   0.13 µs/op
uniq flat_map dedup    10000000   0.38 µs/op
flat_map ets              50000   34.25 µs/op
uniq flat_map ets         10000   101.70 µs/op
```

paulswartz (approved):
🍰

PR #392, opened 2021-07-01
Update to Erlang 24 / Elixir 1.12

losvedir:
Several commits, should be reviewable independently. The `nimble_parsec` one isn't really related, just more of a bonus. The `ex_aws` update was necessary because Erlang 24 removed a deprecated crypto function that was used by the version we were on.

Most of the updates were around dealing with the deprecation warning from using the old `import Supervisor.Spec` approach. We had a lot of `worker(Module, [])` children, which invokes `Module.start_link/0`. But the new way has normalized using `Module.start_link/1`, and indeed that's what the new `Module` or `{Module, args}` syntax calls (with the former passing a `[]` as the only argument). So I had to update a number of `start_link/0`'s to be `start_link(_opts \\ [])`.

The upgrade to Elixir 1.12 (specifically, [this feature](https://elixir-lang.org/blog/2020/10/06/elixir-v1-11-0-released/#compiler-checks-data-constructors) from Elixir 1.11) also caught a bug! It's smarter about struct handling, and identified that we were accessing fields that didn't exist. Looking in our git history, it looks like that function never worked, so I removed it.


apps/api_web/lib/api_web/event_stream/supervisor.ex
```diff
@@ -76,4 +71,16 @@ defmodule ApiWeb.EventStream.Supervisor do
...
+
+  @impl Supervisor
+  def init(_init_arg) do
+    children = [
+      {Registry, name: ServerRegistry, keys: :unique},
+      {DynamicSupervisor, name: ServerSupervisor, strategy: :one_for_one}
+    ]
+
+    Supervisor.init(children, strategy: :one_for_all)
+  end
```
paulswartz, line 85: Can you use `Supervisor.start_link/2` instead of `/3` with an `init` callback? 

https://hexdocs.pm/elixir/Supervisor.html#start_link/2

edit: it looks like that's what was happening before: what was the error here?

apps/api_web/lib/request_track.ex
```diff
@@ -2,6 +2,9 @@ defmodule RequestTrack do
   @moduledoc """
   Track the number of outstanding requests by API key.
   """
+
+  use GenServer
```
paulswartz, line 6: 👍

apps/fetch/lib/fetch.ex
```diff
@@ -3,6 +3,8 @@ defmodule Fetch do
   Fetches URLs from the internet
   """
 
+  use DynamicSupervisor
```
paulswartz, line 6: 👍

paulswartz (commented):
Can you throw this up on dev-green (if you haven't already)?

paulswartz:
1.12.2 just dropped: https://github.com/elixir-lang/elixir/releases/tag/v1.12.2

apps/api_web/lib/api_web/event_stream/supervisor.ex
losvedir: Changing from `supervisor(EventStream.Supervisor, [])` to `EventStream.Supervisor` means its supervisor would invoke `EventStream.Supervisor.child_spec/1`. However, this module did not `use Supervisor` previously, so that function was not defined. I could have defined a `child_spec/1` myself, but I opted instead to `use Supervisor` since it looked like a standard supervisor and I thought that was more conventional.

But that means its supervisor will be expecting an `EventStream.Supervisor.start_link/1`, contra its existing `start_link/0`. The minimal fix is updating the existing one to `start_link/1`, but the compiler logs a warning that the module, because of the `use Supervisor`, is not implementing the supervisor behaviour because of the missing `init/1`. And indeed, I read in [the docs](https://hexdocs.pm/elixir/Supervisor.html#module-module-based-supervisors):

> A general guideline is to use the supervisor without a callback module only at the top of your supervision tree, generally in the `Application.start/2` callback. We recommend using module-based supervisors for any other supervisor in your application, so they can run as a child of another supervisor in the tree. 

So I decided to stick with `use Supervisor` and switch to a module-based approach. The alternative is to drop the `use Supervisor`, implement a `child_spec/1`, and keep the existing `start_link` based configuration. This upgrade prompted me to re-read the Supervisor docs, and I _think_ what I have here is now the preferred approach, but I'm not really married to it. What's the reason you prefer `start_link`?

paulswartz: Thanks, I hadn't seen that in the docs! If this is what they recommend, then it's probably what we should do.

Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN mix release
 # The one the elixir image was built with
 FROM alpine:3.11.6
```
paulswartz, line 24: This also needs to be updated to 3.13.3.

paulswartz:
The failure looks like something that can be added to `.dialyzer.ignore-warnings`: the Phoenix Router tends to generate code that Dialyzer doesn't like but is otherwise fine.

losvedir:
Okay, looks like CI is finally appeased. This branch is on dev green and seems to work fine. I ran our locustfile in /load_tests and didn't see any error responses on the various endpoints.

Is there anything you'd want me to do to measure performance before-and-after with API prod? Might be a little interesting to see what Erlang/OTP 24 gets us.

paulswartz:
There's a dashboard in Splunk with performance: I'll be curious to see what it looks like!

paulswartz (approved):
🍰

PR #430, opened 2021-09-15
fix: add Haverhill-Andover shuttle route pattern prefix overrides

bklebe:
Adds route pattern prefix overrides for the Haverhill-Andover shuttles.


jfabi (commented):
https://green.dev.mbtace.com/schedules/CR-Haverhill/timetable?date=2021-09-20 now looks a lot better (you might need to cache-clear or incognito to see it)! (Well, it's still a bit of a mess because of the sheer number of trips, but at least all the expected trips and stops appear.)

jfabi:
Only thing here is to adjust the `route_id` here if you change it in https://github.com/mbta/gtfs_creator/pull/1321.

jfabi (approved):
🎉

PR #434, opened 2021-10-14
feat: AWS SecretsManager support

losvedir:
This adds support for pulling environment variable-like data out of AWS Secrets Manager. Largely modeled on what we did for Arrow (https://github.com/mbta/arrow/pull/180) which in turn was based on Delta, but with two differences:

* I opted to have the app crash and fail to start up if we try to load the secrets and that fails. This felt better / more secure to me. If the AWS environment isn't configured correctly or something, better to noisily 💥 so we can fix it, than to silently move on, perhaps with empty data.
* Since the API is an umbrella app, I made a new application in the umbrella for this. I wasn't super sure this was the way to go. The only data in it right now is `secret-key-base`, used by `api_web`, but subsequently I'll be adding data for the `state_mediator` app. Since multiple apps would take advantage of this, it felt like it should be outside any one of them. But it didn't look like we could have "umbrella as a whole" modules, so I made another app.

For now, I'm only putting `SECRET_KEY_BASE` in secrets manager, since that's the only real secret the API has. But the motivation for this change is to add the Keolis firebase credentials with my CR crowding task. The credentials are a giant JSON object, which doesn't work in ElasticBeanstalk (whitespace and/or special character issues), while my attempted base64 encode was too large for EB env vars. Double encoding the JSON might have worked (that's what I had to do to get it running locally), but felt messy and annoying if we need to change credentials in the future.

SecretsManager values transparently become environment variables in ECS, so technically dev-blue doesn't need this, but alas prod and the other environments do for now. Firebase JSON credentials in Secrets Manager -> ECS environment variables works, as demonstrated by the `commuter_rail_boarding` app.

I have a PR https://github.com/mbta/devops/pull/173 to add the secrets to the API via terraform.


config/releases.exs
```diff
@@ -15,8 +15,6 @@ config :api_accounts,
   table_prefix: System.get_env("DYNAMO_TABLE_PREFIX"),
   migrate_on_start: true
 
-config :api_web, ApiWeb.Endpoint, secret_key_base: System.get_env("SECRET_KEY_BASE")
-
 config :api_web, :signing_salt, System.get_env("SIGNING_SALT")
```
paulswartz, line 18: question: why not this secret as well?

mix.exs
```diff
@@ -25,6 +25,7 @@ defmodule ApiUmbrella.Mixfile do
           applications: [
             runtime_tools: :permanent,
             alb_monitor: :permanent,
+            config_providers: :permanent,
```
paulswartz, line 28: question: Does this actually have any servers that need to run? It seems like it's more of a library application which wouldn't need to be listed here, or possibly a `:temporary` application because it shuts down after loading the configuration?

config/releases.exs
losvedir: Heh, just to minimize the amount of AWS finagling I had to do. I had toyed with not moving any of the EB env vars, and just setting up the config provider for the future Keolis stuff, but figured I should have _something_ so I can test it end to end. But you're right, that I might as well move all the secrets in. Will updated.

mix.exs
losvedir: I don't _think_ it has any servers that run (though the release approach something like "start up the app and do some configuration and write it down, and then restart it" I'm still a little hazy on). My main concern was the release stripping out that code. I thought if the applications weren't mentioned, it might not be compiled into the final artifact, but I can try removing and see if it works when I deploy to dev-blue and dev-green.

paulswartz (approved):
🍰

config/releases.exs
```diff
@@ -15,10 +15,6 @@ config :api_accounts,
   table_prefix: System.get_env("DYNAMO_TABLE_PREFIX"),
   migrate_on_start: true
 
-config :api_web, ApiWeb.Endpoint, secret_key_base: System.get_env("SECRET_KEY_BASE")
-
-config :api_web, :signing_salt, System.get_env("SIGNING_SALT")
-
```
paulswartz, line 21: suggestion: You might want to keep these but make them optional. Otherwise you might have a hard time getting the ElasticBeanstalk deploy to take the first time during the transition, and when the API migrates to ECS these will go back to being loaded by environment variables anyways.

PR #441, opened 2021-10-29
test: adjust assertions for new GL core terminals

digitalcora:
**Asana:** [🚧 Remove old GL terminals from API tests](https://app.asana.com/0/584764604969369/1201227620034052)

### DIFF
--- a/apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
+++ b/apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
@@ -61,9 +61,9 @@ defmodule StateMediator.Integration.GtfsTest do
       assert_first_last_stop_id("CR-Haverhill", "place-north", "place-WR-0329")
       assert_first_last_stop_id("CR-Lowell", "place-north", "place-NHRML-0254")
       assert_first_last_stop_id("CR-Kingston", "place-sstat", "place-KB-0351")
-      assert_first_last_stop_id("Green-B", ["place-pktrm", "place-gover"], "place-lake")
-      assert_first_last_stop_id("Green-C", ["place-gover", "place-north"], "place-clmnl")
-      assert_first_last_stop_id("Green-D", ["place-gover", "place-north"], "place-river")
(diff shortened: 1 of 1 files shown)


apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
```diff
@@ -61,9 +61,9 @@ defmodule StateMediator.Integration.GtfsTest do
       assert_first_last_stop_id("CR-Haverhill", "place-north", "place-WR-0329")
       assert_first_last_stop_id("CR-Lowell", "place-north", "place-NHRML-0254")
       assert_first_last_stop_id("CR-Kingston", "place-sstat", "place-KB-0351")
-      assert_first_last_stop_id("Green-B", ["place-pktrm", "place-gover"], "place-lake")
-      assert_first_last_stop_id("Green-C", ["place-gover", "place-north"], "place-clmnl")
-      assert_first_last_stop_id("Green-D", ["place-gover", "place-north"], "place-river")
+      assert_first_last_stop_id("Green-B", ["place-gover"], "place-lake")
```
jfabi, line 64: The "first stops" here for the lines you changed no longer need to be in a list form since they've been reduced to only one entry.

jfabi (commented):
This looks fine, excepting the one comment made, though note that if this branch doesn't pass CI right now, it almost certainly will on Monday (note Asana task's ⚠️) once the pre-terminal realignment trips fully drop out of the feed.

jfabi:
It looks like this branch is still failing CI. I want to say it's because it seems (according to the README.md) to be using  the prod GTFS version by default (which still hasn't been deployed since last week). So, at least a couple of options:
- You can test your https://github.com/mbta/gtfs_creator/pull/1354 branch against this `api` branch and, 
- We can either wait until GTFS-prod is redeployed, or, if GTFS CI won't pass (more likely), we can force-merging this branch first.

jfabi (approved):
🏁

PR #461, opened 2022-01-05
fix: better enforce restrictions on `occupancies`

digitalcora:
**Asana:** [🍎](https://app.asana.com/0/584764604969369/1201382502692999)

The `occupancies` relationship on Trips was only intended to be accessible when the client has opted into experimental features, but this restriction only worked when trips were accessed via the `/trips` endpoints. If loading trips via another resource's endpoints, such as `/schedules?include=trip.occupancies`, the restriction did not apply.

This patch addresses the issue in basically the same way we handle other "conditionally existing" relationships: if accessed indirectly, the request will still be allowed, but the relationship will not contain any data. This does not attempt to address the broader inconsistency that validation of the `include` parameter only works at the "top level" of the include tree.

**Note:** This includes, as separate commits, some refactors to `Params.validate_includes` I did as part of investigating other approaches for this task. The actual operative change does not involve `validate_includes` at all.

### DIFF
--- a/apps/api_web/lib/api_web/views/trip_view.ex
+++ b/apps/api_web/lib/api_web/views/trip_view.ex
@@ -107,13 +107,17 @@ defmodule ApiWeb.TripView do
     end
   end
 
-  @spec occupancies(Model.Trip.t(), Plug.Conn.t()) :: [Model.CommuterRailOccupancy.t()] | nil
+  @spec occupancies(Model.Trip.t(), Plug.Conn.t()) :: [Model.CommuterRailOccupancy.t()]
   def occupancies(%{name: nil}, _conn) do
-    nil
+    []
   end
 
-  def occupancies(%{name: name}, _conn) do
+  def occupancies(%{name: name}, %{assigns: %{experimental_features_enabled?: true}}) do
     CommuterRailOccupancy.by_trip_name(name)
+  end
+
+  def occupancies(_trip, _conn) do
+    []
   end
 
   defp optional_predictions(trip_id, conn) do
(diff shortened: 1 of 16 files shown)


apps/api_web/lib/api_web/views/trip_view.ex
```diff
@@ -109,13 +109,17 @@ defmodule ApiWeb.TripView do
 
   @spec occupancies(Model.Trip.t(), Plug.Conn.t()) :: [Model.CommuterRailOccupancy.t()] | nil
```
paulswartz, line 110: question: can this still return `nil`?

digitalcora: Ah, I changed it from `nil` to `[]` to align with the type of the relationship, but forgot to change the spec!

losvedir (approved):
Looks good!

Last commit message is "wip:" , though, in case you aren't going to squash the commits on merge.

digitalcora:
I like to leave wip commits until it's time to merge, so reviewers can see the specific changes made in response to feedback. It'll be fixed up!

losvedir:
Ah, gotcha. I usually do fixups for that so I can `--autosquash`, but I guess your way does say what exactly was changed (not just which commit the fix was for).

digitalcora:
Yeah, I think the main thing keeping me from using autosquash regularly is just that it doesn't allow for the to-be-squashed commit to have its own identifying title. I wish it could work like `fixup! 5bcd90a (arbitrary title)`, but for some reason it only operates based on commit messages.

PR #469, opened 2022-02-04
chore: Exclude Green Line shuttle from consideration in stops-on-route

jfabi:
#### Summary of changes

**Asana Ticket:** [🚧 Upcoming Lechmere–Gov't Center shuttles](https://app.asana.com/0/584764604969369/1201695996672166/f)

In support of GTFS-creator pull request https://github.com/mbta/gtfs_creator/pull/1430. Ensures that stops on the new `Shuttle-GovernmentCenterLechmere` route do not make their way onto the API's stops-on-route.


digitalcora (approved):
After hours of tracing through `api` code I'm no closer to understanding why this is needed, but it seems clear that it is needed...

PR #477, opened 2022-02-22
feat: Add prod deploy yaml script

uchukwu:
Update the GitHub Action to deploy prod to ECS 🚀

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -1,0 +1,29 @@
+name: Deploy to Prod (ECS)
+
+on: workflow_dispatch
+
+jobs:
+  deploy:
+    name: Deploy
+    runs-on: ubuntu-latest
+    environment: prod
+    concurrency: prod
+    env:
+      ECS_CLUSTER: api
+      ECS_SERVICE: api
+
+    steps:
+      - uses: actions/checkout@v2
+      - name: Deploy to ECS
+        uses: mbta/actions/deploy-ecs@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          ecs-cluster: ${{ env.ECS_CLUSTER }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-prod.yml
```diff
@@ -1,6 +1,9 @@
-name: Deploy to Prod
+name: Deploy to Prod (ECS)
 
-on: workflow_dispatch
+on:
+  workflow_dispatch:
+  push:
```
losvedir, line 5: For prod, we actually *don't* want this. Basically, your workflows here should maintain the previous behavior. I had you add the `push: branches: [master]` bit to the dev one because the old dev one had it.

This `on:` section says when the workflow should run. `on: workflow_dispatch` means the workflow should run when we manually trigger it via the GitHub interface, going to "Actions", and clicking the "Deploy to Prod". `on: push: branches: [master]` means it *automatically* run when new code is added to the master branch (i.e.: a PR is merged). We want that for the dev environment, but not for prod; for prod we only deploy when we're around to monitor for it, and after searching the dev logs a bit to make sure we haven't introduced any issues.

losvedir:
Actually, can you add this as a new workflow, but leave the old one around for now? That way we can deploy to the old prod in Elastic Beanstalk or the new prod in ECS at our choosing. We just replaced dev and dev-green workflows because we were ready to switch to the new ECS environment immediately. However, prod we might have the old and new for a little while (at least until Ian gets back next week), and it would be good to have the ability to push to the old prod (which is what everyone is still using) until we've switched.

.github/workflows/deploy-prod.yml
uchukwu: Complete 👍

uchukwu:
> Actually, can you add this as a new workflow, but leave the old one around for now? That way we can deploy to the old prod in Elastic Beanstalk or the new prod in ECS at our choosing. We just replaced dev and dev-green workflows because we were ready to switch to the new ECS environment immediately. However, prod we might have the old and new for a little while (at least until Ian gets back next week), and it would be good to have the ability to push to the old prod (which is what everyone is still using) until we've switched.

Complete 👍

losvedir (approved):
LGTM

PR #478, opened 2022-02-23
fix: API prod service is api-prod

losvedir:
The deploy to the new ECS service [failed](https://github.com/mbta/api/runs/5309879271?check_suite_focus=true). The service in ECS is actually `api-prod` rather than just plain `api`.


uchukwu (approved):
LGTM 🍰

PR #479, opened 2022-02-23
fix: ECS deploy needs build-push step

losvedir:
One more issue with the deploy: the `deploy-ecs` action expected a docker tag based on a previous step: `docker-tag: ${{ steps.build-push.outputs.docker-tag }}`, but we didn't have a `build-push` step! We could also just calculate the tag ourselves and pass that in, without requiring the `build-push` step, but I checked other repos (glides and skate) and they do the `build-push` step even in prod, so I figured I'd be consistent with them.

And I deployed to prod using this branch and it [looks like it's going through](https://github.com/mbta/api/actions/runs/1889960742).

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -14,6 +14,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


uchukwu (approved):
Interesting. LGTM 🍰

PR #484, opened 2022-03-04
fix: Deploy to Prod (EB)

losvedir:
Previously, the prod deploy to EB relied on a previous dev deploy to have setup the "deployment package" ("Application Version" in the EB console). But now that dev goes to ECS, prod will have to do a couple of the steps that dev used to do.

### DIFF
--- a/.github/workflows/prod.yml
+++ b/.github/workflows/prod.yml
@@ -15,6 +15,16 @@ jobs:
       - uses: actions/checkout@v2
       - id: metadata
         uses: mbta/actions/commit-metadata@v1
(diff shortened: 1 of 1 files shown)


paulswartz (approved):
@losvedir looks fine to me: could you try doing the deploy from this branch to see if it works?

losvedir:
Yep, I did before opening the PR. The commit doesn't match because I ended up tweaking some (I believe) inconsequential things but I just kicked off a fresh deploy from the current commit to make sure.

paulswartz:
looks good!

PR #486, opened 2022-03-11
chore: Remove old prod ElasticBeanstalk deploy workflow

losvedir:
Now that we've switched to ECS entirely, we don't need this workflow anymore. (The ECS one is `deploy-prod.yml`.)


digitalcora (approved):
➖ ✅

PR #488, opened 2022-03-15
chore: Supress GLX stops, shapes from dotcom pre-opening

jfabi:
#### Summary of changes

**Asana Ticket:** [❎ Publish v1 static data for March 21 opening](https://app.asana.com/0/584764604969369/1201964847789047/f)

_Companion to the GTFS-creator pull request, https://github.com/mbta/gtfs_creator/pull/1464._

Prevents Science Park, Lechmere, and Union Square from showing up on the Green Line D or Green Line E [line diagrams on MBTA.com](https://green.dev.mbtace.com/schedules/Green-E/line). This matches the current desire for MBTA.com to not show GLX information pre-opening on the line pages before opening.

Of course, we'll still want to show the GLX stations on the site once it opens, so there will have to be another pull request and additional deploy Sunday/Monday.

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -153,7 +153,10 @@ config :state, :shape,
   },
   suffix_overrides: %{
     # shuttles are all -1 priority
-    "-S" => -1
+    "-S" => -1,
+    # Prior to GLX opening on March 21st, suppress Union Square shapes
+    "8000001" => -1,
+    "8000002" => -1
   }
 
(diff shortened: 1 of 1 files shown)


losvedir (approved):
Seems fine, though I'm not thrilled about the Sunday deploy.

Will be nice once canonical routes are in and we don't have to mess with this file anymore! But even then it occurs to me that dotcom might want to hide a stop that's on the canonical route, for marketing/timing reasons. But at that point, I think it might be on dotcom to have some sort of explicit configuration for that sort of thing, rather than the API.

PR #489, opened 2022-03-16
fix: Temporary code for removing tenable api users

uchukwu:
Temporary code for removal of accounts matching `wasscan*@tenable.com*` added in api_accounts.exs and additional code added in api_accounts_test.exs for testing the function.

### DIFF
--- a/apps/api_accounts/lib/api_accounts.ex
+++ b/apps/api_accounts/lib/api_accounts.ex
@@ -4,6 +4,7 @@ defmodule ApiAccounts do
   """
 
   alias ApiAccounts.{Changeset, Dynamo, Key, NoResultsError, User}
+  require Logger
 
   @default_version Application.compile_env(:api_web, :versions)[:default]
 
@@ -213,6 +214,32 @@ defmodule ApiAccounts do
       |> Task.async_stream(&delete_key/1)
       |> Stream.run()
     end
+  end
+
+  @doc """
+  Temporary Function: Deletes tenable users.
+
+  Any keys belonging to tenable users will also be deleted.
+
+  ## Examples
+
+      iex> delete_tenable_users()
+      :ok
+
+      iex> delete_tenable_users()
+      {:error, ...}
+
+  """
+  @spec delete_tenable_users() :: :ok | {:error, any}
+  def delete_tenable_users do
+    list_users()
+    |> Enum.each(fn user ->
+      if user.email =~ ~r/wasscan.*@tenable.com/ do
+        Logger.info("delete_tenable_users, deleting #{user.email}")
+        result = delete_user(user)
+        Logger.info("delete_tenable_users, result=#{inspect(result)}")
+      end
+    end)
   end
 
   @doc """
--- a/apps/api_accounts/test/api_accounts_test.exs
+++ b/apps/api_accounts/test/api_accounts_test.exs
@@ -11,6 +11,31 @@ defmodule ApiAccountsTest do
     email: "test@test.com"
   }
 
+  @valid_attrs_2 %{
+    active: true,
+    blocked: true,
+    join_date: DateTime.from_naive!(~N[2010-04-17 14:00:00], "Etc/UTC"),
+    phone: "some phone 2",
+    username: "some username 2",
+    email: "test2@test.com"
+  }
+
+  @invalid_wasscan_tenable_email_attr %{
+    active: false,
+    blocked: false,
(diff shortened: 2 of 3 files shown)


apps/api_accounts/test/api_accounts_test.exs
```diff
@@ -123,6 +176,21 @@ defmodule ApiAccountsTest do
...
     end
 
+    test "delete_tenable_user/1 deletes the user wasscan tenable" do
+      valid_users = list_of_valid_users()
+      valid_and_invalid_users = add_invalid_users()
+      assert ApiAccounts.list_users() == valid_and_invalid_users
+
+      for user <- valid_and_invalid_users do
+        for _ <- 1..5 do
+          {:ok, _} = ApiAccounts.create_key(user)
```
losvedir, line 186: Since you're creating keys for the users (which is a good idea), can you add an assertion that the keys are deleted as well?

losvedir (commented):
The function here looks good. Let's talk about how to actually activate it. I think the best thing to do is to trigger your function when the app starts up, and then we can deploy your branch to dev-green to test it against the wasscan users you're creating there (and verifying it doesn't touch the _other_ users), and then we can deploy the branch to prod.

One of the issues with this approach is a potential race condition, since there are actually a number of API server instances running, so when you deploy a branch, it will deploy this code several times. If multiple servers do `ApiAccounts.list_users()` and then they all go about deleting them, subsequent servers may try to delete a user which isn't present. However, I believe that this will simply return an `:error`, which should be fine; we'll log it and move on. My only concern would be if it actually crashes, rather than returning an error, but the typespec doesn't mention that, and [the docs](https://hexdocs.pm/aws/AWS.DynamoDB.html#delete_item/3) say:

> Unless you specify conditions, the `DeleteItem` is an idempotent operation; running it multiple times on the same item or attribute does not result in an error response.

so I think we're okay on that front.

As for how and where to trigger your function, let's make it run when the `ApiAccounts` application starts up. We can use Elixir's [`Task`](https://hexdocs.pm/elixir/Task.html) for this; that's a little more straightforward than a whole `GenServer`, like we talked about this morning. You would update `ApiAccounts.Application` to have a special "task" as a child that runs after the other children start up.

```elixir
Supervisor.start_link(
  [
    :hackney_pool.child_spec(:ex_aws_pool, []),
    ApiAccounts.Keys,
    {Task, fn ->
      Logger.info("beginning_delete_users_task")
      ApiAccounts.delete_tenable_users()
      Logger.info("finished_delete_users_task")
    end}
  ],
  # ...
)
```

Since we're logging here, you'll have to add `require Logger` in `ApiAccounts.Application` up towards the top.

apps/api_accounts/test/api_accounts_test.exs
uchukwu: Added :+1:

apps/api_accounts/lib/api_accounts.ex
```diff
@@ -215,6 +215,30 @@ defmodule ApiAccounts do
...
+      iex> delete_tenable_users()
+      {:error, ...}
+
+  """
+  @spec delete_tenable_users() :: :ok | {:error, any}
+  def delete_tenable_users do
+    list_users()
+    |> Enum.each(fn user ->
+      if user.email =~ ~r/wasscan.*@tenable.com/ do
+        delete_user(user)
```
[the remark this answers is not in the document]
uchukwu: Added :+1:

apps/api_accounts/test/api_accounts_test.exs
```diff
@@ -123,6 +176,34 @@ defmodule ApiAccountsTest do
...
+        end
+      end
+
+      assert ApiAccounts.delete_tenable_users() == :ok
+      assert ApiAccounts.list_users() == valid_users
+
+      invalid_users =
+        Enum.map(valid_and_invalid_users, fn x ->
```
losvedir, line 194: Since this code isn't getting merged, you don't have to fix it, but just FYI, what would be better here is `Enum.filter`. If you want to go through a list and reduce its size by some criterion, that's a filter; if you want to go through a list and return a new list _of the same size_ but with every element transformed, that's a map.

So this could be:

```elixir
valid_and_invalid_users
|> Enum.filter(fn x -> x.email =~ ~r/wasscan.*@tenable.com/ end)
|> Enum.each(fn user -> 
  assert ApiAccounts.list_keys_for_user(user) == []
end)
```

losvedir:
I deployed this branch and it cleaned up the users. Thanks! Closing it now without merging.

PR #493, opened 2022-03-25
chore: Update Elixir 1.12 -> 1.13 and erlang patch version

losvedir:
Updates to the latest Elixir and Erlang at this time.

The update to Elixir 1.13 finally deprecated the old style `use Mix.Config` in favor of the `import Config` introduced in Elixir 1.9. There sure are a lot of config files in an umbrella app... The `Config` docs have a [little section](https://hexdocs.pm/elixir/1.13/Config.html#module-migrating-from-use-mix-config) on migrating, which I followed.

Credo needed to be updated to work with Elixir 1.13, so I did that. It pulled in a few new rules, so I updated the code to handle them, except for one rule (no piping into anonymous functions), which we do in the API a fair bit and which I didn't have a real problem with, and so I just disabled that rule.

Lastly, Elixir 1.13 updated `mix test --cover` to fail (return status code 3) if the test coverage was below 90%, which one of the apps was, so I wrote a little test to get us back over 90% (rather than changing the default threshold).


digitalcora (approved):
General suggestion: I think it'd be better to squash the code coverage commit into the Elixir upgrade commit, with a note in the commit message explaining the change. That way each individual commit has a passing test suite/CI (the same as how you did the Credo change). But looks good!

PR #495, opened 2022-03-30
feat: Update API to use ENV vars from Secrets Manager

uchukwu:
Asana ticket: [🍎 Update API to use ENV vars from Secrets Manager](https://app.asana.com/0/584764604969369/1201852904717912/f)

Add script in release.exs for [run-time configuration](https://hexdocs.pm/mix/Mix.Tasks.Release.html#module-run-time-configuration) of AWS Secrets. Removed ConfigProvider, given that Secrets are assigned environment variables through terraform, and are to be retrieved through run-time configuration. 

Still need further assistance/guidance to be sure this is correct or being setup along the right direction.

Additionally, still requires removal of permission in terraform for api to query Secret Manager.


config/releases.exs
```diff
@@ -28,3 +28,8 @@ config :state_mediator, Realtime,
 config :state_mediator, State.Prediction, source: System.get_env("MBTA_TRIP_SOURCE")
 
 config :state_mediator, State.Vehicle, source: System.get_env("MBTA_VEHICLE_SOURCE")
+
+config :api_secrets,
```
losvedir, line 32: If you look in the deleted `secrets_manager.ex` file, you should see how the values should be configured. I think:

```ex
config :api_web, signing_salt: value
config : api_web, ApiWeb.Endpoint, secret_key_base: value
config :state_mediator, :commuter_rail_crowding, firebase_credentials: value
```

where `value` is the `System.fetch_env!`.

You can try deploying your branch to api-dev-green to see if it works.

config/releases.exs
```diff
@@ -28,3 +28,8 @@ config :state_mediator, Realtime,
...
+  firebase_credentials: System.fetch_env!("CR_CROWDING_FIREBASE_CREDENTIALS")
+  secret_key_base: System.fetch_env!("SECRET_KEY_BASE")
+  signing_salt: System.fetch_env!("SIGNING_SALT")
```
losvedir, line 35: There should be a newline at the end of every file. (That's what the red circle in the diff here means; the lack of a newline.) You can configure your editor to automatically ensure that when saving. In VSCode, for example, if you open the Settings (cmd-comma) and then search for "newline" you'll see `Files: Insert Final Newline` to check.

uchukwu:
I want to be careful about resolving the conflict here.
 
It's complaining about the removal of `config_providers/secrets_manager.exs`. Do you understand why that conflict arises and am I to follow the steps under `Use the command line`? Is it directing my to rebase my current branch with the remote copy of the branch through `git checkout -b uc-config-api-secrets origin/uc-config-api-secrets`, followed by merging my local `master` to my rebased branch? And following that, in Step 2, am I to merge my merged rebased branch into `master` and push that in? DON'T THINNK YOU WANT ME MERGING INTO MASTER :/ 

In conclusion, I'm confused :thinking_face: and I believe this is causing Github Action not to run

losvedir:
Ah, yeah, the conflict is because you deleted `secrets_manager_test.exs` while I just updated that file in #493 . So it doesn't know if deleting it or updating it is correct. (Deleting it is correct.)

You're right that you shouldn't be merging to `master`! I'll pair with you because I'm not sure what state your `git` is in, but the basic idea is you should try to do a `git rebase master` on your branch, it will alert you to the conflict, you can resolve the conflict by deleting the file again with `git rm secrets_manager_test.exs`, and then finishing the rebase with `git rebase --continue`. At that point you can force push up your branch to your branch on github.

PR #508, opened 2022-05-17
chore: Adjust Rockport Branch GTFS integration test

jfabi:
_Related GTFS-creator pull request: https://github.com/mbta/gtfs_creator/pull/1535_

**Related Asana task:** [🏝 Summer CR rating effective May 23](https://app.asana.com/0/584764604969369/1202203324773368/f)

Now that the Rockport Branch of the Newburyport/Rockport Line will be going all the way to Rockport...the API's integration tests can believe otherwise.


jfabi:
**Note on this PR:** The validation is failing because the GTFS data it's looking at doesn't yet have Rockport trips. This Rockport service is introduced in https://github.com/mbta/gtfs_creator/pull/1535.

So it's a bit of a chicken-and-egg situation.

I think the preferred solution would be, upon approval of this API branch, for @losvedir to force-merge it, and then we let the GTFS-creator branch pass CI on the API `master` branch. Thoughts?

bklebe:
Yes, that's what I was expecting.

bklebe (approved):

PR #518, opened 2022-07-23
chore: Reconfigure Green Line branches for Medford/Tufts opening

jfabi:
#### Summary of changes

**Asana Ticket:** [❎ Medford/Tufts Dec 12 opening in GTFS-static](https://app.asana.com/0/584764604969369/1203503515189544/f)

These changes are currently up on https://dev-blue.mbtace.com/ and have been vetted by the dotcom dev team, but an extra pair of eyes can't ~help~ hurt! Note that this pull request must be merged in before the related GTFS-creator, https://github.com/mbta/gtfs_creator/pull/1578.

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -436,6 +436,11 @@ config :state, :stops_on_route,
       "place-haecl"
     ],
     {"Green-D", 0} => [
+      "place-mdftf",
+      "place-balsq",
+      "place-mgngl",
+      "place-gilmn",
+      "place-esomr",
       "9070150",
       "9434",
       "1521",
@@ -459,7 +464,12 @@ config :state, :stops_on_route,
       "9070180",
       "1804",
       "1807",
-      "9070150"
+      "9070150",
+      "place-esomr",
+      "place-gilmn",
+      "place-mgngl",
+      "place-balsq",
+      "place-mdftf"
     ],
     {"Green-E", 0} => [
       "place-unsqu",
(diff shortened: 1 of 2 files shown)


apps/state/config/config.exs
```diff
@@ -75,6 +75,10 @@ config :state, :shape,
     "831_0008" => 2,
     # Green-C
     "831_0009" => 2,
+    # Green-D (Medford/Tufts)
+    "8000020" => -1,
+    # Green-D (Medford/Tufts)
+    "8000021" => -1,
```
thecristen, line 81: I don't know enough about how this config works to evaluate: my guess is this adjusts these shape priority to `-1`. But why do that for these shapes? :O

clem-hertling: same question, the rest looks good

jfabi: Yeah, so these shapes in particular go Riverside <-> Medford/Tufts and are used for the atypical early-morning scheduled trips. We usually specify these unusual shapes here so that the API does not attempt to add the extra stops to the list of stops belonging to a given route-direction.

FWIW, though, the `stops_on_route` function is a bit of a black box to me and likely needs some retooling so that it doesn't need to be dependent on specifying `shape_id`s (eg, instead of relying only on `route_pattern_id`s).

clem-hertling: hmm, that sounds less than ideal, but also a problem for another day. makes sense in the meantime!

clem-hertling (approved):

PR #520, opened 2022-08-09
chore: Reconfigure Green Line branches for Fall 2022 start

jfabi:
_**In support of:**_ _[🍁 Compile GTFS for new rating](https://app.asana.com/0/584764604969369/1202725904167356/f)_ and _[🚧 Union Sq–Gov't Center shuttles Aug 22–Sept 18](https://app.asana.com/0/584764604969369/1202717862713833/f)_

When the Fall 2022 rating schedule data is introduced into GTFS, the Green Line D still terminates at North Station, and the Green Line E still terminates at Union Square. This commit ensures those branches continue to be displayed with their current terminals until the GLX surge happens later in August.


apps/state/config/config.exs
```diff
@@ -415,13 +427,22 @@ config :state, :stops_on_route,
       "21458",
       "9070206",
       "30203",
-      "4510"
+      "4510",
+      "9070503",
```
bfauble, line 431: why are these the bus stations vs the underlying stations since the shuttle stops are temporary?

jfabi: The shuttle bus stops above are technically _not_ part of their nearby/constituent stations, at least according to our data model, which follows GTFS best practice of only including stops as part of stations if they are enclosed within a building or otherwise off-street area.

A stop on the side of the general-traffic road, by this standard, isn't part of a station. Thus, it is these "random bus stops" are showing up as part of the Green Line.

bfauble (approved):

PR #523, opened 2022-08-19
🍎 🔱 Surface canonical trips/patterns in API

bfauble:
Asana Ticket: [🍎 🔱 Surface canonical trips/patterns in API](https://app.asana.com/0/584764604969369/1201746583170194/f)


apps/model/lib/model/route_pattern.ex
```diff
@@ -22,8 +23,9 @@ defmodule Model.RoutePattern do
           direction_id: Model.Direction.id(),
           name: String.t(),
           time_desc: String.t() | nil,
-          typicality: 0..4,
+          typicality: 0..5,
           sort_order: integer(),
-          representative_trip_id: Model.Trip.id()
+          representative_trip_id: Model.Trip.id(),
+          is_canonical: 0..2 | nil
```
bklebe, line 29: question: Why are we saying this could be possibly `nil`? I thought we were using `0` to represent `null` in JSON…am I missing something here?

apps/api_web/test/api_web/controllers/route_pattern_controller_test.exs
```diff
@@ -253,6 +254,34 @@ defmodule ApiWeb.RoutePatternControllerTest do
...
+               %{
+                 "id" => "rp1",
+                 "attributes" => %{"direction_id" => 0, "is_canonical" => true}
+               },
+               %{
+                 "id" => "rp2",
+                 "attributes" => %{"direction_id" => 1, "is_canonical" => true}
+               }
+             ] = json_response(conn, 200)["data"]
+    end
```
bklebe, line 283: suggestion: We should probably test route pattern filtering when `is_canonical = 0`.

apps/api_web/lib/api_web/params.ex
```diff
@@ -203,6 +203,13 @@ defmodule ApiWeb.Params do
   def route_types(%{"route_type" => route_types}), do: integer_values(route_types)
   def route_types(_), do: []
 
+  @doc """
+  Parse is_canonical filter param into integer
+  """
+  def is_canonical("true"), do: 1
+  def is_canonical("false"), do: 2
+  def is_canonical(_), do: nil
```
bklebe, line 211: question: Shouldn't a route with no canonical patterns be included by `filter[is_canonical]=false`? Looking at the RFC, `is_canonical = 0` indicates that a route has no canonical pattern defined. I see below in `RoutePattern.filter_by/1` you're filtering routes that don't match this number parsed out of the filter param, but it's not clear to me that this is the correct behavior, because it means that you'll never see routes with no canonical pattern defined when using this filter. Is that intended?

apps/api_web/lib/api_web/views/route_pattern_view.ex
```diff
@@ -25,10 +25,15 @@ defmodule ApiWeb.RoutePatternView do
...
+    :is_canonical
   ])
 
   def representative_trip(%{representative_trip_id: trip_id}, conn) do
     optional_relationship("representative_trip", trip_id, &State.Trip.by_primary_id/1, conn)
   end
+
+  def is_canonical(%{is_canonical: 1}, _), do: true
+  def is_canonical(%{is_canonical: 2}, _), do: false
+  def is_canonical(%{is_canonical: _}, _), do: nil
```
bklebe, line 38: question: Same question here as for `Params.is_canonical/1`.

bklebe (commented):
Looks good mostly, nice job on the tests. Just a few questions about how this new boolean filter parameter maps to the three-valued field it's filtering on.

apps/api_web/lib/api_web/params.ex
bfauble: agreed, will update based on the is true or everything else vs specifically false.

apps/state/lib/state/route_pattern.ex
```diff
@@ -78,4 +88,8 @@ defmodule State.RoutePattern do
     |> Enum.map(& &1.route_pattern_id)
     |> Enum.uniq()
   end
+
+  defp canonical_comparison(_, nil), do: true
+  defp canonical_comparison(1, bool), do: bool
+  defp canonical_comparison(_, bool), do: !bool
```
bklebe, line 94: question: do we need to accept `null` as a filter value?

bklebe (commented):
I'm not sure whether we intend to allow filtering on all of the possible values of `is_canonical`. My understanding was that this is a boolean filter, which accepts `true` or `false`, and that route patterns with `is_canonical: 0` would be returned when `false` is provided. The Asana ticket doesn't seem to be clear on the relationship between the field in the response and the filter parameter either, so we may need to clarify this with the rest of the team. Do you know?

apps/state/lib/state/route_pattern.ex
bfauble: so the possibilities this covers is if the client passes either nil or any non true/false value. The `filter_by` isn't hit if there's no `is_canonical` key, but nil, empty string, etc will all end up here as nil rather than just skipping. Following all the discussion that took place regarding this is kind of scattered, the `nil|true|false` definition was mentioned in the ticket's comments and some ad-hoc discussions

bfauble:
> I'm not sure whether we intend to allow filtering on all of the possible values of `is_canonical`. My understanding was that this is a boolean filter, which accepts `true` or `false`, and that route patterns with `is_canonical: 0` would be returned when `false` is provided. The Asana ticket doesn't seem to be clear on the relationship between the field in the response and the filter parameter either, so we may need to clarify this with the rest of the team. Do you know?

 In a similar boat as you where a lot of the discussions regarding `is_canonical` took place before my time so likely missing quite a bit of context regarding the nullable status of the `is_canonical` filter. I don’t have any super strong feelings regarding allowing `null` or not, but I think, from a client’s perspective at least, It makes it easier to make the request without necessarily needing to drop the key. There’s also other examples that use a similar approach within the Params module of effectively ‘nulling out’ bad data rather than returning an error. `direction_id`, `route_types`, and anything passed to `split_on_comma` will return either `nil` or `[]` when given bad data.

bklebe (approved):

apps/api_web/lib/api_web/controllers/route_pattern_controller.ex
```diff
@@ -10,7 +10,7 @@ defmodule ApiWeb.RoutePatternController do
 
   plug(:ensure_path_matches_version)
 
-  @filters ~w(id route direction_id stop)
+  @filters ~w(id is_canonical route direction_id stop)
```
paulswartz, line 13: question: why did we go with `is_canonical` vs only `canonical`? The other boolean we have is `timepoint` on `/schedules` and it doesn't have the `is_` prefix.

apps/parse/lib/parse/route_patterns.ex
```diff
@@ -12,7 +12,8 @@ defmodule Parse.RoutePatterns do
       time_desc: copy_string(row["route_pattern_time_desc"]),
       typicality: copy_int(row["route_pattern_typicality"]),
       sort_order: copy_int(row["route_pattern_sort_order"]),
-      representative_trip_id: copy_string(row["representative_trip_id"])
+      representative_trip_id: copy_string(row["representative_trip_id"]),
+      is_canonical: parse_is_canonical(row["canonical_route_pattern"])
```
paulswartz, line 16: question: is `is_canonical` the right name, since it's not a boolean?

apps/api_web/lib/api_web/controllers/route_pattern_controller.ex
```diff
@@ -210,6 +214,20 @@ defmodule ApiWeb.RoutePatternController do
...
+              """
+              Indicates whether or not the route pattern can be considered canonical and the default set of stops
+              for the given route and direction.
+
+              | Value | Description |
+              |-|-|
+              | `true` | Route pattern should be considered canonical for this route in this direction. If branching regularly occurs, this route-direction may have more than one canonical pattern. |
+              | `false` | Route pattern should be not considered canonical for this route in this direction. |
+              """,
+              "x-nullable": true
```
paulswartz, line 229: question: is this actually nullable?

apps/api_web/lib/api_web/controllers/route_pattern_controller.ex
jfabi: No objections here to dropping the `is_` prefix. 🙂

apps/model/lib/model/route_pattern.ex
```diff
@@ -22,8 +23,9 @@ defmodule Model.RoutePattern do
           direction_id: Model.Direction.id(),
           name: String.t(),
           time_desc: String.t() | nil,
-          typicality: 0..4,
+          typicality: 0..5,
```
paulswartz, line 26: suggestion: it looks like the documentation in `route_pattern_controller.ex` also needs to be updated with this new value.

apps/state/lib/state/route_pattern.ex
```diff
@@ -78,4 +88,8 @@ defmodule State.RoutePattern do
     |> Enum.map(& &1.route_pattern_id)
     |> Enum.uniq()
   end
+
+  defp canonical_comparison(_, nil), do: true
```
paulswartz, line 92: question: is it possible for this value to be `nil`? It doesn't appear to be possible from the types. If it's not possible, then it appears that this function is the same as `==`.

apps/state/test/state/route_pattern_test.exs
```diff
@@ -47,6 +47,11 @@ defmodule State.RoutePatternTest do
       assert filter_by(%{route_ids: ["route"], direction_id: 1}) == []
       assert filter_by(%{route_ids: ["route"], stop_ids: ["stop"]}) == [route_pattern]
       assert filter_by(%{route_ids: ["not_route"], stop_ids: ["stop"]}) == []
+      assert filter_by(%{canonical: true}) == [route_pattern]
+      assert filter_by(%{canonical: false}) == []
+      assert filter_by(%{canonical: nil}) == [route_pattern]
```
paulswartz, line 52: suggestion: this doesn't seem valid, as the `RoutePattern.filters()` type stays that `canonical` is a `boolean`.

bin/dynamodb/DynamoDBLocal_lib/DynamoDBPartiQLShared-1.0.jar
paulswartz, line 1: question: did you mean to check in the local DynamoDB Java implementation?

apps/state/lib/state/route_pattern.ex
paulswartz: From what I can see, there are two existing patterns for empty or otherwise invalid values:
- https://api-v3.mbta.com/routes/?filter[id]= doesn't match anything, and returns an empty list (this is the most common)
- https://api-v3.mbta.com/routes/?page[limit]= ignores the invalid value

There are also cases where we return a bad request error, such as https://api-v3.mbta.com/schedules/?filter[invalid]= but not based on the values that I can see.

paulswartz: I think that treating it the same as if nothing was specified is fine. But even with that approach, that's part of the query parameter parsing, not something that should be passed into `State.RoutePattern`: instead, `canonical` should not be a key in the map passed into `filter_by`.

apps/api_web/test/api_web/controllers/route_pattern_controller_test.exs
```diff
@@ -253,6 +254,110 @@ defmodule ApiWeb.RoutePatternControllerTest do
       assert [] == json_response(conn, 200)["data"]
     end
 
+    defp canonical_test_route_patterns do
+      [
+        %RoutePattern{id: "rp1", route_id: "route1", direction_id: 0, canonical: true},
```
paulswartz, line 259: question: should these also have updated `typicality` values?

apps/api_web/test/api_web/controllers/route_pattern_controller_test.exs
```diff
@@ -253,6 +254,110 @@ defmodule ApiWeb.RoutePatternControllerTest do
...
+    end
+
+    test "filtering by canonical null is treated the same as it not being included", %{conn: conn} do
+      State.RoutePattern.new_state(canonical_test_route_patterns())
+
+      conn =
+        get(
+          conn,
+          route_pattern_path(conn, :index, %{
+            "filter" => %{"canonical" => nil}
```
paulswartz, line 329: question: what's the actual path that's generated here?

shared-local-instance.db
paulswartz, line 1: suggestion: this also looks like something that shouldn't have been checked in.

suggestion: should it be added to `.gitignore` as well?

apps/api_web/lib/api_web/params.ex
```diff
@@ -282,7 +289,13 @@ defmodule ApiWeb.Params do
         bad_filters = Map.keys(filter) -- keys
 
         if conn.assigns.api_version < "2019-04-05" or bad_filters == [] do
-          {:ok, Map.take(filter, keys)}
+          valid_filters =
+            filter
+            |> Map.take(keys)
+            |> Enum.reject(fn {k, v} -> k == "canonical" and v == "" end)
```
paulswartz, line 295: suggestion: this doesn't seem like the right place for this filtering: I don't see anything else in here which knows about specific keys or values.

apps/api_web/lib/api_web/controllers/route_pattern_controller.ex
paulswartz, line 1: suggestion: instead of teaching `Params` about `canonical`, you can filter out invalid `canonical` values before you pass the filters into `RoutePattern.filter_by` on line 79.
```elixir
...
|> reject_invalid_canonical()
|> RoutePattern.filter_by()
...

defp reject_invalid_canonical(filters)
defp reject_invalid_canonical(%{canonical: nil} = filters), do: Map.delete(filters, :canonical)
defp reject_invalid_canonical(filters), do: filters
```

paulswartz (commented):
can you also push this up to dev-green when you have a chance, so the dotcom team can work with it?

jfabi:
@paulswartz Could you give another review of this PR when you have the chance?

jfabi:
> @jfabi the code looks good to me. Is it returning the information that you / @mbta/dotcom are expecting?

The API output certainly looks good from my perspective.

jfabi (approved):

PR #527, opened 2022-09-08
Remove max page limit of 100 from api requests

bfauble:
#### Summary of changes
**Asana Ticket:** [🍎🐛 API bug with pagination and offsets](https://app.asana.com/0/584764604969369/1202640206231353/f)

remove page limit max to avoid undocumented behavior of limiting to 100 items.


losvedir (approved):
Looks good! The asana task also mentions updating the person who raised the issue. Once this is merged and deployed can you take care of that as well?

PR #528, opened 2022-09-08
catch empty array state in canary terminate callback

bfauble:
#### Summary of changes
**Asana Ticket:** [🍎 Round two of API 502 investigation](https://app.asana.com/0/584764604969369/1202645043407373/f)

- Guards against non-function params before trying to call the function param in terminate callback.
- If no function is passed, call default notify function. Alternatively, could just return `:ok` here

### DIFF
--- a/apps/api_web/lib/api_web/event_stream/canary.ex
+++ b/apps/api_web/lib/api_web/event_stream/canary.ex
@@ -21,7 +21,7 @@ defmodule ApiWeb.EventStream.Canary do
   end
 
   @impl true
-  def terminate(:shutdown, notify_fn), do: notify_fn.()
+  def terminate(:shutdown, notify_fn) when is_function(notify_fn), do: notify_fn.()
   def terminate({:shutdown, _reason}, notify_fn), do: notify_fn.()
   def terminate(_, _), do: :ok
 end
--- a/apps/api_web/lib/api_web/event_stream/canary.ex
+++ b/apps/api_web/lib/api_web/event_stream/canary.ex
@@ -22,6 +22,7 @@ defmodule ApiWeb.EventStream.Canary do
 
   @impl true
   def terminate(:shutdown, notify_fn) when is_function(notify_fn), do: notify_fn.()
+  def terminate(:shutdown, _), do: @default_notify_fn.()
   def terminate({:shutdown, _reason}, notify_fn), do: notify_fn.()
   def terminate(_, _), do: :ok
 end


apps/api_web/lib/api_web/event_stream/canary.ex
```diff
@@ -21,7 +21,8 @@ defmodule ApiWeb.EventStream.Canary do
   end
 
   @impl true
-  def terminate(:shutdown, notify_fn), do: notify_fn.()
+  def terminate(:shutdown, notify_fn) when is_function(notify_fn), do: notify_fn.()
+  def terminate(:shutdown, _), do: @default_notify_fn.()
```
paulswartz, line 25: I assume this is happening because in practice, the GenServer is started like:
```elixir
Canary.start_link([])
```

Rather than specifying the `default_notify_fn` twice, what about changing the `start_link` to (correctly) treat the arguments as a Keyword list?
```elixir
def start_link(args) do
  notify_fn = case Keyword.fetch(args, :notify_fn) do
    {:ok, fun} when is_function(fun, 0) -> fun
    {:ok, not_fun} -> raise ArgumentError, "expect function/0 for notify_fn, got #{inspect(not_fun)}"
    _ -> @default_notify_fn
  end
  
  GenServer.start_link(__MODULE__, notify_fn, [])
end
```

paulswartz (approved):
🍰

apps/api_web/lib/api_web/event_stream/canary.ex
bklebe: Do we want to throw an exception in `start_link` or should we return `{:stop, reason}` from `init` instead? https://hexdocs.pm/elixir/1.14.0/GenServer.html#c:init/1 According to the docs:
> Returning {:stop, reason} will cause [start_link/3](https://hexdocs.pm/elixir/1.14.0/GenServer.html#start_link/3) to return {:error, reason} and the process to exit with reason reason without entering the loop or calling [terminate/2](https://hexdocs.pm/elixir/1.14.0/GenServer.html#c:terminate/2).

apps/api_web/lib/api_web/event_stream/canary.ex
```diff
@@ -9,19 +9,33 @@ defmodule ApiWeb.EventStream.Canary do
...
+
   def init(notify_fn) do
     Process.flag(:trap_exit, true)
     {:ok, notify_fn}
   end
 
   @impl true
-  def terminate(:shutdown, notify_fn), do: notify_fn.()
+  def terminate(:shutdown, notify_fn) when is_function(notify_fn), do: notify_fn.()
   def terminate({:shutdown, _reason}, notify_fn), do: notify_fn.()
```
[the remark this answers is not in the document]
bklebe: If line 30 is never called, couldn't we delete it? It feels wrong to leave in a line that represents a logic error, since we know that `notify_fn` is not a function..

bklebe (approved):

PR #532, opened 2022-09-13
chore: bump to latest Erlang 24

bklebe:
Discovered while attempting to build gtfs_creator, which requires the API to run tests. Erlang 24.3.2 appears not to build on Apple Silicon (or macOS 12.6) because of problems with wxWidgets. It's possible to disable wxWidgets in kerl and build it anyway, but I figure we might as well bump the version since it's essentially broken and 24.3.4.4 appears to work.


.tool-versions
```diff
@@ -1,4 +1,4 @@
-elixir 1.13.3-otp-24
-erlang 24.3.2
+elixir 1.13.4-otp-24
+erlang 24.3.4.4
```
bfauble, line 2: it looks like there's actually 24.3.4.5 out as of today it looks like not sure if you want to go ahead and bump to that or not.

bklebe: Sure thing.

PR #535, opened 2022-09-29
chore: Show Lynn on CR stops-on-route with shuttles only

jfabi:
#### Summary of changes

**Asana Ticket:** [[Extra] 🚧 Lynn–Swampscott shuttles from Oct 1](https://app.asana.com/0/584764604969369/1203051194767246/f)

Per the image below, deploying the Lynn shuttles in https://github.com/mbta/gtfs_creator/pull/1674 would lead to the bus times being shown for Swampscott, but not for Lynn. This is because Lynn would not be served by the train route, so the API's logic omits that stop from being shown in the Newburyport/Rockport Line timetable page.

(before)
Screen Shot 2022-09-29 at 17 37 46

To remedy this, we add an exception for the API's stops-on-route logic to additionally consider stations that are solely found in the Lynn–Swampscott route.


clem-hertling (approved):
fair enough 👍

PR #542, opened 2022-10-19
fix: shutdown the ALBMonitor normally

paulswartz:
Anything other than a `:normal` shutdown is reported as a crash/exit:

```
2022-10-19 16:12:23.577 [error] GenServer #PID<0.1181.0> terminating
** (stop) nil
Last message: :check
```

I'm not sure this is related to the 502 errors, but I did notice it during troubleshooting.

### DIFF
--- a/apps/alb_monitor/lib/alb_monitor/monitor.ex
+++ b/apps/alb_monitor/lib/alb_monitor/monitor.ex
@@ -41,7 +41,8 @@ defmodule ALBMonitor.Monitor do
   def child_spec(_opts) do
     %{
       id: __MODULE__,
-      start: {__MODULE__, :start_link, []}
+      start: {__MODULE__, :start_link, []},
+      restart: :transient
     }
   end
 


apps/alb_monitor/test/alb_monitor/monitor_test.exs
```diff
@@ -11,9 +11,11 @@ defmodule ALBMonitor.MonitorTest do
     mock_instance_ip("10.0.0.2")
     mock_health_response([{"10.0.0.1", "healthy"}, {"10.0.0.2", "draining"}])
 
-    start!()
+    monitor_pid = start!()
 
     assert_receive :shutdown
+
+    refute Process.alive?(monitor_pid)
```
bklebe, line 18: question: wouldn't this pass even if the process has crashed instead of exiting normally?

apps/alb_monitor/lib/alb_monitor/monitor.ex
```diff
@@ -69,7 +69,7 @@ defmodule ALBMonitor.Monitor do
       "draining" ->
         log_info("shutdown")
         shutdown.()
-        {:stop, nil, state}
+        {:stop, :normal, state}
```
bklebe, line 72: question: any idea why returning `nil` is considered an error here? [The docs for `handle_info`](https://hexdocs.pm/elixir/GenServer.html#c:handle_info/2) say `reason :: any()` which would suggest it's legitimate. If it's not and the callback type signature is too permissive, should we consider writing a custom typespec for this function so Dialyzer can ensure we don't return anything but `:normal`?

apps/alb_monitor/test/alb_monitor/monitor_test.exs
paulswartz: I was surprised, but in fact the process was still alive when I tested this locally.

apps/alb_monitor/lib/alb_monitor/monitor.ex
paulswartz: https://hexdocs.pm/elixir/Supervisor.html#module-exit-reasons-and-restarts has some more information:

>     `:normal` - in such cases, the exit won't be logged, there is no restart in transient mode, and linked processes do not exit
>    any other term - in such cases, the exit will be logged, there are restarts in transient mode, and linked processes exit with the same reason unless they're trapping exits

paulswartz:
I'm actually going to deploy this to dev-green and see if this helps at all as well.

apps/alb_monitor/lib/alb_monitor/monitor.ex
bklebe: Interesting, but this still doesn't say that returning `nil` causes an error. Seems a bit surprising to me.

paulswartz: "nil" is in the "any other term" case.

bklebe: ah, so that explains why the process wasn't dead then! it was probably restarted.

paulswartz:
I poked at this a little more yesterday, while not showing the `exit: :nil` is beneficial, it's not actually related to the issue.

paulswartz:
Closing this out in favor of a new PR which I believe does address the 502 issues.

PR #544, opened 2022-10-21
gracefully shut down the API during deploys (no 502s)

paulswartz:
#### Summary of changes

**Asana Ticket:** [Round three of 502 investigation](https://app.asana.com/0/881264583703207/1203193998678560)

This PR makes three main changes:
- have the ALBMonitor.Monitor process stop normally when it detects that the draining process has started
- add a Plug (`CheckForShutdown`) to notify connections (both long-running and short-term) that they should disconnect
- update `Canary` to monitor for both shutdown and draining, and trigger the disconnection process in both cases

In practice, what this means is that when draining starts, all existing connections start to be disconnected in order to be routed to other nodes. New connections can still be accepted, but they aren't kept open. Once the deregistration delay expires in the ALB Target Group, the node will be removed from the ALB and terminated (with SIGTERM).

### DIFF
--- a/apps/alb_monitor/test/alb_monitor/monitor_test.exs
+++ b/apps/alb_monitor/test/alb_monitor/monitor_test.exs
@@ -11,9 +11,11 @@ defmodule ALBMonitor.MonitorTest do
     mock_instance_ip("10.0.0.2")
     mock_health_response([{"10.0.0.1", "healthy"}, {"10.0.0.2", "draining"}])
 
-    start!()
+    monitor_pid = start!()
 
     assert_receive :shutdown
+
+    refute Process.alive?(monitor_pid)
   end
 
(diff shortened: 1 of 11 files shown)


apps/alb_monitor/test/alb_monitor/monitor_test.exs
```diff
@@ -11,18 +11,18 @@ defmodule ALBMonitor.MonitorTest do
     mock_instance_ip("10.0.0.2")
     mock_health_response([{"10.0.0.1", "healthy"}, {"10.0.0.2", "draining"}])
 
-    start!()
+    ref = start!()
```
bklebe, line 14: question: what does this `start!()` function do? I can't find where it's defined.

bklebe (approved):

PR #555, opened 2022-12-19
perf: small tweak to improve `/schedules/?include=predictions` performance

paulswartz:
Asana Ticket: https://app.asana.com/0/881264583703207/1203656158386261/f

I did some quick investigation since most of the 10 slowest queries were of this form.

### DIFF
--- a/apps/state/lib/state/prediction.ex
+++ b/apps/state/lib/state/prediction.ex
@@ -6,27 +6,8 @@ defmodule State.Prediction do
     recordable: Model.Prediction,
     hibernate: false
 
-  @doc """
-  Selects a distinct group of Prediction state sources, with filtering.
-
-  ## Examples
-
-    iex> [
-      State.Prediction
-    ]
-    |> State.Prediction.select_grouped(matchers, index, opts)
-    [...]
-  """
-
   import Parse.Time, only: [service_date: 1]
   import State.Route, only: [by_types: 1]
-
-  def select_grouped(sources, matchers, index, opts \\ []) do
-    sources
-    |> Stream.flat_map(& &1.select(matchers, index))
-    |> Enum.uniq_by(&prediction_key/1)
-    |> State.all(opts)
-  end
 
   def filter_by_route_type(predictions, nil), do: predictions
   def filter_by_route_type(predictions, []), do: predictions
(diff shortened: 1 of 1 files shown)


bklebe (commented):
Any chance I could get you to write up a little bit about how you did this investigation? Either in a ticket or somewhere else? Also, do you have any benchmarks or performance measurements for this change?

I'm imagining you looked at the slowest queries, and then traced back to this function, and noticed that most cases only had one source. Is that accurate?

paulswartz:
A writeup is a great idea! You have the basic idea right. I get a daily e-mail with the slowest queries, and at the moment they're mostly of the form `schedules/?filter[...]=...&include=predictions`, so I added some `.timer.tc/1` calls to various points to see what was taking the most time.

Test: `ab -l -c 2 -n 50 'http://localhost:4000/schedules/?filter[route]=Red&include=stop,trip,prediction&sort=arrival_time'`

Before:
```
Percentage of the requests served within a certain time (ms)
  50%   1303
  66%   1380
  75%   1498
  80%   1919
  90%   2566
  95%   2572
  98%   3039
  99%   3039
 100%   3039 (longest request)
```

After:
```
Percentage of the requests served within a certain time (ms)
  50%    651
  66%    662
  75%    671
  80%    675
  90%    684
  95%    692
  98%    826
  99%    826
 100%    826 (longest request)
```

I doubt that the actual speedup will be this dramatic in practice, though.

paulswartz:
@bklebe on a second look, there isn't anywhere that's using `select_grouped` anymore, so I removed it entirely.

bklebe (approved):

bklebe:
Any ideas what's up with CI?

paulswartz:
> Any ideas what's up with CI?

It's an issue with Poetry; https://github.com/mbta/api/pull/557 includes an upgrade, but I wasn't sure what Poetry was being used for.

bklebe:
> > Any ideas what's up with CI?
> 
> It's an issue with Poetry; #557 includes an upgrade, but I wasn't sure what Poetry was being used for.

I believe it's used for load testing (see https://github.com/mbta/api/blob/master/load_tests/pyproject.toml)

PR #557, opened 2023-01-04
feat: harden Docker image, audit packages

paulswartz:
Asana: [run API in a read-only container](https://app.asana.com/0/116562505129567/1203637202523166)

- ~update to Elixir 1.14 / Erlang 24.3.4.7 (duplicated by #569)~
- update to Phoenix 1.16.15
- run application in Docker image as a non-privileged user
- support running in a read-only filesystem (by updating Tzdata at build time)
- audit packages for security vulnerabilities in CI
- add `Content-Security-Policy` header as recommended by Sobelow

## pre-merge checklist
- [ ] remove `wip` commit


.github/workflows/deploy-dev-green.yml
```diff
@@ -1,16 +1,16 @@
...
     name: Deploy
     runs-on: ubuntu-latest
-    environment: dev-green
-    concurrency: dev-green
+    environment: read-only
+    concurrency: read-only
     env:
       ECS_CLUSTER: api
-      ECS_SERVICE: api-dev-green
+      ECS_SERVICE: api-read-only
```
bklebe, line 13: question: is this a permanent name change for this environment? We have an RFC that specifies how environments are supposed to be named now, and I'd like to stick to it if possible: https://github.com/mbta/technology-docs/blob/main/rfcs/accepted/0015-consistent-environment-naming.md. Also, the file is still called `deploy-dev-green.yml`.

mix.exs
```diff
@@ -62,6 +62,7 @@ defmodule ApiUmbrella.Mixfile do
     [
       {:credo, ">= 0.0.0", only: [:dev, :test], runtime: false},
       {:lcov_ex, "~> 0.2", only: [:dev, :test], runtime: false},
+      {:mix_audit, "~> 2.0", only: [:dev, :test], runtime: false},
```
bklebe, line 65: Is this needed in `MIX_ENV=test`?

config/runtime.exs
```diff
@@ -0,0 +1,44 @@
...
+
+  config :ex_aws,
+    dynamodb: [
+      port: "DYNAMO_PORT" |> System.get_env("8080") |> String.to_integer(),
+      scheme: System.get_env("DYNAMO_SCHEME"),
+      host: System.get_env("DYNAMO_HOST")
+    ]
+
+  config :alb_monitor,
+    ecs_metadata_uri: System.get_env("ECS_CONTAINER_METADATA_URI"),
```
bklebe, line 17: Are all these `get_env`s expected to be nil sometimes vs the `fetch_env!`s which aren't? Do they have defaults set somewhere else?

apps/api_web/lib/api_web/router.ex
```diff
@@ -23,14 +23,24 @@ defmodule ApiWeb.Router do
...
+    @content_security_policy Enum.join(
+                               [
+                                 "default-src 'none'",
+                                 "img-src 'self' cdn.mbta.com",
+                                 "style-src 'self' 'unsafe-inline' maxcdn.bootstrapcdn.com fonts.googleapis.com",
+                                 "script-src 'self' maxcdn.bootstrapcdn.com code.jquery.com",
+                                 "font-src fonts.gstatic.com maxcdn.bootstrapcdn.com"
+                               ],
+                               "; "
+                             )
```
bklebe, line 38: Do we have a list of these other origins that need to be allowed or is this off the top of your head/from digging through the source?

apps/api_web/lib/api_web/plugs/validate_date.ex
```diff
@@ -44,15 +44,17 @@ defmodule ApiWeb.Plugs.ValidateDate do
   defp error do
     %{
       code: :no_service,
-      description: "The current rating does not describe service on that date.",
+      detail: "The current rating does not describe service on that date.",
```
bklebe, line 47: Was this flagged by Sobelow? Looks like this is the proper format for ErrorSerializer, just curious how you found it.

.github/workflows/deploy-dev-green.yml
paulswartz: No, I'll remove this change before merging. The step needed to be in a file which exists on the default branch, so I couldn't create a new deploy file only for this PR. I'll also remove all the infrastructure for this `api-read-only` test environment.

mix.exs
paulswartz: Good point, it shouldn't be needed.

config/runtime.exs
paulswartz: This is probably left over from when we were migrating from Beanstalk -> ECS. I think they can be all switched to `fetch_env!`.

apps/api_web/lib/api_web/router.ex
paulswartz: A combination of digging through the source and trying it out in my browser to see what was rejected by Firefox.

config/runtime.exs
bklebe: Fair enough, sorry I didn't noticed this is more of a moved file than a new change.

bklebe (approved):

PR #560, opened 2023-01-06
test: split ASDF install from rest of build

paulswartz:
#### Summary of changes

**Asana Ticket:** [experiment: run API in a read-only filesystem](https://app.asana.com/0/116562505129567/1203637202523166)

Splits CI into stages, allowing resilience to failures and improved performance.
- ASDF
- Fetching Elixir dependencies
- Tasks which run in the Test Mix environment (Unit Tests, Integration Tests)
- Dialyzer
- Tasks which run in the Dev Mix environment (everything else)

Extracted from #557.

I looks like it's still running because there's no longer a "Build and test" action, which is what the merge check is waiting for.


.github/workflows/elixir.yml
```diff
@@ -12,60 +12,159 @@ concurrency:
...
+        uses: actions/cache@v3
+        with:
+          path: deps
+          key: ${{ runner.os }}-mix-v2-${{ hashFiles('**/mix.lock') }}
+          restore-keys: ${{ runner.os }}-mix-v2-
+      - name: Restore dev build cache
+        uses: actions/cache@v3
+        with:
+          path: _build/dev
+          key: ${{ runner.os }}-mix-dev-${{ hashFiles('**/mix.lock') }}
```
bklebe, line 102: Seems like we'll end up restoring old results in this directory. Is the idea that reusing those is still probably faster than a full rebuild?

paulswartz: Yeah; like with the build directory on your local machine, even if you update some dependencies, most of the _build/dev directory is still used for the unmodified deps.

bklebe: But doesn't `_build/dev` also contain built artifacts of the application when you run `mix run` or `mix compile`?

paulswartz: It does, yes. But as with your local machine, Elixir will recompile files that have been modified since the cache was updated.

bklebe (approved):

PR #574, opened 2023-02-02
feat: add recaptcha functionality to API registration

nlwstein:
#### Summary of changes

**Asana Ticket:** [🍎 Add recaptcha to signing up for API](https://app.asana.com/0/295455480314405/1201862282943725/f)

This PR adds recaptcha functionality to the API, and implements it on the non-admin user create form. 

A specific commit is pinned for the included library because the older published version causes dependency resolution issues. 

NOTE: **Before releasing this feature, we must** create recaptcha keys on a production/shared Google account, and add the following environment variables to the various deployments: `RECAPTCHA_PUBLIC_KEY`, `RECAPTCHA_PRIVATE_KEY`.


nlwstein:
@bklebe @paulswartz Do either of you have an opinion on how to handle the updated library assuming the external PR does not get merged? I am leaning towards just copying the code, with documentation in the README or something like that.

apps/api_web/lib/api_web/templates/client_portal/user/_new.html.heex
```diff
@@ -29,7 +29,10 @@
       </.form_group>
 
       <div class="form-group">
-        <%= submit "Register", class: "btn btn-primary" %>
+        <%= raw Recaptcha.Template.display %>
```
paulswartz, line 32: question: I thought the formatting these days used `()` for all function calls?
```suggestion
        <%= raw Recaptcha.Template.display() %>
```

apps/api_web/lib/api_web/controllers/portal/user_controller.ex
```diff
@@ -20,18 +20,26 @@ defmodule ApiWeb.ClientPortal.UserController do
...
-    case ApiAccounts.register_user(user_params) do
-      {:ok, user} ->
-        conn
-        |> put_session(:user_id, user.id)
-        |> configure_session(renew: true)
-        |> redirect(to: portal_path(conn, :index))
-
-      {:error, %ApiAccounts.Changeset{} = changeset} ->
+  def create(conn, %{"user" => user_params, "g-recaptcha-response" => recaptcha}) do
+    case Recaptcha.verify(recaptcha) do
```
paulswartz, line 24: suggestion: this is probably clearer (and definitely less nested) as a `with`:
```elixir
with {:ok, _} <- Recaptcha.verify(recapcha),
  {:ok, user} <- ApiAccounts.register_user(user_params) do
    # good path
else
  {:error, %ApiAccounts.ChangeSet{} = changeset} ->
    # changeset error
  {:error, error} ->
    # log error
    # redirect user
end
```

apps/api_web/lib/api_web/router.ex
```diff
@@ -28,11 +28,12 @@ defmodule ApiWeb.Router do
 
     @content_security_policy Enum.join(
                                [
-                                 "default-src 'none'",
+                                 "default-src 'self'",
```
paulswartz, line 31: question: I had left this out because not all of the options needed `'self'` (namely fonts). Is there a reason to add it back?

paulswartz (commented):
If it's not merged upstream, we can fork it into the `mbta` organization. We've done that for other dependencies in the past, and it'll probably help the MBTA.com team in the future as well.

apps/api_web/lib/api_web/controllers/portal/user_controller.ex
nlwstein: @paulswartz All about it, probably just my Python-brain 😂

nlwstein:
> If it's not merged upstream, we can fork it into the `mbta` organization. We've done that for other dependencies in the past, and it'll probably help the MBTA.com team in the future as well.

I'll give them a few days to respond, and will proceed with that!

apps/api_web/lib/api_web/router.ex
nlwstein: > question: I had left this out because not all of the options needed `'self'` (namely fonts). Is there a reason to add it back?

Reverted this, it does still work without it. I had changed this prior to the other policies 🙂

apps/api_web/lib/api_web/controllers/portal/user_controller.ex
nlwstein: Updated - https://github.com/mbta/api/commit/5f443347a834bda7c27c87f8afb4673578dcdf0c

apps/api_web/config/config.exs
```diff
@@ -62,6 +62,11 @@ config :mime, :types, %{
 
 config :ja_serializer, key_format: :underscored
 
+config :recaptcha,
+  verify_url: "https://www.google.com/recaptcha/api/siteverify",
+  public_key: {:system, "RECAPTCHA_PUBLIC_KEY"},
+  secret: {:system, "RECAPTCHA_PRIVATE_KEY"}
```
bklebe, line 68: suggestion: for dynamic configuration, we typically load it using [`System.get_env/2` ](https://hexdocs.pm/elixir/1.14.3/System.html#get_env/2)in `config/runtime.exs` now.

nlwstein: > suggestion: for dynamic configuration, we typically load it using [`System.get_env/2` ](https://hexdocs.pm/elixir/1.14.3/System.html#get_env/2)in `config/runtime.exs` now.

I added the env vars to that file, and added a feature flag to disable the recaptcha code outside of prod builds.

nlwstein:
This is available for review on `dev-green`, at least for now.

paulswartz:
I tried it out, and the first thing I noticed was that the form fields were cleared if I submitted without the captcha. Otherwise it seems to work well.

apps/api_web/config/config.exs
nlwstein: > I tried it out, and the first thing I noticed was that the form fields were cleared if I submitted without the captcha. Otherwise it seems to work well.

I added some frontend code to only enable the submit button once the user completes the recaptcha, so this won't happen outside of someone attempting to do something weird. 

Deployed to dev-green 🙂

nlwstein: > I tried it out, and the first thing I noticed was that the form fields were cleared if I submitted without the captcha. Otherwise it seems to work well.

I added some frontend code to only enable the submit button once the user completes the recaptcha, so this won't happen outside of someone attempting to do something weird. 

Deployed to dev-green 🙂

apps/api_web/lib/api_web/controllers/portal/user_controller.ex
```diff
@@ -20,18 +20,47 @@ defmodule ApiWeb.ClientPortal.UserController do
...
+
+  # Allow missing recaptcha response
   def create(conn, %{"user" => user_params}) do
-    case ApiAccounts.register_user(user_params) do
-      {:ok, user} ->
-        conn
-        |> put_session(:user_id, user.id)
-        |> configure_session(renew: true)
-        |> redirect(to: portal_path(conn, :index))
+    _create(conn, user_params, nil)
```
paulswartz, line 29: suggestion(non-blocking): You could combine these with `Map.get` which defaults to `nil`:

```elixir
def create(conn, %{"user" => user_params} = params) do
  recaptcha = Map.get(params, "g-recaptcha-response")
  ...
end
```

apps/api_web/lib/api_web/controllers/portal/user_controller.ex
```diff
@@ -20,18 +20,47 @@ defmodule ApiWeb.ClientPortal.UserController do
...
+  end
 
+  defp _create(conn, user_params, recaptcha) do
+    # Only test recaptcha if the feature is enabled (prod):
+    recaptcha_verified =
+      if Application.get_env(:recaptcha, :enabled) do
+        Recaptcha.verify(recaptcha)
+      else
+        {:ok, False}
```
paulswartz, line 38: suggestion: you probably meant `false` here.
```suggestion
        {:ok, false}
```
However, I'm a little confused about how `false` would be the correct return value here. What does `Recaptch.verify/1` normally return?

paulswartz (commented):
question: should there be some tests?

nlwstein:
> question: should there be some tests?

TL;DR Definitely! I added one that proves that the widget shows up when [supplying the test creds](https://stackoverflow.com/a/52189530): https://github.com/mbta/api/pull/574/files#diff-740356136f3337b207b35a2c46f8b47dcab0944918c977f6e8e0d04bd9eab52dR11. 

Due to the nature of ReCaptcha, a true end to end test isn't feasible, but this should ensure the path it takes is laid out correctly if that makes sense 🙂

nlwstein:
@paulswartz I accepted your change so the conversation got swallowed, but with regard to the `false` - the library returns an object [like this by default](https://github.com/samueljseay/recaptcha/blob/master/lib/recaptcha.ex#L46). 

I went with `false` but perhaps `nil` would be a better choice: I don't want any code to confuse our dev mode path for a real ReCaptcha response. If that is too loosely typed, I could return their object with all the fields set to `nil` or something like that? 

But really, we should only ever evaluate whether it gives `:ok` or `:error` which can happen [here when there's an API issue, most likely bad keys](https://github.com/samueljseay/recaptcha/blob/master/lib/recaptcha.ex#L42) or [here when the challenge fails](https://github.com/samueljseay/recaptcha/blob/master/lib/recaptcha.ex#L50). The result in that other object would only be something we would want to look at as devs vs. presenting on the frontend I think.

paulswartz (approved):
🍰

PR #586, opened 2023-02-24
ci: fix silent test failures & resolve failing tests

nlwstein:
#### Summary of changes

Currently, the Action that runs `mix test` [does not trigger a failure](https://github.com/mbta/api/actions/runs/4236971650/jobs/7362404936#step:8:1452). 

This branch fixes that, and also fixes the unit test failures, which are totally my fault 😂 

A couple notes: 

1. The trim and lowercase func was reformulated and moved up the pipe chain because the email validation library rejects those. 
1. Pretty much every canned email address in unit tests has been switched to `@mbta.com` instead of `@test.com` because we have a valid MX record, and `test.com` does not. If we don't want these nonsense emails indexed anywhere, we could always use some other valid domain.


.github/workflows/elixir.yml
```diff
@@ -167,11 +167,10 @@ jobs:
...
           key: ${{ runner.os }}-mix-test-${{ hashFiles('**/mix.lock') }}
           restore-keys: ${{ runner.os }}-mix-test-
+      - name: Run lcov
+        run: mix lcov
       - name: Run tests
-        run: |
-          mix lcov
-          cd apps/state_mediator
-          mix test --only integration
+        run: mix test
```
paulswartz, line 173: question: did you mean to drop the integration tests?

paulswartz (commented):
Thank you for cleaning up the test cases!

However, I think the underlying issue is that `mix lcov` does not return an non-0 exit code when the tests fail. Can we try to fix that, rather than running the tests twice?

nlwstein:
> Thank you for cleaning up the test cases!
> 
> However, I think the underlying issue is that `mix lcov` does not return an non-0 exit code when the tests fail. Can we try to fix that, rather than running the tests twice?

I'm quite sure this is an Actions behavior when you provide multiple commands vs an issue with that Elixir library, but I am confirming now.

.github/workflows/elixir.yml
nlwstein: > question: did you mean to drop the integration tests?

This is all set!

paulswartz:
> @paulswartz Unfortunately this cannot be fixed with an option passed to the mix command [...]

@nlwstein but it can be fixed with a patch upstream. I submitted https://github.com/dariodf/lcov_ex/pull/17 this afternoon. It looks like it correctly fails on the default API branch as well: https://github.com/paulswartz/api/actions/runs/4285638316/jobs/7464301038

We can either wait for the change to be merged, or you could use my fork of `lcov_ex` temporarily.

nlwstein:
> > @paulswartz Unfortunately this cannot be fixed with an option passed to the mix command [...]
> 
> @nlwstein but it can be fixed with a patch upstream. I submitted [dariodf/lcov_ex#17](https://github.com/dariodf/lcov_ex/pull/17) this afternoon. It looks like it correctly fails on the default API branch as well: https://github.com/paulswartz/api/actions/runs/4285638316/jobs/7464301038
> 
> We can either wait for the change to be merged, or you could use my fork of `lcov_ex` temporarily.

Sounds good! If you're planning on following through with a fork or merge, I can remove the relevant changes in this branch constraining it to just the unit test fixes. I figure it's best to hold off until it's in the `mbta` org or the PR is accepted, so that we don't tie the build to your personal account 🙂

paulswartz:
It landed today as version 0.3.1. You'll also need to include the new `--exit` flag when calling `mix lcov`.

nlwstein:
> It landed today as version 0.3.1. You'll also need to include the new `--exit` flag when calling `mix lcov`.

Sounds good, updating the branch!

nlwstein:
> It landed today as version 0.3.1. You'll also need to include the new `--exit` flag when calling `mix lcov`.

@paulswartz The PR is updated with the manual test removed, the dependency updated, and the `--exit` flag added.

nlwstein:
Failure case proof - https://github.com/mbta/api/actions/runs/4295935380/jobs/7487035183

.github/workflows/elixir.yml
```diff
@@ -167,11 +167,11 @@ jobs:
...
           restore-keys: ${{ runner.os }}-mix-test-
-      - name: Run tests
-        run: |
-          mix lcov
-          cd apps/state_mediator
-          mix test --only integration
+      - name: Run lcov and tests
+        run: mix lcov --exit
+      - name: Run integration tests
+        working-directory: apps/state_mediator
```
paulswartz, line 173: clever!

apps/api_accounts/lib/api_accounts/changeset.ex
```diff
@@ -288,6 +288,26 @@ defmodule ApiAccounts.Changeset do
     end
   end
 
+  @doc """
+  Ensure email address is downcase and trimmed.
```
paulswartz, line 292: nitpick: typo
```suggestion
  Ensure email address is downcased and trimmed.
```

apps/api_accounts/lib/api_accounts/changeset.ex
```diff
@@ -288,6 +288,26 @@ defmodule ApiAccounts.Changeset do
...
+  @spec format_email(t, atom) :: t
+  def format_email(%Changeset{} = changeset, field) do
+    unformatted_email = Map.get(changeset.changes, field, "")
+
+    formatted_email =
+      unformatted_email
+      |> String.downcase()
+      |> String.trim()
+
+    put_change(changeset, :email, formatted_email)
```
paulswartz, line 308: question: should this be `field` instead of `:email`?
```suggestion
    put_change(changeset, field, formatted_email)
```

apps/api_accounts/lib/api_accounts/changeset.ex
```diff
@@ -288,6 +288,26 @@ defmodule ApiAccounts.Changeset do
...
+  @doc """
+  Ensure email address is downcase and trimmed.
+
+  ## Examples
+
+      format_email(changeset, :email)
+
+  """
+  @spec format_email(t, atom) :: t
+  def format_email(%Changeset{} = changeset, field) do
```
paulswartz, line 300: question: the previous version only formatted the e-mail address if the changeset was otherwise valid. Do we want to keep that behavior?

nlwstein: > question: the previous version only formatted the e-mail address if the changeset was otherwise valid. Do we want to keep that behavior?

This is intentionally changed behavior. The email validator library rejects the input that we used to accept and clean up (`Nstein@domain.com ` => `nstein@domain.com`). This function now serves as a pass that takes place prior to validation to allow for those previously-acceptable inputs. 

We could also just not allow these inputs by removing the format email step entirely.

paulswartz (approved):
🍰

PR #594, opened 2023-03-01
fix: don't deploy on forks

paulswartz:
#### Summary of changes

**Asana Ticket:** [don't deploy automatically in forks](https://app.asana.com/0/584764604969369/1204089064359689/f)

Dev-blue automatically deploys, but this shouldn't happen in forks. This adds a check to the job to ensure that it only runs in the `mbta` owned repository.


nlwstein:
@paulswartz Do we want to also lock it down to just this repo as well? Org is definitely good enough, though!

paulswartz:
I suppose that if we have another fork of the API in our organization, then it's okay for it to deploy automatically 😄

PR #595, opened 2023-03-01
chore: add sentry library and configuration

nlwstein:
#### Summary of changes

This PR adds a sentry.io configuration for production-mode builds. As currently configured, all logs error and above will be published as events. 

I suggest that we configure error notifications to Slack / email for "error" and above in the Sentry dashboard. 

Proof of functionality in dev (local configuration was different to trigger this): 

1. Basic error in `api_web` (injection point): https://mbtace.sentry.io/issues/3967460762/?environment=dev&project=4504753674584064&query=is%3Aunresolved&referrer=issue-stream
2. Basic error in `api_accounts`: https://mbtace.sentry.io/issues/3967494870/?environment=dev&project=4504753674584064&query=is%3Aunresolved&referrer=issue-stream
3. Errors from background processes: https://mbtace.sentry.io/issues/3967464414/?environment=dev&project=4504753674584064&referrer=project-issue-stream
4. Errors from dev-green, deployed - https://mbtace.sentry.io/projects/api/?environment=dev-green&project=4504753674584064

SENTRY_ENV and SENTRY_DSN are available on all environments. 

**Asana Ticket:** [Set up Sentry](https://app.asana.com/0/584764604969369/1203870721173155/f)


paulswartz:
question: is there a reason to generically capture logs into Sentry? These logs already go into Splunk.

nlwstein:
> question: is there a reason to generically capture logs into Sentry? These logs already go into Splunk.
> 
> I looked at the errors on the first page of https://mbtace.sentry.io/issues/3967460762/?environment=dev&project=4504753674584064&query=is%3Aunresolved&referrer=issue-stream and they don't look like actual errors? One was a health check failure (expected) and the others were invalid API requests which correctly returned a 400.

1. It was asked for in this story - https://app.asana.com/0/584764604969369/1203870721173155/f My understanding is that it may help with debugging certain issues because it gives you greater insight into the call stack vs just the error log itself. 

2. That is likely caused by the logger being set to `warn` as opposed to `error`. I decided to set it to that to help with testing. We could change it to `error` to reduce the noise before releasing to production, or even make that an environment-dependent setting.

paulswartz:
Unless I'm missing something, all I see in the ticket is logging errors into Sentry, and what I'm seeing aren't errors. Maybe @bklebe can weigh in on whether this is what the team is expecting to see?

nlwstein:
> Unless I'm missing something, all I see in the ticket is logging errors into Sentry, and what I'm seeing aren't errors. Maybe @bklebe can weigh in on whether this is what the team is expecting to see?

This is because it was set to `:warn` instead of `:error`. I have updated the branch to `:error` since it sounds like that is preferable.

paulswartz:
Is there a way to filter out the 404s? https://mbtace.sentry.io/issues/?environment=dev-green&project=4504753674584064&referrer=sidebar

nlwstein:
> https://mbtace.sentry.io/issues/?environment=dev-green&project=4504753674584064&referrer=sidebar

My mild `opinion` here is that we should make 404s not actually trigger an `error` tier log at all instead. In my mind treating them as a `warn` alongside 400s makes sense. 

My rationale is that a thrown exception is also categorized as an `error` as opposed to `critical`: https://mbtace.sentry.io/issues/3967494870/?environment=dev&project=4504753674584064&referrer=project-issue-stream&statsPeriod=14d and I would rather not add more code that could potentially miss a real error. 

Do you agree? If not, I can try to specifically avoid posting 404 errors.

apps/api_web/lib/api_web.ex
```diff
@@ -12,6 +12,9 @@ defmodule ApiWeb do
   def start(_type, _args) do
     runtime_config!()
 
+    # Invoke Sentry logger:
+    _ = Logger.add_backend(Sentry.LoggerBackend)
```
paulswartz, line 16: question: should this be in `config/prod.exs`? That's how it's implemented in [RTR](https://github.com/mbta/rtr/blob/master/rtr/config/prod.exs#L10).

apps/api_web/lib/api_web/sentry_event_filter.ex
```diff
@@ -0,0 +1,6 @@
+defmodule ApiWeb.SentryEventFilter do
+  @behaviour Sentry.EventFilter
+
+  def exclude_exception?(%Phoenix.Router.NoRouteError{}, :plug), do: true
```
paulswartz, line 4: question: is this working? I still see a 404 logged after this deployed: https://mbtace.sentry.io/issues/?environment=dev-green&project=4504753674584064&referrer=sidebar&statsPeriod=1h

suggestion: You might also want to ignore 500s from the `/_health` endpoint.

apps/api_web/lib/api_web.ex
nlwstein: That should be fine. I was following their README - https://github.com/getsentry/sentry-elixir#capture-crashed-process-exceptions

apps/api_web/lib/api_web/sentry_event_filter.ex
nlwstein: Looking into this, it was working locally 🤔

nlwstein: (or at least, appeared to be)

nlwstein: 404: This was occurring in prod but not in dev (mode) because of `handle_sasl_reports: true` on the logger config. I added another rule to handle this case, and it seems to have stopped in `dev-green`. 

503: These errors were actually propagating from the `logster` library. Additionally, because they weren't exceptions Sentry can't filter them. I just disabled the error logging for now, because this will still be visible in Splunk and there's no clean way to filter it. If we ever want the error logging in the future, we can just mute this particular one in the UI.

apps/api_web/lib/api_web.ex
nlwstein: Updated to this approach

config/runtime.exs
```diff
@@ -6,6 +6,24 @@ is_release? = not is_nil(System.get_env("RELEASE_MODE"))
 if is_prod? and is_release? do
   config :tzdata, :autoupdate, :disabled
 
+  sentry_env = System.get_env("SENTRY_ENV")
+
+  if not is_nil(sentry_env) do
```
paulswartz, line 11: suggestion(non-blocking): 
```suggestion
  if is_binary(sentry_env) do
```

paulswartz (approved):

PR #581, opened 2023-02-21
chore: Suppress shuttle stops from Red Line stops list

jfabi:
#### Summary of changes

**Asana Ticket:** [🚧 Schedule for Alewife–Kendall/MIT shuttles](https://app.asana.com/0/584764604969369/1203887866352783/f)

Companion PR in support of https://github.com/mbta/gtfs_creator/pull/1805. What this change does is exclude certain shuttle bus stops from appearing on the Red Line's stops-on-route (ie, what gets output when you call [https://api-v3.mbta.com/stops?filter[route]=Red](https://api-v3.mbta.com/stops?filter%5Broute%5D=Red)).

PR #567, opened 2023-01-19
chore: Suppress North Station shuttle stop from Green Line E list

jfabi:
#### Summary of changes

**Asana Ticket:** [🚧 Extend OL weekend diversion to Ruggles](https://app.asana.com/0/584764604969369/1203657274721109/f)

Companion PR in support of https://github.com/mbta/gtfs_creator/pull/1776. What this change does is exclude the shuttle bus stop at North Station from appearing on the Green Line E's stops-on-route (ie, what gets output when you call https://api-v3.mbta.com/stops?filter[route]=Green-E). The other stops on the North Station–Government Center route, Haymarket and Government Center, are already hardcoded in this list from some previous shuttle.

PR #608, opened 2023-04-18
chore: Do not allow RL shuttle stops on rail route

jfabi:
#### Summary of changes

**Asana Ticket:** [🚧 JFK–Kendall/MIT shuttles, weekends Apr 22–30](https://app.asana.com/0/584764604969369/1204170395295667/f)

Resolves CI issue in GTFS-creator where Red Line shuttle bus stops, which are not in the parent stations, are being returned as part of the route's stops. This violates the API validation (that runs during GTFS-creator CI) that all the returned stops for subway routes must be part of a parent station.


jfabi:
Also, for reference, here are the last few times we've done this sort of change in the API
- https://github.com/mbta/api/pull/581
- https://github.com/mbta/api/pull/567
- https://github.com/mbta/api/pull/551

PR #624, opened 2023-05-31
fix: remove field that is not part of swagger json spec

nlwstein:


apps/api_web/lib/api_web/swagger_helpers.ex
```diff
@@ -182,7 +182,8 @@ defmodule ApiWeb.SwaggerHelpers do
         desc
       end
 
-    filter_param(path_object, :id, Keyword.merge(opts, desc: desc, name: :stop))
+    clean_opts = Keyword.delete(opts, :includes_children)
```
paulswartz, line 185: question: what keywords are allowed to be here? maybe it's cleared to do `Keyword.take(opts, [:name, :desc])` (or whatever other keywords are valid)?

bklebe: agreed, we should be parsimonious about what opts we pass through. fix this and I'll approve.

bklebe (changes requested):
please use `Keyword.take/2` per @paulswartz's comment.

bklebe (approved):

PR #632, opened 2023-06-23
chore: add connected stop override for ferries

nlwstein:
#### Summary of changes

### DIFF
--- a/apps/state/lib/state/connecting_stops.ex
+++ b/apps/state/lib/state/connecting_stops.ex
@@ -31,7 +31,9 @@ defmodule State.ConnectingStops do
     "place-DB-2205" => %{add: ~w(16391)},
     "place-FR-0064" => %{add: ~w(2137)},
     "place-GRB-0118" => %{add: ~w(3806)},
-    "place-harsq" => %{add: ~w(110)}
+    "place-harsq" => %{add: ~w(110)},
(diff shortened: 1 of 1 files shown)


apps/state/lib/state/connecting_stops.ex
```diff
@@ -31,7 +31,9 @@ defmodule State.ConnectingStops do
     "place-DB-2205" => %{add: ~w(16391)},
     "place-FR-0064" => %{add: ~w(2137)},
     "place-GRB-0118" => %{add: ~w(3806)},
-    "place-harsq" => %{add: ~w(110)}
+    "place-harsq" => %{add: ~w(110)},
+    "place-aqucl" => %{add: ~w(Boat-Long Boat-Long-4 Boat-Aquarium)},
```
jfabi, line 35: `Boat-Long-4` should be `Boat-Long-South-4`.

jfabi (changes requested):
One small adjustment and then we should be good!

apps/state/lib/state/connecting_stops.ex
nlwstein: Updated

jfabi (approved):
I deployed this to `dev-green` and checked the website and API (eg, https://api-dev-green.mbtace.com/stops/place-aqucl?include=connecting_stops). Looks good to me!

PR #633, opened 2023-06-27
explicitly include :all records for Boat-F6 in StopsOnRoute

bfauble:
#### Summary of changes

### DIFF
--- a/apps/state/lib/state/stops_on_route.ex
+++ b/apps/state/lib/state/stops_on_route.ex
@@ -186,15 +186,15 @@ defmodule State.StopsOnRoute do
     |> Enum.concat(explicit_override_records)
   end
 
-  defp gather_explicit_overrides(%{id: "Boat-F6"}, 0),
-    do: [
-      {"Boat-F6", 0, :all, "Boat-F6-Wdy-Smr-23", true,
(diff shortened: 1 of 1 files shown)


apps/state/lib/state/stops_on_route.ex
```diff
@@ -179,9 +179,27 @@ defmodule State.StopsOnRoute do
...
+    explicit_override_records = gather_explicit_overrides(route, direction_id)
+
+    shape_records
+    |> Enum.concat(other_records)
+    |> Enum.concat(explicit_override_records)
   end
 
+  defp gather_explicit_overrides(%{id: "Boat-F6"}, 1),
+    do: [
+      {"Boat-F6", 1, :all, "Boat-F6-Wdy-Smr-23", true,
```
jfabi, line 191: Why is the `service_id` included here? Is it required? Does this mean that override will fail if the `service_id` drifts from "Boat-F6-Wdy-Smr-23"? If so, that's a big liability that we will want to handle (likely as a follow-up task, though).

jfabi: So it sounds like if "Boat-F6-Wdy-Smr-23" suddenly (and automatically, as happens in GTFS-creator) was grouped with and consolidated into the `service_id` (eg) "Smmr-Wkd", the override would still proceed, right?

If so, given that this is already a hacky hack, let's leave it as is. 🙂

PR #636, opened 2023-07-05
ci: fix asana workflow for dependabot

paulswartz:
#### Summary of changes

### DIFF
--- a/.github/workflows/asana.yml
+++ b/.github/workflows/asana.yml
@@ -2,38 +2,16 @@ name: Asana integration for Github PRs
 on: 
   workflow_dispatch:
   pull_request:
-    types: [closed, edited, opened, reopened]
+    types: [review_requested, closed]
 
 jobs:
-  move-to-merged-asana-ticket-job:
-    runs-on: ubuntu-latest
-    if: github.event.pull_request.merged == true
-    steps:
-      - name: Github-Asana action from Insurify
-        uses: insurify/github-asana-action@v1.0.3
-        with:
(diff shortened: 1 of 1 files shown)


paulswartz:
@bklebe it looks like this still needs the `if` block for the repository. I suppose it could move into the shared action as well, but that would potentially make it less useful for non-MBTA repositories (if we care about that).

bklebe:
@paulswartz: the `if` block meaning your change about checking the repository owner, or something else?

paulswartz:
Yep, checking the repository owner.

bklebe:
> it looks like this still needs the if block for the repository.

Do you know why this is needed here but not RTR? Is it because you run a fork of this repo?


> I suppose it could move into the shared action as well, but that would potentially make it less useful for non-MBTA repositories (if we care about that).

I don't personally care about it very much. We could always add it as a parameter with a default value though. I'll defer to you what's best here; I think using the shared workflow is nice because it's one less difference to maintain between this app and others.

paulswartz:
> Do you know why this is needed here but not RTR? Is it because you run a fork of this repo?

@bklebe I think that's right. I'll look at the shared workflow to see if I can add the relevant check there.

bklebe (changes requested):
Looks like this now has a merge conflict. I approved the update to the shared workflow though, so hopefully now you can use that?

bklebe:
@meagharty is this still relevant?

meagharty:
Closing this since it was fixed in https://github.com/mbta/api/pull/702

PR #642, opened 2023-07-12
chore: exclude GL-B shuttle stops from stops-en-route

clem-hertling:
#### Summary of changes

**Asana Ticket:** [⚡️🚧 Green Line B surge from July 17](https://app.asana.com/0/584764604969369/1204867421964067/f)


clem-hertling:
oh this makes no sense, nvm

PR #653, opened 2023-07-27
manually override stop ordering for green line branch merging

bfauble:
#### Summary of changes

Current logic defaults to placing smaller branch after longer branch. In the case of Green-E/D the branch between Union Square/Medford and North station is considered a smaller branch than the remainder of the line and placed at the wrong end of the list. This is an explicit catch for this unique scenario and should not affect any others.

### DIFF
--- a/apps/state/lib/state/stops_on_route.ex
+++ b/apps/state/lib/state/stops_on_route.ex
@@ -348,6 +348,123 @@ defmodule State.StopsOnRoute do
 
   def merge_ids([], _), do: []
 
+  def merge_ids(
+        [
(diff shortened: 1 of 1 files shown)


jfabi:
@bfauble I just deployed this to `dev-blue`, and things seem to look good here 🎉—reconfigured the GTFS-creator branch and gave CI another run there. I'll wait to make sure it all passes (I anticipate it will), and then will give the ✅ approval here.

jfabi (approved):
🙏 🎉 Thank you for all the spelunking you had to do to put this together. Here's to rebuilding our stops-on-route logic very soon!

PR #657, opened 2023-08-07
fix: unable to limit fields on route_patterns

Whoops:
#### Summary of changes

**Asana Ticket:** [🍎🐛 Can't filter V3 API fields for /route_pattern/](https://app.asana.com/0/584764604969369/1205106233782383/f)

This adds the `fields` parameter to route_values, and updates how view modules are resolved to properly handle converting `snake_case` types with multiple words to `CamelCase`.

### DIFF
--- a/apps/api_web/lib/api_web/api_controller_helpers.ex
+++ b/apps/api_web/lib/api_web/api_controller_helpers.ex
@@ -197,7 +197,8 @@ defmodule ApiWeb.ApiControllerHelpers do
   end
 
   defp view_module_for_type(type) do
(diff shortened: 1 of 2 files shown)


paulswartz (commented):
question: could we add some tests to demonstrate what's being fixed (and to make sure we don't break it again in the future)?

Whoops:
I've added testing around the helper and controllers. The meat of the change is buried inside some private methods, so that's as close as I can get, but should be sufficient to prevent regression.

paulswartz (approved):

PR #663, opened 2023-08-22
fix: 500 when getting vehicle with OL crowding data

Whoops:
#### Summary of changes

**Asana Ticket:** [👥🔮 Surface car-level crowding data through RTR](https://app.asana.com/0/584764604969369/1204735391632694/f)

This fixes the 500 error when attempting to encode vehicles with carriages. It also adds carriages to the tests, and adds the remaining legal occupancy statuses.

### DIFF
--- a/apps/api_web/lib/api_web/views/vehicle_view.ex
+++ b/apps/api_web/lib/api_web/views/vehicle_view.ex
@@ -97,7 +97,6 @@ defmodule ApiWeb.VehicleView do
   defp encode_carriage(carriage) do
     %{
       label: carriage.label,
-      carriage_sequence: carriage.carriage_sequence,
       occupancy_status: occupancy_status(carriage, nil),
       occupancy_percentage: carriage.occupancy_percentage
     }
--- a/apps/api_web/lib/api_web/views/vehicle_view.ex
+++ b/apps/api_web/lib/api_web/views/vehicle_view.ex
@@ -69,13 +69,17 @@ defmodule ApiWeb.VehicleView do
       |> Atom.to_string()
       |> String.upcase()
 
-    def occupancy_status(%{occupancy_status: unquote(status)}, _conn) do
+    def occupancy_status(%{occupancy_status: unquote(status)}) do
       unquote(status_binary)
     end
   end
 
-  def occupancy_status(_, _) do
+  def occupancy_status(_) do
     nil
+  end
+
+  def occupancy_status(conveyance, _conn) do
+    occupancy_status(conveyance)
   end
 
   defp backwards_compatible_attributes(attributes, vehicle, "2017-11-28") do
@@ -86,12 +90,8 @@ defmodule ApiWeb.VehicleView do
     attributes
   end
 
-  defp encode_carriages(%{carriages: nil} = vehicle) do
-    vehicle
-  end
-
   defp encode_carriages(vehicle) do
-    Map.put(vehicle, :carriages, Enum.map(vehicle.carriages, &encode_carriage/1))
+    Map.put(vehicle, :carriages, Enum.map(vehicle.carriages || [], &encode_carriage/1))
   end
 
   defp encode_carriage(carraige) do
--- a/apps/api_web/lib/api_web/views/vehicle_view.ex
+++ b/apps/api_web/lib/api_web/views/vehicle_view.ex
@@ -63,7 +63,7 @@ defmodule ApiWeb.VehicleView do
   end
 
   for status <-
-        ~w(empty many_seats_available few_seats_available standing_room_only crushed_standing_room_only full not_accepting_passengers)a do
+        ~w(empty many_seats_available few_seats_available standing_room_only crushed_standing_room_only full not_accepting_passengers no_data_available not_boardable)a do
     status_binary =
       status
       |> Atom.to_string()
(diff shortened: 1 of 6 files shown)


apps/api_web/lib/api_web/views/vehicle_view.ex
```diff
@@ -84,4 +85,21 @@ defmodule ApiWeb.VehicleView do
...
+
+  defp encode_carriages(%{carriages: nil} = vehicle) do
+    vehicle
+  end
+
+  defp encode_carriages(vehicle) do
+    Map.put(vehicle, :carriages, Enum.map(vehicle.carriages, &encode_carriage/1))
+  end
+
+  defp encode_carriage(carraige) do
```
paulswartz, line 97: nitpick:
```suggestion
  defp encode_carriage(carriage) do
```

apps/api_web/lib/api_web/views/vehicle_view.ex
```diff
@@ -84,4 +85,21 @@ defmodule ApiWeb.VehicleView do
   defp backwards_compatible_attributes(attributes, _, _) do
     attributes
   end
+
+  defp encode_carriages(%{carriages: nil} = vehicle) do
+    vehicle
```
paulswartz, line 90: question: should this be an empty array rather than `nil`?

apps/api_web/lib/api_web/views/vehicle_view.ex
```diff
@@ -84,4 +85,21 @@ defmodule ApiWeb.VehicleView do
...
+  end
+
+  defp encode_carriages(vehicle) do
+    Map.put(vehicle, :carriages, Enum.map(vehicle.carriages, &encode_carriage/1))
+  end
+
+  defp encode_carriage(carraige) do
+    %{
+      label: carraige.label,
+      carriage_sequence: carraige.carriage_sequence,
```
paulswartz, line 100: question: 
- if the order is already present given that this is an ordered list, does the `carriage_sequence` need to be included? 
- and if it does need to be included, couldn't the key be `sequence` since the item is already known to be a carriage?

apps/api_web/lib/api_web/views/vehicle_view.ex
```diff
@@ -84,4 +85,21 @@ defmodule ApiWeb.VehicleView do
...
+      occupancy_status: occupancy_status(carraige, nil),
```
paulswartz, line 101: suggestion(non-blocking): rather than passing `nil` in for the conn, what about making `occupancy_status/1` be the result of the macro, and`occupancy_status/2` call `occupancy_status/1` without the conn?

```elixir
for status <- [...]
  def occupancy_status(unquote(status)), do: ...
end

def occupancy_status(vehicle, _conn), do: occupancy_status(vehicle.occupancy_status)

defp encode_carriage(carriage) do
  %{
    ...,
    occupancy_status: occupancy_status(carriage.occupancy_status)
  }
end
```

apps/api_web/lib/api_web/views/vehicle_view.ex
Whoops: My initial implementation actually was `[]`, but I changed to nil to keep consistent with the current behavior. It looks like the `nil` gets coerced to `[]` anyway, so I've gone back to that.

apps/api_web/lib/api_web/views/vehicle_view.ex
Whoops: Changed. It looks like what you describe is what JaSerializer actually does, defining `occupancy_status/1` and having `occupancy_status/2` call `/1`. I'm not sure why other views use the `/2` other than that's what the examples in the JaSerializer documentation do.

apps/api_web/lib/api_web/views/vehicle_view.ex
Whoops: Normally I wouldn't want to change the API contact in a bugfix, but I suppose this is a bit of a special case, since this information wasn't available before, and wouldn't have worked if it was, so there shouldn't be any downstream consumers depending on the shape.

For my 2c, I'd probably leave `carriage_sequence` out entirely, since it's implied by the array and the API breaks from the GTFS spec in other ways (such as putting these on the `carriage` field). That said these fields to map directly to the corresponding GTFS fields, and `carriage_sequence` is actually the only field required in the spec. For context it looks like this was originally implemented in [this ticket](https://app.asana.com/0/584764604969369/1204735400214299/f) that calls for those fields.

I'll defer to you on the final shape. If you want it dropped it's easy enough to do.

paulswartz: Assuming that there's some other part of the system (maybe the consist/carriage parsing?) that's ensuring the list is in `carriage_sequence` order, I'd prefer to drop it.

Whoops: The GTFS spec requires they be in order, and RTR is generating the sequence numbers based on their position in the array so we should be covered there. The field has been removed.

paulswartz (approved):
:cake: should we put this up on dev-green for more testing before merging?

Whoops:
> 🍰 should we put this up on dev-green for more testing before merging?

The initial version has been in dev-green for the last couple days. I've gone ahead and sent the new version. I don't think there's any particular rush for this to go out, just needs to happen before the RTR crowding changes land.

PR #667, opened 2023-08-28
feat: block access to admin pages when 2fa is disabled

nlwstein:
#### Summary of changes

**Asana Ticket:** [🍎 Force 2FA for API admins](https://app.asana.com/0/295455480314405/1205277400179415/f)

This PR also addresses [🍎 API "Login" and "Register" links should not appear if user is already logged in](https://app.asana.com/0/295455480314405/1205212005457786/f)

Adds a `Plug` that can restrict access to URLs until the user enables 2 factor authentication, and applies it to `/admin` routes. 

Updates footer behavior so that `USER_EMAIL` / `Log Out` is shown instead of `Login` / `Register` when the user is already in a logged-in session.


Whoops (approved):
👍

PR #671, opened 2023-09-14
fix: Use IDs for Stoughton routes on Providence/Stoughton line for Fall 2023 diversion

clem-hertling:
#### Summary of changes

**Asana Ticket:** [🚧 Splice together new Stoughton Branch shapes](https://app.asana.com/0/584764604969369/1205320592017665/f)

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -412,6 +412,26 @@ config :state, :stops_on_route,
       "place-DB-2222",
       "place-DB-2205"
     ],
+    {"CR-Providence", 0} => [
+      "place-DB-2265",
+      "place-DB-2258",
+      "place-DB-2249",
+      "place-DB-2240",
+      "place-DB-2230",
+      "place-DB-2222",
+      "place-DB-2205",
+      "place-DB-0095"
+    ],
+    {"CR-Providence", 1} => [
+      "place-DB-2265",
+      "place-DB-2258",
+      "place-DB-2249",
+      "place-DB-2240",
+      "place-DB-2230",
+      "place-DB-2222",
+      "place-DB-2205",
+      "place-DB-0095"
+    ],
     {"CR-Fairmount", 0} => [
       "place-FB-0166",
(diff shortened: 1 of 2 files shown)


jfabi (commented):
I'll be honest, I didn't quite understand whether or not it would be expected for this to be needed with the new, non-typical pattern, at first...but upon closer inspection, yes, this looks right to me.

A few additional things, however:
- In _apps/state/config/config.exs_, please add the Fairmount Line-specific stations to the `not_on_route` configuration for the Providence/Stoughton Line; you'll see that this has already been done for the Franklin/Foxboro Line. This is sort of an insurance that will make sure we don't accidentally associate those Fairmount Line stations with the Providence/Stoughton Line on MBTA.com pages...
- Over in https://github.com/mbta/gtfs_creator/pull/2125, I'm going to suggest that we add "via" messaging to the `route_pattern_name`s, and that **might** result in changes to the `name`s here... I would suggest making those changes first and then ensuring that the API validation still passes on that branch.
- AFAIK (unless you've heard otherwise), this change to add the new pattern is permanent, so you should remove the "temporarily" wording from the PR and commit titles.

clem-hertling:
mmh okay so,
- fair enough, I'll add this!
- did that, and it does change the name, which breaks the validation, so I'll fix that here now
- I thought the new pattern was only being used during work on the tracks? Or is that the one that's related to the Amtrak changes? either way, I can remove the temporarily, yeah

clem-hertling:
though after checking, you're right: with the new non-typical patterns, this doesn't look like it's needed... but are the fairmount patterns really non-typical?

clem-hertling:
ah, nevermind, I get it now. whoops!

jfabi (approved):

clem-hertling:
I can't merge this PR bc the test failed, but I think that's because it tests on gtfs_creator's `master` branch, so I can't really solve this without breaking gtfs_creator beforehand

PR #677, opened 2023-09-29
fix: increase max ports in health checker

Whoops:
#### Summary of changes

**Asana Ticket:** No ticket

Increase max ports to prevent the health checker from recycling pods before they are overloaded.


nlwstein (approved):
I would also be cool w/ 800 as you said in Slack 😄

nlwstein (approved):
🧑‍🚒

PR #678, opened 2023-10-02
feat: switch from Cowboy to Bandit, upgrade to Phoenix 1.7

paulswartz:
[Bandit](https://hexdocs.pm/bandit/Bandit.html) is a new web server, designed around supporting the Plug interface that Phoenix uses.

It also logs errors much more aggressively, which Phoenix 1.7 handles much better than Phoenix 1.6.

Best to review this commit-by-commit: the reformatting commit adds a bit of noise.

Formatting has moved into the "Dev" tasks, as it now depends on the compiled dependencies. I'll remove it from the requirements when this merges.


apps/api_web/config/config.exs
```diff
@@ -7,13 +7,18 @@ import Config
...
 config :api_web, ApiWeb.Endpoint,
+  adapter: Bandit.PhoenixAdapter,
   url: [host: "localhost"],
   root: Path.dirname(__DIR__),
   secret_key_base: "v1EHfW07QPr8ai7bi0hooadtBorROPNjhSWx7CGv7AiCOhEyGoeT1jagMTNCE3PU",
   render_errors: [accepts: ~w(json html)],
   http: [
-    compress: true,
-    protocol_options: [idle_timeout: 86_400_000, request_timeout: 120_000]
+    http_1_options: [
```
nlwstein, line 16: Was this intentionally only set for `http_1_options`? (https://hexdocs.pm/bandit/Bandit.html#t:http_2_options/0)

nlwstein (changes requested):
Do you think it would be worth it to (have you already?) load test this module vs. Cowboy - I'm not opposed to just throwing it into production and reverting if necessary but it might be interesting to see how much extra stability / performance we can get out of it in a synthetic test too. 

Specifically, # of concurrent connections per instance would be very interesting given the issue last week. 

My only q w/ what's here is the `http_1_options` - I assume that was added for a reason so I just wanted to make sure we wouldn't benefit from the compression on HTTP2 as well.

apps/api_web/config/config.exs
paulswartz: We don't receiving incoming HTTP2 connections from the ALB, so it wouldn't be used.

paulswartz:
@nlwstein I'm throwing it back up on dev-blue for further testing.

When I was testing it last night, I was able to get it up to around 1,000 `/vehicle` streams per server before I saw AWS scale up the number of instances.

paulswartz:
@nlwstein doing some further testing:
- I can have 2,000 open connections without appreciably increasing the memory or CPU usage on `dev-blue`
- `dev-green` gets more traffic than `dev` (not sure why) but it's using 40% of the CPU and a bit less memory

I'd be okay with doing a temporary deploy to `api-prod` to see what it looks like with even more traffic, but I'll leave that up to you and @bklebe.

bklebe:
I support trying a prod deploy.

nlwstein (approved):

paulswartz:
Running on `api-prod` for now. Things are looking okay but will continue to monitor during the afternoon/evening.

paulswartz:
CPU and Memory look good:
Screenshot 2023-10-03 at 1 38 49 PM
However, I'm seeing a lot of confusing errors:
```
Child :undefined of Supervisor #PID<0.4312.0> (Supervisor.Default) terminated
```
They don't appear to be causing problems, but I'd like to know why they're happening before we merge.

paulswartz:
The errors are better, but I still need to investigate the test failures.

paulswartz:
So I went down a bit of a rabbit hole on this one looking into the various errors. It turns out that this was a known issue with how Bandit logs errors coming from Phoenix (see https://github.com/phoenixframework/phoenix/pull/5445 for more), so Phoenix needed to be upgraded in order to address the errors. There were quite a few changes in Phoenix between 1.6 and 1.7, so the upgrade was a bit more involved than simply bumping the version in the `mix.exs` file.

All that being said: I'll run this on dev-green overnight to make sure there aren't any other unexpected issues, and then maybe we can try it on production again?

nlwstein (approved):

PR #685, opened 2023-10-12
fix: always show route_pattern_prefix_overrides trips

nlwstein:
#### Summary of changes

**Asana Ticket:** [[extra] 🐛🍎 API: Ashland stop disappeared on the Worcester Line](https://app.asana.com/0/584764604969369/1205604636963286/f)

**Asana Ticket:** [🍎 Update merge_ids logic for StopsOnRoute](https://app.asana.com/0/584764604969369/1204975607244422/f)

For this (`Boat-F6`), I am suggesting that we just whack the hardcoded configuration - from my understanding of the code, the output we are currently seeing is correct. For reference, locally I am seeing: 

NOT SPECIFIED: 

INBOUND:

OUTBOUND: 

According to the PDF schedule and my reading of the static schedule in GTFS, that is correct. unless 

Some extra context can be found [here](https://github.com/mbta/api/commit/ddf61fef3c606f5a8cf81013943e204b16c1734b), though honestly stepping through this code with a debugger gave me a better understanding than anything else 💫

### DIFF
--- a/apps/state/lib/state/helpers.ex
+++ b/apps/state/lib/state/helpers.ex
@@ -16,6 +16,9 @@ defmodule State.Helpers do
         Application.compile_env(:state, :stops_on_route)[:route_pattern_prefix_overrides] do
     def stops_on_route?(%Trip{route_pattern_id: unquote(route_pattern_id) <> _}),
       do: unquote(include?)
+
+    def stops_on_route_by_shape?(%Trip{route_pattern_id: unquote(route_pattern_id) <> _}),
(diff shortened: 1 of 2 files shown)


nlwstein:
@bfauble I updated this branch to not remove the override to avoid the unordered Logan. I suggest that we close the investigation ticket and either make a follow up ticket to support overriding global stop order (IE: when pulling stops without a direction_Id specified) or leave as-is and hopefully refactor the entire approach at some point.

PR #694, opened 2023-10-27
chore: bump erlang/elixir to sonoma friendly versions

Whoops:
#### Summary of changes

Updating Erlang to a version that works on Sonoma

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
-elixir 1.14.3-otp-25
-erlang 25.2.2
+elixir 1.14.5-otp-25
+erlang 25.3.2.7
 python 3.9.16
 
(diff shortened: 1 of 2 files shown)


Whoops:
Not bumping straight to Erlang 26 and Elixir 1.15 due to [issues with old dependencies](https://app.asana.com/0/1204137169527258/1205823045134240/f) we should sort out. Erlang 25.3.2.7 seems to resolve the Sonoma issues in the meantime.

nlwstein:
@Whoops The Dockerfile (and perhaps other locations related to CI) does not pull from the ASDF config: https://github.com/mbta/api/blob/master/Dockerfile

nlwstein (changes requested):
(Added comment prior to submitting review - mainly just double check that CI builds pull the same version 🙂 otherwise this looks good!)

Whoops:
Good eye! I knew we we using ASDF for CI, but didn't think to check the Dockerfile. I checked for other instances of hardcoded versions and didn't find any, so I think this covers it now.

nlwstein (approved):
🚀

PR #696, opened 2023-10-31
feat: concurrent connections limiter

nlwstein:
#### Summary of changes

**Asana Ticket:** [🍎 Reject API requests when there are too many concurrent requests](https://app.asana.com/0/295455480314405/1205450740821028/f)

This provides a new rate limiter that is mainly useful for clamping down on too many active streaming connections. It is also able to track and control long-running "static" requests (for example, trying to load all of the predictions at once). It is configurable on a per-user basis in the admin panel (`-1` to disable). The largest value between user config and the base config is used. 

The static concurrency locks are released in real-time, whereas the streaming ones depend on the hibernate loop cycle, so there can be a latency of under a minute for releasing those. The limits I have added to the config file are guidance for testing, not necessarily for production. 

`Opinion`: I am not sure if `memcached` is the ideal tool for this, but I used it because I don't want to introduce new infrastructure if possible and it was already in-use for a similar job. Changing out the storage mechanism in the future should be relatively easy. 

The whole rate limiting system seems like a good use case for Redis (or perhaps there's something new that's even better), primarily because: 

1. You can more efficiently filter / upsert nested data, or even just on a key prefix basis. 
2. It has a concept of locks and transactions


paulswartz:
Thanks for taking a look at this!

question: did you consider using the existing concurrency tracking in [`RequestTrack`](https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/request_track.ex)?

question: rather than nesting data in Memcache, did you consider using Memcache's [incr/decr](https://github.com/memcached/memcached/wiki/Commands#incrdecr) commands for managing the count? That's what the existing [rate limit implementation](https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/api_web/rate_limiter/memcache.ex#L21) does.

nlwstein:
> question: did you consider using the existing concurrency tracking in [`RequestTrack`](https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/request_track.ex)?

I tried this just now and it seems to only register an increment (no decrement) for streaming requests. Let me know if you think this is worth further poking 🙂 

> question: rather than nesting data in Memcache, did you consider using Memcache's [incr/decr](https://github.com/memcached/memcached/wiki/Commands#incrdecr) commands for managing the count? That's what the existing [rate limit implementation](https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/api_web/rate_limiter/memcache.ex#L21) does.

Perhaps it's overly complex but I wanted to be able to track on a per-process level vs. just a sum. Without being able to ask memcache for `select all values where keys match $user_$type_*` I'm not sure if that's viable w/ just a counter. The reason for this is if someone subscribes to a stream on a backend process, it's possible that they may keep them open for ~days at a time. One consequence of this is that we could never allow expiry of the count (which I don't think memcache supports). We could always live with the understanding that a few long-running connections would outlive the tracking, if you'd prefer it to work that way. 

Do you have an opinion on switching the rate limit state over to Redis or similar in a future task? I don't think the dev time would be that high and it would give us more flexibility for shared state between the ECS tasks (mainly atomic transactions, querying against key prefixes for this particular feature).

paulswartz:
> > question: did you consider using the existing concurrency tracking in [RequestTrack](https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/request_track.ex)?
> I tried this just now and it seems to only register an increment (no decrement) for streaming requests. Let me know if you think this is worth further poking 🙂

Decrementing for streaming processes is handled by the process monitors: https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/request_track.ex#L77-L89

> Perhaps it's overly complex but I wanted to be able to track on a per-process level vs. just a sum. 

Can you say more about why you wanted to do that? There's a separate process for each connection, so the per-process count is either 0 or 1.

apps/api_web/config/config.exs
```diff
@@ -30,6 +30,26 @@ config :api_web, :rate_limiter,
   max_registered_per_interval: 100_000,
   wait_time_ms: 0
 
+config :api_web, :rate_limiter_concurrent,
+  enabled: False,
+  memcache: False,
```
bklebe, line 35: issue (blocking): boolean literals are not capitalized (https://hexdocs.pm/elixir/1.16/basic-types.html#booleans-and-nil)

bklebe (changes requested):

nlwstein:
> > > question: did you consider using the existing concurrency tracking in [RequestTrack](https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/request_track.ex)?
> > > I tried this just now and it seems to only register an increment (no decrement) for streaming requests. Let me know if you think this is worth further poking 🙂
> 
> Decrementing for streaming processes is handled by the process monitors:
> 
> https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/request_track.ex#L77-L89
> 
> > Perhaps it's overly complex but I wanted to be able to track on a per-process level vs. just a sum.
> 
> Can you say more about why you wanted to do that? There's a separate process for each connection, so the per-process count is either 0 or 1.

Sure! Let's say a user has two active streaming connections, each one happening to be to a separate ECS task. These are long-running connections (spawned at application startup and in theory only recreated on crash or restart) - maybe for an application like the one that caused the issue that created this task, but implemented on the backend so that it didn't scale per client. When their app starts up, the counter would get bumped to 2 for. After n minutes / hours / etc. when the Memcache TTL hits, the entry would be cleared, and the connection(s) would still be open. 

I can think of a situation where someone has an app in a degraded state where it keeps opening connections after some error condition is reached without closing the old tasks. Eventually, it would be allowed to continue opening connections by the rate limiter. 

I could very well be missing something here, but that's the scenario I was considering. 

We could potentially add a loop somewhere to update a simpler memcache value that's just {counter, timestamp} whenever an outstanding connection is present, if you think that is more straightforward, but I'm not sure if you can do that with the increment / decrement commands.

nlwstein:
> > > question: did you consider using the existing concurrency tracking in [RequestTrack](https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/request_track.ex)?
> > > I tried this just now and it seems to only register an increment (no decrement) for streaming requests. Let me know if you think this is worth further poking 🙂
> 
> Decrementing for streaming processes is handled by the process monitors:
> 
> https://github.com/mbta/api/blob/bbfa321004a2729ce8fdfa921696f7c8be5b0e20/apps/api_web/lib/request_track.ex#L77-L89

OK got it, thanks! I should be able to update to use this for detecting the streaming disconnect at the very least, which was the one event not handled by the plug.

apps/api_web/config/config.exs
nlwstein: 🐍 on the 🧠...

nlwstein:
> TBH, I was expecting something a little simpler, more like:
> 
> ```elixir
> # in apps/api_web/lib/api_web/plugs/request_track.ex
>   def call(conn, table_name) do
>     key = conn.assigns.api_user
>     RequestTrack.increment(table_name, key)
>     count = RequestTrack.count(table_name, key))
>     _ = Logger.metadata(concurrent: count)
> 
>     conn = register_before_send(conn, fn conn ->
>       # don't decrement if we're using chunked requests (streaming)
>       if conn.state == :set do
>         RequestTrack.decrement(table_name)
>       end
> 
>       conn
>     end)
> 
>     if count > max_concurrent_for_user(key) do
>       conn
>       |> put_status(429)
>       |> put_view(ApiWeb.ErrorView)
>       |> render("429.json-api", [])
>       |> halt()
>     else
>       conn
>     end
>   end
> ```

This would only be per ECS task, though, right? So whenever that scales so would our rate limit. I'm cool with that if you are, but I assumed we would want to have the limit apply across all of the prod deployments.

paulswartz:
I'd prefer to start with something simpler and expand it if necessary than start with something complex, but I'll defer to you and @bklebe on the final approach.

bklebe:
@paulswartz I think we may want a meeting to discuss this — broadly I agree with you though.

@nlwstein I think perhaps with streaming connections that sharing it across the cluster is not quite as important, mostly because streaming connections impose more resource consumption on the client side compared to one-shot requests. 

The problem with not sharing the streaming connection limit across the cluster would be that clients could cause the cluster to scale up by maxing out streaming connections. @paulswartz do you think that's acceptable, at least to start with?

paulswartz:
I do think that's acceptable, but happy discuss it more!

nlwstein:
@paulswartz @bklebe As I mentioned on the call, I intend to move my hooks for detecting streaming open / close to the existing logger, and from there updating the logs to reference the global state.

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,175 @@
...
+    key = "concurrent_#{user.id}_#{type}"
+
+    {
+      type,
+      key
+    }
+  end
+
+  defp get_current_unix_ts do
+    DateTime.utc_now() |> DateTime.to_unix()
```
paulswartz, line 31:
```suggestion
    System.system_time(:second)
```

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,175 @@
...
+  defp get_heartbeat_tolerance do
+    Keyword.fetch!(Application.get_env(:api_web, :rate_limiter_concurrent), :heartbeat_tolerance)
+  end
+
+  def get_locks(%ApiWeb.User{} = user, event_stream?) do
+    if enabled?() do
+      current_timestamp = get_current_unix_ts()
+      heartbeat_tolerance = get_heartbeat_tolerance()
+      {_type, key} = lookup(user, event_stream?)
+      {:ok, locks} = GenServer.call(__MODULE__, {:memcache_get, key, %{}})
```
paulswartz, line 43: suggestion: I think that having a `GenServer.call/2` here is going to mean we won't be able to have concurrent API calls: they'll be serialized as messages to this single GenServer. Could we share the pooled Memcache connections we're already using for the regular rate limiter?

apps/api_web/config/config.exs
```diff
@@ -30,6 +30,26 @@ config :api_web, :rate_limiter,
...
+  connection_opts: [
+    namespace: "api_concurrent_rate_limit_dev",
+    # The total lifetime (s) of an API key + request type (static or streaming) key:
+    ttl: 60,
+    # Allows us to hold a nested payload per key, this is necessary because memcached does not offer
+    # a great way to group keys:
+    coder: Memcache.Coder.JSON
+  ],
+  # Default concurrent connections - these can be overridden on a per-key basis in the admin UI:
+  max_anon_streaming: 1,
```
paulswartz, line 48: suggestion: I forget if I added this as a comment before, but there are no anonymous streams. An API key is required to use the event-stream interface.

bklebe (changes requested):
This is out-of-date with master and needs to be updated.

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,175 @@
...
+    {:reply,
+     {:ok,
+      case Memcache.get(state.memcache_pid, key) do
+        {:ok, result} -> result
+        _ -> default_value
+      end}, state}
+  end
+
+  def handle_call({:memcache_set, key, value}, _from, state) do
+    {:reply, {:ok, Memcache.set(state.memcache_pid, key, value)}, state}
```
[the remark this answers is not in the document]
nlwstein: * Yes - I believe this could only be fixed by using a data store that understands transactions (such as redis). In practice, I suspect that the operation will be fast enough that we won't run into this, and I feel it's worth at least testing. 

* I pushed an update to always include an instance specific UUID as well to cover this, thanks!

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
nlwstein: Are you saying that calling `GenServer.call` makes the entire call chain synchronous, or just the writes to memcache? 

I'm not entirely sure we would want asynchronous writes to memcache in this particular case due to the fact that it doesn't support transactions - this might inadvertently prevent the race mentioned earlier at least on a per-instance basis 🤔 

I have some test code locally that we may be able to use to take a similar approach w/r/t memcache supervisor if this doesn't work as expected.

apps/api_web/config/config.exs
nlwstein: You're right, I don't think this code path is reachable. Should I remove it? It doesn't *harm* anything but I could see the value in having less code if we don't think we will ever allow anonymous streams.

bklebe (changes requested):
@nlwstein, thanks for the PR. Would you mind cleaning up the commit history or squashing all these WIPs down if you only intend to merge one?

nlwstein:
> @nlwstein, thanks for the PR. Would you mind cleaning up the commit history or squashing all these WIPs down if you only intend to merge one?

@bklebe All set! 

In the future, would you prefer squash prior to review? I had been following a pattern of squashing prior to merge. 

I have no strong opinion, though if I had to pick one, it can be helpful to step through intermediate commits to see someone's thought process during code review.

apps/api_web/lib/api_web/plugs/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  defp put_concurrent_rate_limit_headers(conn, limit, remaining) do
+    conn
+    |> put_resp_header("x-concurrent-ratelimit-limit", "#{limit}")
+    |> put_resp_header("x-concurrent-ratelimit-remaining", "#{remaining}")
+  end
+
+  def enabled? do
+    Keyword.fetch!(Application.get_env(:api_web, :rate_limiter_concurrent), :enabled) == true
```
paulswartz, line 50: question: since this isn't set at runtime, could it be a `compile_env` check to avoid the `Application` lookup at runtime?

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,175 @@
...
+    active_connections = user |> get_locks(event_stream?) |> Map.keys() |> length
+
+    limit =
+      case {event_stream?, user.type} do
+        {true, :registered} ->
+          if user.streaming_concurrent_limit >= 0,
+            do:
+              max(
+                user.streaming_concurrent_limit || 0,
+                Keyword.fetch!(
```
paulswartz, line 70: suggestion: can we do these application lookups at compile time instead, since they aren't changing at runtime?

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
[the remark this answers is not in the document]
nlwstein: The UUID fix has been pushed - somehow I broke a merge from the default branch, so that's why I took this back to draft but it should be correct now.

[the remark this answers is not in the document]
nlwstein: I have updated to use `cas` - the memcache library provides a clean function where you can supply an anonymous function that updates the value w/ automatic retry.

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
nlwstein: Thank you, that makes sense! I have updated to share the pool with `RateLimiter`.

apps/api_web/config/config.exs
nlwstein: All set.

nlwstein:
@paulswartz Apologies for the busted state with the prior review, I messed something up when I squashed my newest work on top of the old branch. I have also pushed new changes to address all of your comments 😄

apps/api_web/config/dev.exs
```diff
@@ -28,3 +28,6 @@ config :logger, :console, format: "[$level] $message\n", level: :debug
 # Do not configure such in production as keeping
 # and calculating stacktraces is usually expensive.
 config :phoenix, :stacktrace_depth, 20
+
+config :api_web, :rate_limiter_concurrent, enabled: true, memcache: true
+config :logger, level: :error
```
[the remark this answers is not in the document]
nlwstein: This was unintentional - removed.

apps/api_web/config/config.exs
```diff
@@ -28,7 +28,21 @@ config :api_web, :rate_limiter,
...
   max_registered_per_interval: 100_000,
-  wait_time_ms: 0
+  wait_time_ms: 0,
+  connection_opts: [
+    coder: Memcache.Coder.JSON
+  ]
+
+config :api_web, :rate_limiter_concurrent,
+  enabled: false,
+  memcache: false,
```
bklebe, line 38: suggestion: enable the rate limiter by default. Then we can get rid of the calls enabling it in `dev` and `prod`, simplifying configuration and ensuring that we default to rate limiting.

nlwstein: My only concern with that is it would then require memcached running locally - this feature was built with the intent of it being a global limiter, so there's no local-only implementation to fall back on.

apps/api_web/lib/api_web/plugs/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,54 @@
...
+  import Phoenix.Controller, only: [render: 3, put_view: 2]
+
+  alias ApiWeb.RateLimiter.RateLimiterConcurrent
+
+  @rate_limit_concurrent_config Application.compile_env!(:api_web, :rate_limiter_concurrent)
+
+  def init(opts), do: opts
+
+  def call(conn, _opts) do
+    if enabled?() do
```
paulswartz, line 16: suggestion: can we change this to check the rate limit and log the result if `memcache` is true, but only enforce the rate limit if `enabled` is true? I'm thinking about how we roll this out, and I don't think there's a good way to understand the expected values for some of our higher-throughput applications (dotcom, screens) without getting some benchmark data first.

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,170 @@
...
+    }
+  end
+
+  def get_pid_key(pid) do
+    sub_key = pid |> :erlang.pid_to_list() |> to_string
+    get_uuid() <> sub_key
+  end
+
+  defp get_uuid do
+    {:ok, uuid} = GenServer.call(__MODULE__, :get_uuid)
```
paulswartz, line 35: suggestion: this is creating another bottleneck, as the server can only respond to one `get_uuid` request at a time. Some options would be:
- a protected ETS table based on the name of the server
- a `:persistent_term` value set when the server starts

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,170 @@
...
+  def memcache? do
+    Keyword.fetch!(@rate_limit_concurrent_config, :memcache) == true
+  end
+
+  def handle_call(:get_uuid, _from, state) do
+    {:reply, {:ok, state.uuid}, state}
+  end
+
+  def memcache_update(key, default_value, update_fn) do
+    Memcache.cas(Supervisor.random_child(), key, update_fn, default: default_value)
```
paulswartz, line 168: praise: great find!

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,170 @@
...
+  use GenServer
+  require Logger
+  alias ApiWeb.RateLimiter.Memcache.Supervisor
+
+  @rate_limit_concurrent_config Application.compile_env!(:api_web, :rate_limiter_concurrent)
+
+  def start_link([]), do: GenServer.start_link(__MODULE__, nil, name: __MODULE__)
+
+  def init(_) do
+    {:ok, %{uuid: UUID.uuid1()}}
```
paulswartz, line 16: suggestion: if all you need is some random bytes to use as a prefix, you don't need a new library:
```elixir
iex(1)> Base.url_encode64(:crypto.strong_rand_bytes(16), padding: false)
"OFfLZZsevvnNTepVgztYBg"
```

nlwstein: I have historically used GUID/UUID algorithms for this because it's guaranteed unique, but this is probably good enough / comparable 😄

paulswartz: Gotcha; if we're already including it then this is fine.

apps/api_web/lib/api_web/plugs/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  defp put_concurrent_rate_limit_headers(conn, limit, remaining) do
+    conn
+    |> put_resp_header("x-concurrent-ratelimit-limit", "#{limit}")
+    |> put_resp_header("x-concurrent-ratelimit-remaining", "#{remaining}")
+  end
+
+  def enabled? do
+    Keyword.fetch!(@rate_limit_concurrent_config, :enabled) == true
+  end
```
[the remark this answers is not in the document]
bklebe: nitpick (non-blocking)

apps/api_web/lib/api_web/rate_limiter/memcache.ex
```diff
@@ -5,15 +5,9 @@ defmodule ApiWeb.RateLimiter.Memcache do
...
-    clear_interval = div(clear_interval_ms, 1000)
-
-    connection_opts =
-      [ttl: clear_interval * 2] ++ ApiWeb.config(RateLimiter.Memcache, :connection_opts)
-
-    Supervisor.start_link(connection_opts)
+  @impl true
+  def start_link(_) do
+    {:ok, self()}
   end
```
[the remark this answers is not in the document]
nlwstein: I have removed `start_link/1` from the behavior as it is no longer necessary here. The code that ultimately invokes the `start_link` has been updated to check if the function exists first instead, to handle the ETS based implementation.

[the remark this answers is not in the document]
nlwstein: > question: what's the intention of this change? It appears to turn this `start_link` into a no-op, instead of starting the process defined by the current module, as is convention.

I found that it was simplest to fix this by simply making the memcache implementation a GenServer, to get rid of this misleading code.

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
nlwstein: I have updated to use `:persistent_term`.

apps/api_web/lib/api_web/plugs/rate_limiter_concurrent.ex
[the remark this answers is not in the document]
nlwstein: I have removed the `== true` but I don't think it will work as just `@rate_limit_concurrent_config.enabled`, because the config isn't already in a map-like structure.

apps/api_web/config/config.exs
nlwstein: This should be fixed now, they should be disabled in `dev`.

apps/api_web/lib/api_web/plugs/rate_limiter_concurrent.ex
nlwstein: I have added two additional feature flags: `:log_statistics` and `:limit_users` to support these behaviors, and implemented a log statement w/ API key + current state, and added :limit_users == true to the condition that actually blocks the user from accessing until their cooldown.

apps/api_web/lib/api_web/event_stream.ex
```diff
@@ -130,6 +138,13 @@ defmodule ApiWeb.EventStream do
   end
 
   defp unsubscribe(state) do
+    if Map.has_key?(state.conn.assigns, :api_user) do
```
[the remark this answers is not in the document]
nlwstein: I feel it is worth having the condition and warning given that this is a separate function but I also don't feel super strongly - if you want it removed, I trust your judgment 🙂

apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex
```diff
@@ -0,0 +1,167 @@
...
+  def start_link([]), do: GenServer.start_link(__MODULE__, nil, name: __MODULE__)
+
+  def init(_) do
+    uuid = UUID.uuid1()
+    :persistent_term.put(@uuid_key, uuid)
+    {:ok, %{uuid: uuid}}
+  end
+
+  defp lookup(%ApiWeb.User{} = user, event_stream?) do
+    type = if event_stream?, do: "event_stream", else: "static"
```
paulswartz, line 22: question: it doesn't look like `type` is ever used?

nlwstein: Removed

paulswartz (approved):

PR #698, opened 2023-11-03
🍎🚫💰 Surface prediction 'uncertainty' values through v3 API

bfauble:
#### Summary of changes

**Asana Ticket:** [🍎🚫💰 Surface prediction 'uncertainty' values through v3 API](https://app.asana.com/0/295455480314405/1205761861495479/f)

* Parse uncertainty values (if present) from trip updates and add to prediction model.
* Add `arrival_uncertainty` and `departure_uncertainty` attributes to `Prediction` resource.
* Note: fields currently not present in concentrate feed. Pending: [🧠🚫💰 Surface non-revenue train information through our public-facing data feeds/API](https://app.asana.com/0/505721188639414/1205761861495473/f)

example response: 
```
{
  "data": [
    {
      "attributes": {
        "arrival_time": null,
        "arrival_uncertainty": null,
        "departure_time": "2023-11-03T13:13:00-04:00",
        "departure_uncertainty": null,
        "direction_id": 1,
        "schedule_relationship": null,
        "status": null,
        "stop_sequence": 50
      },
      "id": "prediction-59095632-70094-50",
      "relationships": {
        "route": {
          "data": {
            "id": "Red",
            "type": "route"
          }
        },
        "stop": {
          "data": {
            "id": "70094",
            "type": "stop"
          }
        },
        "trip": {
          "data": {
            "id": "59095632",
            "type": "trip"
          }
        },
        "vehicle": {
          "data": {
            "id": "R-54795DFB",
            "type": "vehicle"
          }
        }
      },
      "type": "prediction"
    },
    ...

### DIFF
--- a/apps/api_web/lib/api_web/controllers/prediction_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/prediction_controller.ex
@@ -13,7 +13,7 @@ defmodule ApiWeb.PredictionController do
   alias ApiWeb.LegacyStops
   alias State.Prediction
 
-  @filters ~w(stop route trip latitude longitude radius direction_id stop_sequence route_type route_pattern)s
+  @filters ~w(stop route trip latitude longitude radius direction_id stop_sequence route_type route_pattern revenue)s
   @pagination_opts ~w(offset limit order_by)a
   @includes ~w(schedule stop route trip vehicle alerts)
 
@@ -61,6 +61,7 @@ defmodule ApiWeb.PredictionController do
     filter_param(:stop_id, includes_children: true)
     filter_param(:id, name: :route)
     filter_param(:id, name: :trip)
+    filter_param(:revenue, desc: "Filter predictions by revenue status.")
 
     parameter("filter[route_pattern]", :query, :string, """
     Filter by `/included/{index}/relationships/route_pattern/data/id` of a trip. Multiple `route_pattern_id` #{comma_separated_list()}.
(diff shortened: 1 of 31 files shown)

### CODE REVIEW

**Comment by paulswartz** [comment]
question: is there a reason to ship this before we're providing the data in Concentrate?

File: apps/parse/lib/parse/trip_updates.ex [hunk 1]
```diff
@@ -54,6 +56,12 @@ defmodule Parse.TripUpdates do
     nil
   end
 
+  defp parse_uncertainty(%{uncertainty: uncertainty}) when is_integer(uncertainty) do
```
nlwstein, line 59: Since your documentation (nice!) only allows three values, would it make sense to only allow those three values through? I think you can use a guard like `when is_integer(uncertainty) and uncertainty in [60, 120, 360]` but I haven't tested it.

apps/api_web/lib/api_web/controllers/prediction_controller.ex
```diff
@@ -261,6 +261,34 @@ defmodule ApiWeb.PredictionController do
               example: "2017-08-14T15:38:58-04:00"
             )
 
+            arrival_uncertainty(
```
nlwstein, line 264: Does this get sucked into Swagger? I can't remember, but if it doesn't, might be worth outputting this to the docs wherever this field shows up.

nlwstein (approved):
A couple of random thoughts but overall this looks good! 🚀

apps/parse/lib/parse/trip_updates.ex
paulswartz: Does this incorporate the uncertainty values that Swiftly uses?

paulswartz (approved):

PR #699, opened 2023-11-08
chore: first time setup updates: poetry in .tool-versions and dynamodb file in .gitignore

meagharty:
#### Summary of changes

**Asana Ticket:** No ticket

* Add poetry back to .tool-versions
If poetry was intentionally removed from .tool-versions, I could instead update the README to remove the line `Run asdf plugin-add poetry to add Poetry plugin`.

* Add local dynamodb metadata file to .gitignore

- **NOTE** on undocumented step: For the local dynamoDB setup, I needed to install java on my machine. I installed java globally via asdf. I didn't add instructions to the api_accounts/README but I could if anyone felt strongly about it.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
 elixir 1.14.5-otp-25
 erlang 25.3.2.7
 python 3.9.16
-
+poetry 1.3.1
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,4 +1,4 @@
 elixir 1.14.5-otp-25
 erlang 25.3.2.7
 python 3.9.16
-poetry 1.3.1
+poetry 1.7.0
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ cover
 # Python artifacts (for locust)
 __pycache__
 
-# local db 
+# local db
 shared-local-instance.db
 /bin
+dynamodb-local-metadata.json


.tool-versions
```diff
@@ -1,4 +1,4 @@
 elixir 1.14.5-otp-25
 erlang 25.3.2.7
 python 3.9.16
-
+poetry 1.3.1
```
bklebe, line 4: question: is there a reason not to use the latest poetry (1.7.0 apparently)?

nlwstein:
@meagharty This is totally not blocking but I have found it helpful to run the dynamodb in Docker - a simple version being `docker run -p 8000:8000 amazon/dynamodb-local` but you could preserve the data in a volume if you wanted to as well.

.tool-versions
meagharty: Not really, beyond that I didn't test that. A minor version upgrade should be fine.

meagharty:
> @meagharty This is totally not blocking but I have found it helpful to run the dynamodb in Docker - a simple version being `docker run -p 8000:8000 amazon/dynamodb-local` but you could preserve the data in a volume if you wanted to as well.

Thank you!! good to know that would just work with our setup.

Is it worth updating the docs with that suggestion? I'm not sure how cautious we are with updating public-facing developer documents.

nlwstein:
> > @meagharty This is totally not blocking but I have found it helpful to run the dynamodb in Docker - a simple version being `docker run -p 8000:8000 amazon/dynamodb-local` but you could preserve the data in a volume if you wanted to as well.
> 
> @nlwstein Thank you!! good to know that would just work with our setup.
> 
> Is it worth updating the docs with that suggestion? I'm not sure how cautious we are with updating public-facing developer documents.

I don't think it needs to be updated yet, just figured I would mention it bc it made my setup easy 🙂 

This looks good to me!

nlwstein (approved):

PR #704, opened 2023-11-16
feat: allow admins to set API key limit by minute

meagharty:
#### Summary of changes

**Asana Ticket:** [🍎 Adjust v3 API Admin Panel "edit" section to allow adjusting per-minute limit](https://app.asana.com/0/1205718273834959/1205807844528810)

This updates the admin edit key form to show the limit as a per-minute limit instead of a daily limit. The admin view for viewing keys already displays the request limit as a per-minute limit in production. The backend stores the limit as a daily limit.

This hard-codes the portal and admin views to show a key's limit as its per minute limit, ignoring the configured interval for clearing the limit. Production is configured to have a clear_interval of one minute (60,000 ms), meaning a per-minute limit.

There's a bit of complexity with our rate limiter that we have in part to allow tests to configure a shorter clear_interval. We have configurations by environment, see https://github.com/mbta/api/blob/b6b1d18aac734d3c25cbff851b459e133056b85f/apps/api_web/config/prod.exs#L57-L60 

I removed view code like the code below because the view doesn't need to know these details, like the rate-limiter does.
https://github.com/mbta/api/blob/b6b1d18aac734d3c25cbff851b459e133056b85f/apps/api_web/lib/api_web/api_view_helpers.ex#L250

By default when a key is created, it doesn't have a daily_limit set. There's some extra code (currently in `api_view_helpers`) to preserve the existing default of `max_registered_per_interval` in `per_minute` when the daily_limit on a key isn't set.

### DIFF
--- a/apps/api_web/lib/api_web/api_view_helpers.ex
+++ b/apps/api_web/lib/api_web/api_view_helpers.ex
@@ -240,9 +240,29 @@ defmodule ApiWeb.ApiViewHelpers do
     |> String.replace("/", "%2F")
   end
 
+  def default_registered_per_interval() do
+    ApiWeb.RateLimiter.max_registered_per_interval()
+  end
+
   def limit(%ApiAccounts.Key{} = key) do
     key
     |> ApiWeb.User.from_key()
+    |> ApiWeb.RateLimiter.max_requests()
+    |> trunc()
+  end
+
+  def limit_value(%ApiAccounts.Key{} = key) do
+    key
+    |> ApiWeb.User.from_key()
+    |> limit_or_default_of_nil()
+  end
+
+  defp limit_or_default_of_nil(%ApiWeb.User{limit: nil}) do
+    nil
+  end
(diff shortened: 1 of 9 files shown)


apps/api_web/lib/api_web/api_view_helpers.ex
```diff
@@ -240,13 +240,50 @@ defmodule ApiWeb.ApiViewHelpers do
...
+  end
+
   def limit(%ApiAccounts.Key{} = key) do
     key
     |> ApiWeb.User.from_key()
     |> ApiWeb.RateLimiter.max_requests()
     |> trunc()
   end
 
+  def limit_value(%ApiAccounts.Key{} = key) do
```
meagharty, line 254: naming is hard, feel free to suggest a better name. this refers to the stored value of limit (as daily_value) in the database.

dks-mbta (approved):

PR #705, opened 2023-11-20
🍎🚫💰 Add "Trip Revenue Status" 'revenue' filter to the API  (Trip/Vehicle/Prediction)

bfauble:
#### Summary of changes

**Asana Ticket:** [🍎🚫💰 Add "Trip Revenue Status" 'revenue_status' filter to the API ](https://app.asana.com/0/295455480314405/1205787751466825/f)

* add `revenue` filter to `Trip` resource
* add `revenue` filter to `Vehicle` resource
* add `revenue` filter to `Prediction` resource
* trips from gtfs currently being treated as `revenue: :REVENUE` by default
* `revenue: true | false` values from VehiclePosition feed used for `Vehicle`
* `revenue: true | false` values from TripUpdates feed used for `Prediction`
* `revenue: :REVENUE | :NON_REVENUE` values from `Predictions` used for `Trips.Added`

Need to merge after: https://github.com/mbta/rtr/pull/1201

### DIFF
--- a/apps/state/lib/state/trip.ex
+++ b/apps/state/lib/state/trip.ex
@@ -171,7 +171,7 @@ defmodule State.Trip do
       |> build_filters(:direction_id, filters[:direction_id])
       |> build_filters(:route_pattern_id, filters[:route_patterns])
       |> build_filters(:id, filters[:ids])
-      |> build_filters(:revenue, Map.get(filters, :revenue, :REVENUE))
+      |> build_filters(:revenue, Map.get(filters, :revenue, [:REVENUE]))
 
     idx = get_index(filters)
     trips = State.Trip.Added.select(matchers, idx) ++ State.Trip.select(matchers, idx)
--- a/apps/state/lib/state/vehicle.ex
+++ b/apps/state/lib/state/vehicle.ex
@@ -62,7 +62,7 @@ defmodule State.Vehicle do
     [%{}]
     |> build_filters(:effective_route_id, Map.get(filters, :routes), filters)
     |> build_filters(:route_type, Map.get(filters, :route_types), filters)
-    |> build_filters(:revenue, Map.get(filters, :revenue, :REVENUE), filters)
+    |> build_filters(:revenue, Map.get(filters, :revenue, [:REVENUE]), filters)
     |> State.Vehicle.select(idx)
     |> do_post_search_filter(filters)
   end


apps/parse/lib/parse/trips.ex
```diff
@@ -60,4 +61,7 @@ defmodule Parse.Trips do
   end
 
   defp copy_if_not_blank(_), do: nil
+
+  defp parse_is_revenue_service("NONREV-" <> _), do: false
```
bklebe, line 65: @jfabi is this the best way we have to determine whether a trip is revenue or not?

bklebe (commented):
Looks like a good start @bfauble, do you plan to follow up with the predictions resource later?

apps/parse/lib/parse/trips.ex
paulswartz: I thought the original plan was to add a `revenue: false` attribute to the TripDescriptor?

bklebe: I don't remember what the actual plan was, for some reason I hadn't realized that nonrevenue vehicles still had a TripDescriptor at all. But I don't think we want to be parsing the trip IDs, that's an implementation detail of RTR.

paulswartz (commented):
question: why is the filter `revenue_service` but the value in the resources `revenue`?

suggestion(non-blocking): if the field in the API is already `revenue`, what about making the query more directly based on that? Something like `/vehicles?filter[revenue]=true,false` to get all the vehicles?

apps/api_web/lib/api_web/params.ex
```diff
@@ -208,6 +208,26 @@ defmodule ApiWeb.Params do
...
+    |> Enum.sort()
+    |> revenue()
+  end
+
+  for value <- ~w(NON_REVENUE REVENUE)a do
+    def revenue([unquote(Atom.to_string(value))]), do: {:ok, unquote(value)}
+  end
+
+  def revenue(["NON_REVENUE", "REVENUE"]), do: {:ok, :ALL}
+  def revenue(_value), do: :error
```
paulswartz, line 229:
```suggestion
```suggestion
  def revenue(values) do
    values
    |> split_on_comma()
    |> Enum.reduce_while({:ok, []}, fn
      "REVENUE", {:ok, acc} -> {:cont, {:ok, [:REVENUE | acc]}}
      "NON_REVENUE", {:ok, acc} -> {:cont, {:ok, [:NON_REVENUE | acc]}}
      _ -> {:halt, :error}
    end)
  end
```

paulswartz (commented):
I think the parsing could be a little simpler (see my suggestion) but otherwise this looks good! Can you throw it up on one of the dev environments for testing?

apps/api_web/lib/api_web/views/prediction_view.ex
```diff
@@ -204,6 +206,12 @@ defmodule ApiWeb.PredictionView do
     |> String.upcase()
   end
 
+  def revenue(%{revenue: atom}) do
+    atom
+    |> Atom.to_string()
+    |> String.upcase()
```
paulswartz, line 212: question: isn't the atom already uppercase? In which case you can pass it directly (I think).

apps/state/lib/state/trip.ex
```diff
@@ -170,6 +171,7 @@ defmodule State.Trip do
       |> build_filters(:direction_id, filters[:direction_id])
       |> build_filters(:route_pattern_id, filters[:route_patterns])
       |> build_filters(:id, filters[:ids])
+      |> build_filters(:revenue, Map.get(filters, :revenue, :REVENUE))
```
paulswartz, line 174: question: the spec says this is taking a list, so should the default be `[:REVENUE]`?

apps/state/lib/state/vehicle.ex
```diff
@@ -61,6 +62,7 @@ defmodule State.Vehicle do
     [%{}]
     |> build_filters(:effective_route_id, Map.get(filters, :routes), filters)
     |> build_filters(:route_type, Map.get(filters, :route_types), filters)
+    |> build_filters(:revenue, Map.get(filters, :revenue, :REVENUE), filters)
```
paulswartz, line 65: question: should the default be `[:REVENUE]`?

paulswartz (approved):

PR #738, opened 2024-01-12
fix: clear Logger.metadata between requests

paulswartz:
Bandit keeps processes alive between requests, so we need to manually
clear the metadata to avoid it being kept between requests.

## Manual Testing
`curl -H "mbta-version: 2017-11-28" "http://localhost:4000/alerts/" "http://localhost:4000/alerts/?api_key=__configure_me__"`

### Before
```
2024-01-12 12:13:25.961 request_id=F6mo3h-7OA5-ItAAABqE api_key=anonymous api_version=2017-11-28 records=101 ip=127.0.0.1 pid=<0.1462.0> [info] state=set duration=56.644 status=200 params={} path=/alerts/ method=GET format=json-api controller=ApiWeb.AlertController action=index
2024-01-12 12:13:26.108 request_id=F6mo3ivHCwp-ItAAAAkn api_key=__configure_me__ api_version=2017-11-28 records=101 pid=<0.1462.0> [warning] state=set duration=1.072 status=403 params={} path=/alerts/ method=GET
```
### After
```
2024-01-12 12:12:50.735 request_id=F6mo1e6JwuZNhccAAABl api_key=anonymous api_version=2017-11-28 records=101 ip=127.0.0.1 pid=<0.1359.0> [info] state=set duration=15.370 status=200 params={} path=/alerts/ method=GET format=json-api controller=ApiWeb.AlertController action=index
2024-01-12 12:12:50.917 request_id=F6mo1fo-pI9NhccAABCC api_key=__configure_me__ ip=127.0.0.1 pid=<0.1359.0> [warning] state=set duration=1.255 status=403 params={} path=/alerts/ method=GET
```
Note that the "records" key is not present in invalid log anymore.

### DIFF
--- a/apps/api_web/lib/api_web/plugs/clear_metadata.ex
+++ b/apps/api_web/lib/api_web/plugs/clear_metadata.ex
@@ -1,0 +1,26 @@
+defmodule ApiWeb.Plugs.ClearMetadata do
+  @moduledoc """
+  Clear Logger metadata at the start of a new request.
+
+  Bandit is more agressive about re-using processes than Cowboy was, which means we
+  can't rely on the old behavior of the Logger metadata being automatically cleared
+  when the process is terminated.
+
+  This takes a list of metadata keys to clear.
+  """
+  @behaviour Plug
+
+  @impl Plug
+  def init(keys_to_clear) do
+    for key <- keys_to_clear do
+      {key, nil}
+    end
+  end
+
+  @impl Plug
+  def call(conn, metadata) do
+    _ = Logger.metadata(metadata)
+
+    conn
+  end
+end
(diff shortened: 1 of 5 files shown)


apps/api_web/lib/api_web/plugs/clear_metadata.ex
```diff
@@ -0,0 +1,26 @@
...
+  @impl Plug
+  def init(keys_to_clear) do
+    for key <- keys_to_clear do
+      {key, nil}
+    end
+  end
+
+  @impl Plug
+  def call(conn, metadata) do
+    _ = Logger.metadata(metadata)
```
Whoops, line 22: Why not use [reset_metadata](https://hexdocs.pm/logger/Logger.html#reset_metadata/1)? Are we logging something we want to retain?

Whoops: Fair enough! My concern is that we'll add a key and forget to add it here, but if that becomes a problem, we can cross that bridge then.

Whoops (approved):

PR #742, opened 2024-01-23
chore: replace ex_csv with nimble_csv

nlwstein:
#### Summary of changes

Fixes broken build by replacing deprecated library.


nlwstein:
The build may still fail due to the Wickford Junction error (present on default branch) but no other tests should fail.

apps/parse/lib/parse/simple.ex
```diff
@@ -36,7 +38,13 @@ defmodule Parse.Simple do
 
   def parse(blob, row_callback) do
     blob
-    |> ExCsv.parse!(headings: true)
+    |> String.trim()
+    |> CSV.parse_string(skip_headers: false)
+    |> Stream.transform(nil, fn
```
nlwstein, line 43: The trim is to address some cases where whitespace results in a blank row being used as a header row. 

The main logic here is basically just this: https://jacobswanner.com/development/2022/nimble-csv-maps/

PR #747, opened 2024-02-06
fix(stops): add updated route_pattern for CR-Providence that includes Forest Hills

meagharty:
#### Summary of changes

**Asana Ticket:** [[extra] Date Filter Hiding Stop from End Point](https://app.asana.com/0/584764604969369/1206469190651881/f)

I was able to fix the issue reported locally by updating the state config to include the current route_pattern for the two trips including Forest Hills (https://api-dev.mbtace.com/trips/AMTKSchCh651068-841 and https://api-dev.mbtace.com/trips/AMTKSchCh651474-839). We do have an extended disruption on CR-Providence relating to Amtrak that affects these trips. I'm not certain if that specifically caused the change in the route_pattern_id. 

I also [started a slack thread](https://mbta.slack.com/archives/C06CAP4N91C/p1707233921986039) about this configuration vs. the canonical GTFS work in gtfs_creator.

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -219,6 +219,8 @@ config :state, :stops_on_route,
     "Shuttle-BraintreeSouthWeymouth-0-" => true,
     # Providence trains stopping at Forest Hills
     "CR-Providence-d01bc229-0" => true,
+    # Providence trains stopping at Forest Hills
+    "CR-Providence-410d3e77-0" => true,
     # Haverhill Line shuttles
     "Shuttle-BallardvaleMaldenCenter-0-" => true,
     "Shuttle-HaverhillMaldenCenter-0-" => true,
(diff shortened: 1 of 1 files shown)


apps/state/config/config.exs
```diff
@@ -218,7 +218,7 @@ config :state, :stops_on_route,
     # Kingston Line shuttles to/from South Weymouth
     "Shuttle-BraintreeSouthWeymouth-0-" => true,
     # Providence trains stopping at Forest Hills
-    "CR-Providence-d01bc229-0" => true,
```
meagharty, line 221: See PR description for more details but this route pattern 404s at the moment:
https://api-dev.mbtace.com/route_patterns/CR-Providence-d01bc229-0

meagharty:
There's a test failing that's unrelated to this PR. This PR will need the update to always resolve route_type within an alert, it's in Brian's PR (https://github.com/mbta/api/pull/746). If that's merged soon I'll rebase on it.

nlwstein (approved):

PR #751, opened 2024-02-09
feat: concurrent connections limiter

nlwstein:
#### Summary of changes

**Asana Ticket:** [🍎 Reject API requests when there are too many concurrent requests](https://app.asana.com/0/295455480314405/1205450740821028/f)

This provides a new rate limiter that is mainly useful for clamping down on too many active streaming connections. It is also able to track and control long-running "static" requests (for example, trying to load all of the predictions at once). It is configurable on a per-user basis in the admin panel (`-1` to disable). The largest value between user config and the base config is used. 

The static concurrency locks are released in real-time, whereas the streaming ones depend on the hibernate loop cycle, so there can be a latency of under a minute for releasing those. The limits I have added to the config file are guidance for testing, not necessarily for production. 

`Opinion`: I am not sure if `memcached` is the ideal tool for this, but I used it because I don't want to introduce new infrastructure if possible and it was already in-use for a similar job. Changing out the storage mechanism in the future should be relatively easy. 

The whole rate limiting system seems like a good use case for Redis (or perhaps there's something new that's even better), primarily because: 

1. You can more efficiently filter / upsert nested data, or even just on a key prefix basis. 
2. It has a concept of locks and transactions


paulswartz:
Can we deploy this to one of the dev-* environments?

nlwstein:
@paulswartz I have confirmed that this is working and properly establishing connection to memcache in dev-blue: https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22api-dev-blue-application%22%20ApiWeb.Plugs.RateLimiterConcurrent&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-30m%40m&latest=now&sid=1707509690.6290337

nlwstein:
@bklebe This is on dev-blue again, and seems to be working as expected: https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22api-dev-blue-application%22%20ApiWeb.Plugs.RateLimiterConcurrent%20event_stream%3Dtrue&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=-30m%40m&latest=now&sid=1707756574.7074763

paulswartz:
@nlwstein hmm, something looks weird about that log. for the first two requests, "concurrent" + "limit" == "remaining", but for the last request, "concurrent" + "limit" != "remaining". That makes me think we either didn't count a request, or one expired from the lock without being actually disconnected.

nlwstein:
> @nlwstein hmm, something looks weird about that log. for the first two requests, "concurrent" + "limit" == "remaining", but for the last request, "concurrent" + "limit" != "remaining". That makes me think we either didn't count a request, or one expired from the lock without being actually disconnected.

@paulswartz I don't think that concurrent field is coming from my code. I'm not familiar with it and suspect it comes from another package. 

My log statement begins with the module name:

```
        Logger.error(
          "ApiWeb.Plugs.RateLimiterConcurrent event=request_statistics api_user=#{conn.assigns.api_user.id} at_limit=#{at_limit?} remaining=#{remaining - 1} limit=#{limit} event_stream=#{event_stream?}"
        )
```

paulswartz:
That's right, the `concurrent` log is coming from the existing concurrency tracker. But since there's only one node, it should always agree with the new concurrency tracker and it doesn't look like that's the case.

paulswartz (commented):
putting in a temporary review so that it drops off the reminder list. open issue is that the existing concurrency logger does not always agree with the new concurrency limiter, and we're not sure why.

nlwstein:
> putting in a temporary review so that it drops off the reminder list. open issue is that the existing concurrency logger does not always agree with the new concurrency limiter, and we're not sure why.

I tried to reproduce this locally, and could not seem to make it happen. The counters were synced, even having the same 'ghost' requests after putting my laptop to sleep for the night 😂 

My guess as to what happened is that a bunch of requests were opened and closed in a short period of time, simultaneously, and when that log fired there was a race between when the request tracker updated its logging metadata, and [this cleanup filter](https://github.com/mbta/api/blob/dceca730d666059da2b9b3178adc7f678e7fa14e/apps/api_web/lib/api_web/rate_limiter/rate_limiter_concurrent.ex#L51) which fires right before that particular log message is emitted. It also relies on a (configurable) heartbeat, so it's possible that it failed a heartbeat check, reducing the count of monitored processes on this feature before the callback for the request tracker fired. 

I also redeployed to `dev-blue` and it [didn't seem to happen again](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22api-dev-blue-application%22%20ApiWeb.Plugs.RateLimiterConcurrent%20event_stream%3Dtrue&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1708491600&latest=1708578000&sid=1708527533.9688880). 

Maybe once we deploy this we should write a Splunk query to determine how frequently this happens? Ideally, it's just the heartbeat mechanism doing it's thing 🙏

paulswartz:
I see it happening a lot in [Splunk](https://mbta.splunkcloud.com/en-US/app/search/search?q=search%20index%3D%22api-dev-blue-application%22%20ApiWeb.Plugs.RateLimiterConcurrent%20%0A%7C%20eval%20expected_limit%3Dconcurrent%2Bremaining%0A%7C%20search%20expected_limit%20!%3D%20limit&display.page.search.mode=verbose&dispatch.sample_ratio=1&workload_pool=standard_perf&earliest=1708491600&latest=1708578000&sid=1708527879.9690323)?

paulswartz (commented):
empty review pending more investigation

paulswartz (approved):
Had a chat with Firas in Slack. Summary:
- merge this as is
- continue to investigate the discrepancy between the original counter and the distributed counter
- if we're unable to make progress on the accuracy of the distributed counter by 2024-03-31, we'll roll it back and discuss other approaches

PR #764, opened 2024-02-21
test: switch to using `@example.com` so e-mails work on the VPN

paulswartz:
Our e-mail validation library doesn't validate `@mbta.com` while on the VPN:
```
iex(1)> EmailChecker.valid?("pswartz@mbta.com")
false
```


bklebe (changes requested):
@paulswartz looks like you have a `mix format` to run :)

bklebe (approved):

bklebe:
Thanks for fixing this!

PR #766, opened 2024-02-22
tests: run MQTT test using container

bklebe:
tests: run MQTT test using container

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -2,7 +2,7 @@ name: Elixir CI
 
 on:
   schedule:
-    - cron: '0 5 * * *'
+    - cron: "0 5 * * *"
   push:
     branches: master
   pull_request:
@@ -127,6 +127,12 @@ jobs:
         image: amazon/dynamodb-local:2.0.0
         ports:
           - 8000:8000
+      mosquitto:
+        image: "eclipse-mosquitto:2.0"
+        ports:
+          - "1883:1883"
+        volumes:
+          - ./mosquitto.conf:/mosquitto/config/mosquitto.conf
 
     steps:
       - uses: actions/checkout@v3
--- a/README.md
+++ b/README.md
@@ -16,6 +16,17 @@ To start your Phoenix app:
 
 Now you can visit [`localhost:4000`](http://localhost:4000) from your browser.
 
+## Tests
+
+To run the tests, first install and setup Podman:
+  1. `brew install podman`
+  2. `podman machine init`
+  3. `podman machine start`
+
+Then, start the Compose configuration in a separate window or tab and run the tests: 
+  1. `podman compose up` 
+  2. `mix test`
+
 ## Environment Variables
 
 In addition to the Elixir config files, the V3 API allows runtime configuration through a collection of environment variables.
--- a/apps/state_mediator/test/state_mediator/mqtt_mediator_test.exs
+++ b/apps/state_mediator/test/state_mediator/mqtt_mediator_test.exs
@@ -28,8 +28,8 @@ defmodule StateMediator.MqttMediatorTest do
 
   # @moduletag capture_log: true
   @opts [
-    url: "mqtt://test.mosquitto.org",
-    topic: "home/#",
+    url: "mqtt://localhost:1883",
+    topic: "home/test",
     state: __MODULE__.StateModule
   ]
 
@@ -55,6 +55,16 @@ defmodule StateMediator.MqttMediatorTest do
 
     test "are sent to the state module" do
       {:ok, _pid} = start_link(@opts)
+
+      {:ok, client} = EmqttFailover.Connection.start_link(configs: [@opts[:url]])
+      Process.sleep(500)
+
+      :ok =
+        EmqttFailover.Connection.publish(client, %EmqttFailover.Message{
+          topic: "home/test",
+          payload: "online"
+        })
+
       assert_receive {:updated, contents}, 5_000
       assert <<_::binary>> = contents
     end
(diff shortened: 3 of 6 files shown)


apps/state_mediator/test/state_mediator/mqtt_mediator_test.exs
```diff
@@ -55,6 +55,16 @@ defmodule StateMediator.MqttMediatorTest do
 
     test "are sent to the state module" do
       {:ok, _pid} = start_link(@opts)
+
+      {:ok, client} = EmqttFailover.Connection.start_link(configs: [@opts[:url]])
+      Process.sleep(500)
```
paulswartz, line 60: question: why do you need this?

.github/workflows/elixir.yml
```diff
@@ -127,9 +127,17 @@ jobs:
         image: amazon/dynamodb-local:2.0.0
         ports:
           - 8000:8000
+      mosquitto:
+        image: eclipse-mosquitto:2.0
+        ports:
+          - 1883:1883
+        volumes:
+          - ./mosquitto:/mosquitto/config/
```
paulswartz, line 135: question: any reason this is different from what's in `docker-compose.yml`?

.github/workflows/elixir.yml
```diff
@@ -127,9 +127,17 @@ jobs:
...
+        options: --name mqtt
 
     steps:
       - uses: actions/checkout@v3
+      - run: docker restart mqtt
```
paulswartz, line 140: question: can you say more about this?

bklebe: When GitHub actions initializes the service, the runner is bare, so we can't mount the config file. This is a workaround from https://github.com/orgs/community/discussions/42127#discussioncomment-7591609.

.github/workflows/elixir.yml
bklebe: That was a mistake, should be fixed now.

apps/state_mediator/test/state_mediator/mqtt_mediator_test.exs
bklebe: Which part? We need a client to publish a message, `test.mqtt.org` seems to do this automatically whenever a client connects. We need `Process.sleep` because `Connection.start_link` is asynchronous and `publish` will fail if we don't pause first, because the client is not actually connected.

.github/workflows/elixir.yml
bklebe: Added a `name` explaining this step.

apps/state_mediator/test/state_mediator/mqtt_mediator_test.exs
paulswartz: It can send a message when connected: https://hexdocs.pm/emqtt_failover/EmqttFailover.ConnectionHandler.Parent.html (example: https://github.com/mbta/concentrate/blob/a478d5fc96db2d259974cdc86516850a4fa8a713/lib/concentrate/sink/mqtt.ex)

bklebe: Updated with `EmqttFailover.ConnectionHandler.Parent`.

apps/state_mediator/test/state_mediator/mqtt_mediator_test.exs
```diff
@@ -55,8 +55,15 @@ defmodule StateMediator.MqttMediatorTest do
 
     test "are sent to the state module" do
       {:ok, _pid} = start_link(@opts)
-      assert_receive {:updated, contents}, 5_000
-      assert <<_::binary>> = contents
+
+      {:ok, _client} =
```
paulswartz, line 59:
```suggestion
      {:ok, client} =
```

apps/state_mediator/test/state_mediator/mqtt_mediator_test.exs
```diff
@@ -55,8 +55,15 @@ defmodule StateMediator.MqttMediatorTest do
 
     test "are sent to the state module" do
       {:ok, _pid} = start_link(@opts)
-      assert_receive {:updated, contents}, 5_000
-      assert <<_::binary>> = contents
```
paulswartz, line 59: suggestion: you've dropped the assertions which are actually about the state module (these ones above).

bklebe:
@paulswartz this test seems to be flaky when we wait for the `:connected` message from `ConnectionHandler.Parent` in a way that it wasn't with `Process.sleep`. It's failing here but when I test it locally it still passes sometimes. Any idea why?

README.md
```diff
@@ -16,6 +16,17 @@ To start your Phoenix app:
 
 Now you can visit [`localhost:4000`](http://localhost:4000) from your browser.
 
+## Tests
+
+To run the tests, first install and setup Podman:
+  1. `brew install podman`
```
paulswartz, line 22: question: I missed this until now, but we've recommended `colima` as the Docker replacement for TID: https://github.com/mbta/technology-docs/blob/main/rfcs/accepted/0010-docker-desktop-replacement.md

Is there a reason that `colima` + docker compose doesn't work here? Should we be updating the RFC?

paulswartz (commented):
I can reproduce it locally with `while mix test test/state_mediator/mqtt_mediator_test.exs:73; sleep 1; end`. It looks like it's a race condition between the `MqttMediator` subscribing to the topic and the message being published.

I don't have a good idea here (other than adding a new interface to the `MqttMediator`), so you can go back to the 500ms wait and we'll move along.

README.md
bklebe: I'll try it with Colima; Podman is just what came to me off the top of my head as the FOSS Docker thing.

paulswartz (approved):

PR #781, opened 2024-03-20
chore: add CR-Newburyport shuttles during suspension to route_pattern…

meagharty:
…_prefix_overrides

#### Summary of changes

**Asana Ticket:** [🚧 🟣 Newburyport/Rockport shuttles, March 23-24](https://app.asana.com/0/584764604969369/1206682488997510/f)

### DIFF
--- a/apps/state/config/config.exs
+++ b/apps/state/config/config.exs
@@ -205,6 +205,10 @@ config :state, :stops_on_route,
     "Shuttle-BeverlyNewburyportLocal-0-" => true,
     "Shuttle-NewburyportSalemExpress-0-" => true,
     "Shuttle-NewburyportSalemLocal-0-" => true,
+    # Newburyport Line Shuttles during suspension
+    "Shuttle-BeverlyNorthStationLocal-" => true,
+    "Shuttle-NewburyportNorthStationLimited-" => true,
+    "Shuttle-RockportNorthStationLimited-" => true,
     # Fitchburg Line shuttles to/from Alewife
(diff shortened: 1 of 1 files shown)


apps/state/config/config.exs
```diff
@@ -205,6 +205,10 @@ config :state, :stops_on_route,
     "Shuttle-BeverlyNewburyportLocal-0-" => true,
     "Shuttle-NewburyportSalemExpress-0-" => true,
     "Shuttle-NewburyportSalemLocal-0-" => true,
+    # Newburyport Line Shuttles during suspension
+    "Shuttle-BeverlyNorthStationLocal-" => true,
```
jfabi, line 209: This first one isn't needed, because `route_id = "Shuttle-BeverlyNorthStationLocal"` is already represented in Line 189 above. In fact, it's because of L189 that [dotcom is showing North Station–Beverly](https://dev-green.mbtace.com/schedules/CR-Newburyport/timetable?date=2024-03-24&schedule_direction[direction_id]=0#direction-filter), rather than nothing at all.

meagharty: I was thinking the first `-0` and `-1` where the direction_id but I see that these route pattern ids are `Shuttle-BeverlyNorthStationLocal-0-1`, `Shuttle-BeverlyNorthStationLocal-0-0` (https://api-dev-green.mbtace.com/routes/Shuttle-BeverlyNorthStationLocal?include=route_patterns)

jfabi (approved):

PR #800, opened 2024-05-09
fix: add Providence/Stoughton shape for CR Rating Spring 2024

meagharty:
#### Summary of changes

Asana Ticket: [🚧🟣 Spring 2024 Commuter Rail rating, starting May 20](https://app.asana.com/0/584764604969369/1207118646684817/f)

* Update tests to reflect new Spring rating from Keolis

### DIFF
--- a/apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
+++ b/apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
@@ -195,13 +195,17 @@ defmodule StateMediator.Integration.GtfsTest do
       assert [%{name: "Braintree - Alewife"}, %{name: "Ashmont - Alewife"}] = shapes_1
     end
 
-    test "Providence/Stoughton has 2 non-ignored shapes each direction" do
+    test "Providence/Stoughton has 4 non-ignored shapes in direction 0 and 2 in direction 1" do
       [shapes_0, shapes_1] = shapes_in_both_directions("CR-Providence")
 
       assert [
+               # South Station - Wickford Junction via Back Bay
                shape_0,
+               # South Station - Stoughton via Back Bay
                %{id: "9890004"},
-               %{id: "SouthStationToStoughtonViaFairmount"}
+               %{id: "SouthStationToStoughtonViaFairmount"},
+               # South Station - Wickford Junction
+               %{id: "9890009"}
              ] = shapes_0
 
       assert shape_0.name =~ "South Station - Wickford Junction"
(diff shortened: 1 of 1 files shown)


apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
```diff
@@ -195,17 +195,31 @@ defmodule StateMediator.Integration.GtfsTest do
...
+                 %{id: "9890004"} -> true
+                 # South Station - Stoughton via Fairmount
+                 %{id: "SouthStationToStoughtonViaFairmount"} -> true
+                 # South Station - Wickford Junction
+                 %{id: "9890009"} -> true
+                 _ -> false
+               end
+             end)
+
+      assert Enum.count(remaining_shapes) >= 1
```
jzimbel-mbta, line 221: Should this assertion be >= 4? (Since the test is for at least 2 outbound shapes and exactly 2 inbound shapes -> at least 4 total shapes)

apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
```diff
@@ -195,17 +195,31 @@ defmodule StateMediator.Integration.GtfsTest do
...
+               case shape do
+                 # South Station - Stoughton via Back Bay
+                 %{id: "9890004"} -> true
+                 # South Station - Stoughton via Fairmount
+                 %{id: "SouthStationToStoughtonViaFairmount"} -> true
+                 # South Station - Wickford Junction
+                 %{id: "9890009"} -> true
+                 _ -> false
+               end
+             end)
```
jzimbel-mbta, line 219: How about
```suggestion
      Enum.each(remaining_shapes, fn shape ->
        assert shape[:id] in ["9890004", "SouthStationToStoughtonViaFairmount", "9890009"]
      end)
```
So that if there's an unexpected shape in the list, the test shows which one it is?

(`shape[:id]` can be `shape.id` if all elements are maps with an `:id` field--I wasn't sure if that was the case)

jzimbel-mbta (approved):
I think this looks good with my currently limited understanding 😁
Just 2 optional suggestions.

apps/state_mediator/test/state_mediator/integration/gtfs_test.exs
meagharty: This list is the shapes in direction 0, which have varied in number over time (from 2 <-> 3 <-> 4) for this route. So at least 2 (including the first in the list being the expected one).

jzimbel-mbta: Oh I see that it's two separate lists now, somehow missed that on the first read-through. Thanks, this makes sense.

PR #802, opened 2024-05-13
chore: Stop list overrides for CR Spring 2024 rating

jfabi:
#### Summary of changes

**Asana ticket:** [[Extra] 🐛 🟣 Show Spring CR schedules on all routes](https://app.asana.com/0/584764604969369/1207303417443267/f)

Corrects timetable treatment of various stops along the Franklin/Foxboro Line and Providence/Stoughton Line in advance of the Spring 2024 Commuter Rail rating.


apps/state/config/config.exs
```diff
@@ -259,14 +259,35 @@ config :state, :stops_on_route,
 config :state, :stops_on_route,
   stop_order_overrides: %{
     {"CR-Franklin", 0} => [
-      ["Norwood Central", "Windsor Gardens", "Plimptonville", "Walpole"],
```
meagharty, line 262: It looks like there's a stops on route test failing around Plimptonville 🤔 
https://github.com/mbta/api/actions/runs/9071064580/job/24924071203#step:9:1205

jfabi: Oh interesting...I just removed that because we've removed those human-readable `stop_id`s from production years ago... Maybe I'll just adjust the test to use the IRL `stop_id`s.

meagharty (approved):
This looks good to go.

Do you mind posting in #dotcom describing the problems that this addressed? There's some tinkering going on with the timetables on their end, we've been trying to keep them up to date on fixes we have to do. It isn't especially clear to me from the PR itself what issues this fixes (issues from before the new rating, just after the new rating).

PR #804, opened 2024-05-15
chore: add swapscott shuttles to stops on route

Whoops:
#### Summary of changes

**Asana Ticket:** [🚧 🟣 Revised Swampscott shuttle schedule May 18–19](https://app.asana.com/0/1204473309455397/1207301628733435/f)


apps/parse/lib/parse/alerts.ex
```diff
@@ -75,6 +76,11 @@ defmodule Parse.Alerts do
     copy(cause)
   end
 
+  defp cause(alert) do
+    Logger.error("No cause found in alert: #{inspect(alert)}")
```
Whoops, line 80: Naturally, now that I have added this, all the alerts are parsing. Still, I think it's a good guard instead of just failing.

meagharty: Do we want to serve alerts with empty string causes? With the existing setup for alerts though I don't see enforcing this contract as the API's job necessarily. So 👍

meagharty (approved):
Well, we did know the API health checks were complex

PR #807, opened 2024-05-21
fix: depend on proper directly

Whoops:
#### Summary of changes

**Asana Ticket:** HOTFIX


jzimbel-mbta (approved):
Maybe create a separate PR with a revert commit so we don't forget to do that

PR #814, opened 2024-06-04
chore: Erlang 26.2.5

bklebe:
#### Summary of changes

### DIFF
--- a/apps/state/test/state/server_test.exs
+++ b/apps/state/test/state/server_test.exs
@@ -251,7 +251,14 @@ defmodule State.ServerTest do
         %Example{id: 2, data: :other}
       ])
 
-      assert [%{data: 38}, %{data: 44}, %{data: :other}] = HooksServer.all()
+      # ETS tables appear to have different internal ordering on Mac and Linux as of OTP 26, so
+      # explicitly sorting here is necessary to fix this test.
+      assert [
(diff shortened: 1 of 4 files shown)


apps/state/test/state/server_test.exs
```diff
@@ -251,7 +251,7 @@ defmodule State.ServerTest do
         %Example{id: 2, data: :other}
       ])
 
-      assert [%{data: 38}, %{data: 44}, %{data: :other}] = HooksServer.all()
+      assert [%{data: :other}, %{data: 38}, %{data: 44}] = HooksServer.all()
```
bklebe, line 254: @paulswartz got a bit of a stumper here. With this update, my Mac wants this list to be in this order, while the Linux GHA runner wants it to be in the old order. I did some digging around in the OTP repo and couldn't find anything obvious that introduced this, but it seems like it might be a bug? Closed thing I found was this, which suggests the behavior observed on Mac is a regression: https://github.com/erlang/otp/pull/6752.

paulswartz:
@bklebe given the size of this PR, is it possible to split it up into an Elixir upgrade (mostly formatting changes I think?) and an Erlang upgrade?

bklebe:
@paulswartz definitely!

cmaddox5:
Updated in #842.

PR #828, opened 2024-08-07
Meag/add duration uncertainty

meagharty:
#### Summary of changes

**Asana Ticket:** [🍎 Surface 'duration_certainty' in v3 API](https://app.asana.com/0/295455480314405/1207944069300404/f)

image

### DIFF
--- a/apps/api_web/lib/api_web/controllers/alert_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/alert_controller.ex
@@ -298,6 +298,7 @@ defmodule ApiWeb.AlertController do
 
     cause_enum = Model.Alert.cause_enum()
     effect_enum = Model.Alert.effect_enum()
+    duration_certainty_enum = Model.Alert.duration_certainty_enum()
 
     %{
       Activity:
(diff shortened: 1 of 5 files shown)


fsaid90:
> https://transit-alerts-documentation.s3.amazonaws.com/IBI_TRANSIT-alerts_GTFS-realtime_Documentation.pdf says there's also an ESTIMATED value. Is that something we need to handle?

Taking a look at this - from previous testing, `ESTIMATED` didn't show up for the example IBI gave in the doc ("later today" translated to `UNKOWN` instead). I'll do some more testing here and get back to you both.

fsaid90:
> > https://transit-alerts-documentation.s3.amazonaws.com/IBI_TRANSIT-alerts_GTFS-realtime_Documentation.pdf says there's also an ESTIMATED value. Is that something we need to handle?
> 
> Taking a look at this - from previous testing, `ESTIMATED` didn't show up for the example IBI gave in the doc ("later today" translated to `UNKNOWN` instead). I'll do some more testing here and get back to you both.

Never mind - showed up as `ESTIMATED` in my testing. I'll update the ticket + the Swagger text.

Sorry about the confusion!!

PR #833, opened 2024-09-30
fix: prediction ids should be unique by route, for multi-route trips

meagharty:
#### Summary of changes

**Asana Ticket:** [🍎 Make prediction_ids for multi-route trips unique](https://app.asana.com/0/295455480314405/1208355953149037/f)

### DIFF
--- a/apps/api_web/test/api_web/controllers/trip_controller_test.exs
+++ b/apps/api_web/test/api_web/controllers/trip_controller_test.exs
@@ -296,7 +296,7 @@ defmodule ApiWeb.TripControllerTest do
                "id" => "trip",
                "relationships" => %{
                  "predictions" => %{
-                   "data" => [%{"id" => "prediction-trip-stop-", "type" => "prediction"}]
+                   "data" => [%{"id" => "prediction-trip-stop--", "type" => "prediction"}]
                  }
                }
              } = json_response(conn, 200)["data"]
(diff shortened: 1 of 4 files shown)


apps/api_web/test/api_web/controllers/trip_controller_test.exs
```diff
@@ -296,7 +296,7 @@ defmodule ApiWeb.TripControllerTest do
                "id" => "trip",
                "relationships" => %{
                  "predictions" => %{
-                   "data" => [%{"id" => "prediction-trip-stop-", "type" => "prediction"}]
+                   "data" => [%{"id" => "prediction-trip-stop--", "type" => "prediction"}]
```
jzimbel-mbta, line 299: nit: Would it make sense to add stop_sequence and route_id values to the Prediction struct created on line 275, so that they can show up in this string?

jzimbel-mbta (approved):
Looks good, just a non-blocking suggestion.

apps/api_web/test/api_web/controllers/trip_controller_test.exs
meagharty: I tried updating this test this morning. Various tests that rely on the State servers for trip (in the same file like line 59 or line 238, and in other files too, like api_view_helpers_test.exs line 37) started failing intermittently. This indicates an issue that's a test only, where the data leaks between tests (as State server data is expected to be shared between controllers in prod).

PR #841, opened 2024-12-16
chore: Fix flaky tests

cmaddox5:
#### Summary of changes

**Asana Ticket:** [[🍎💳] Fix intermittent API test failures](https://app.asana.com/0/584764604969369/1208839347939907/f)

~~I believe the root cause of the flaky tests is the fact that we reuse `trip_id`s across test modules. This causes some logic to include the wrong model in the returned map. In each test module, I changed the `trip_id`s to be unique to the module.~~

Even after the above change, tests were still intermittently failing. Decided to explore ways to ensure fresh state before each test. Found that stopping and starting the State app before each test allows for more predictable state in each process.

### DIFF
--- a/apps/api_web/test/api_web/controllers/alert_controller_test.exs
+++ b/apps/api_web/test/api_web/controllers/alert_controller_test.exs
@@ -50,9 +50,6 @@ defmodule ApiWeb.AlertControllerTest do
 
   setup tags do
     State.Stop.new_state([%Model.Stop{}])
-    State.Trip.new_state([])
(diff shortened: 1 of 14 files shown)


Whoops:
Here's another that includes some failures on our authentication plug
[test.txt](https://github.com/user-attachments/files/18156757/test.txt)
(To be clear if the intent is to do multiple PRs to clean this up then I'm all for going ahead and merging this one)

lemald (approved):

PR #854, opened 2025-01-22
fix(vehicles): add logging if a non-shuttle vehicle has a null direction_id

rudiejd:
**Asana Ticket:** [🍎🐛 v3 API returning null direction_id in certain cases](https://app.asana.com/0/584764604969369/1209119722014888)

Problem: As described in the linked ticket, sometimes we return vehicles with `"direction_id": null`. We aren't sure why. I spent a couple of days looking into this issue and didn't make any progress.

Solution: We are adding logs for when an invalid `direction_id` is returned. Hopefully these can allow us to better ascertain what conditions lead to this happening.

### DIFF
--- a/apps/state/lib/state/vehicle.ex
+++ b/apps/state/lib/state/vehicle.ex
@@ -49,8 +49,7 @@ defmodule State.Vehicle do
   defp has_invalid_dir(vehicle) do
     not_shuttle = vehicle.route_id != nil and not String.contains?(vehicle.route_id, "Shuttle")
 
-    invalid_dir =
-      vehicle.direction_id == nil or vehicle.direction_id > 1 or vehicle.direction_id < 0
+    invalid_dir = vehicle.direction_id not in [0, 1]
 
     not_shuttle and invalid_dir
   end
--- a/apps/state/lib/state/vehicle.ex
+++ b/apps/state/lib/state/vehicle.ex
@@ -38,7 +38,21 @@ defmodule State.Vehicle do
 
   @impl State.Server
   def pre_insert_hook(vehicle) do
+    if has_invalid_dir(vehicle) do
+      Logger.error("Found vehicle with invalid direction: #{inspect(vehicle)}")
+    end
+
     update_effective_route_id(vehicle)
+  end
+
+  @spec has_invalid_dir(Vehicle.t()) :: boolean()
+  defp has_invalid_dir(vehicle) do
+    not_shuttle = not String.contains?(vehicle.route_id, "Shuttle")
+
+    invalid_dir =
+      vehicle.direction_id == nil or vehicle.direction_id > 1 or vehicle.direction_id < 0
+
+    not_shuttle and invalid_dir
   end
 
   defp update_effective_route_id(%Vehicle{trip_id: trip_id} = vehicle) do
--- a/apps/state/lib/state/vehicle.ex
+++ b/apps/state/lib/state/vehicle.ex
@@ -47,7 +47,7 @@ defmodule State.Vehicle do
 
   @spec has_invalid_dir(Vehicle.t()) :: boolean()
   defp has_invalid_dir(vehicle) do
-    not_shuttle = not String.contains?(vehicle.route_id, "Shuttle")
+    not_shuttle = vehicle.route_id != nil and not String.contains?(vehicle.route_id, "Shuttle")
 
     invalid_dir =
       vehicle.direction_id == nil or vehicle.direction_id > 1 or vehicle.direction_id < 0
--- a/apps/state/lib/state/vehicle.ex
+++ b/apps/state/lib/state/vehicle.ex
@@ -39,7 +39,7 @@ defmodule State.Vehicle do
   @impl State.Server
   def pre_insert_hook(vehicle) do
     if has_invalid_dir(vehicle) do
-      Logger.error("Found vehicle with invalid direction: #{inspect(vehicle)}")
+      Logger.warning("Found vehicle with invalid direction: #{inspect(vehicle)}")
     end
 
     update_effective_route_id(vehicle)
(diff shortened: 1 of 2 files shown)


apps/state/lib/state/vehicle.ex
```diff
@@ -38,9 +38,23 @@ defmodule State.Vehicle do
...
+    if has_invalid_dir(vehicle) do
+      Logger.error("Found vehicle with invalid direction: #{inspect(vehicle)}")
+    end
+
     update_effective_route_id(vehicle)
   end
 
+  @spec has_invalid_dir(Vehicle.t()) :: boolean()
+  defp has_invalid_dir(vehicle) do
+    not_shuttle = vehicle.route_id != nil and not String.contains?(vehicle.route_id, "Shuttle")
```
lemald, line 50: **suggestion:** would this logic work? I would assume that `starts_with?/2` would be ever so slightly cheaper than `contains?/2`.

```suggestion
    not_shuttle = vehicle.route_id != nil and not String.starts_with?(vehicle.route_id, "Shuttle-")
```

apps/state/lib/state/vehicle.ex
```diff
@@ -38,9 +38,23 @@ defmodule State.Vehicle do
...
+
+    invalid_dir =
+      vehicle.direction_id == nil or vehicle.direction_id > 1 or vehicle.direction_id < 0
```
lemald, line 53: **suggestion:** instead of numerical comparisons on `direction_id`, how about something like
```suggestion
    invalid_dir = vehicle.direction_id not in [0, 1]
```

lemald:
**question:** would it be possible to add a quick unit test for this?

apps/state/lib/state/vehicle.ex
rudiejd: That works for me. I wasn't sure if it would always be a prefix.

rudiejd:
@lemald yep, I can add one. I wasn't sure whether we wanted to test logging logic

lemald:
> @lemald yep, I can add one. I wasn't sure whether we wanted to test logging logic

It's definitely not our top priority for automated testing but I personally like doing it where feasible, especially for something like this where the situation isn't regularly occurring for us to test the logging manually.

apps/state/lib/state/vehicle.ex
lemald: I'm _pretty_ sure it's always a prefix. If you prefer we can just leave it as-is since I don't feel confident enough to quite say 100%.

apps/state/lib/state/vehicle.ex
```diff
@@ -38,9 +38,22 @@ defmodule State.Vehicle do
 
   @impl State.Server
   def pre_insert_hook(vehicle) do
+    if has_invalid_dir(vehicle) do
+      Logger.error("Found vehicle with invalid direction: #{inspect(vehicle)}")
```
lemald, line 42: **suggestion:** actually, I didn't think of this before, but I think `warning` might be a more appropriate log level to use here.

rudiejd: can you elaborate as to why? My thinking here was that a non-shuttle vehicle with a route assigned but no `direction_id` could cause immediate problems for downstream consumers, especially since that behavior isn't specified in the API docs

apps/state/test/state/vehicle_test.exs
```diff
@@ -114,4 +115,29 @@ defmodule State.VehicleTest do
...
+    assert %Vehicle{id: "43"} = by_id("43")
+    assert log =~ "Found vehicle with invalid direction"
+    assert log =~ "43"
+
+    # it's a shuttle, so we don't care
+    vehicle = %Vehicle{id: "44", trip_id: "2", route_id: "Shuttle-ToTheMoon"}
+    log = capture_log(fn -> new_state([vehicle]) end)
+
+    assert %Vehicle{id: "44"} = by_id("44")
+    assert log == ""
```
lemald, line 141: **suggestion (non-blocking):** asserting empty logs can be a bit of a pitfall when other unrelated logging gets added later (or even when tests run in parallel). This assertion (or rather refutation) might be more robust.
```suggestion
    refute log =~ "Found vehicle with invalid direction"
```

rudiejd: oh nice, this works with `refute`! I was trying to do the same thing with `assert not log =~ "Found vehicle..." but that threw an error when comparing with `""`

apps/state/lib/state/vehicle.ex
lemald: While it is problematic and out of step with the documentation, I don't think it would actually cause any problems for the functioning of the API application itself (and from what we've heard it probably wouldn't cause truly critical issues downstream).

apps/state/lib/state/vehicle.ex
rudiejd: I did a quick check of the shuttles in `gtfs_creator` and the canoncial routes in `routes.txt`, and all of them start with `Shuttle-`. `gtfs_creator` actually uses the same check (prefix of `Shuttle-`) [to see if a route is a shuttle,](https://github.com/mbta/gtfs_creator/blob/979418b5f1f0abdc5190af7efe613ccc3393f5d0/gtfs_creator2/validators/service_on_dates.py#L195) so I think we should be in the clear. It looks like there were some shuttles called `CR-Shuttle` based on that code, but there are no canonical routes without the `Shuttle-` prefix and other parts of the `gtfs_creator` code rely on that prefix.

apps/state/lib/state/vehicle.ex
rudiejd: I was thinking of downstream as in third party applications - I thought it might crash the transitmatters train tracker :joy: but fair that it doesn't actually matter for the performance of the API itself, and I'm happy to change this if you feel strongly about it

lemald: Yeah, for lack of a better way of putting it I think `warning` would be more in line with the vibes of how TID uses log levels, if changing it is all right with you.

lemald (approved):
👍

PR #855, opened 2025-01-23
fix(dynamo): handle dyamo crash

rudiejd:
#### Summary of changes

**Asana Ticket:** N/A

Handle dynamo crash when scanning records

### DIFF
--- a/apps/api_accounts/lib/api_accounts/dynamo.ex
+++ b/apps/api_accounts/lib/api_accounts/dynamo.ex
@@ -372,6 +372,7 @@ defmodule ApiAccounts.Dynamo do
     |> Enum.flat_map(fn
       %Task{} = task -> Task.await(task)
       results when is_list(results) -> results
+      _ -> []
     end)
   end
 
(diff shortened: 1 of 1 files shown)


apps/api_accounts/lib/api_accounts/dynamo.ex
```diff
@@ -372,6 +372,7 @@ defmodule ApiAccounts.Dynamo do
     |> Enum.flat_map(fn
```
jzimbel-mbta, line 372: nit:
I don't really like that neither this nor any of the other calls to `do_paginated_query` handle its potential error tuple return value, but it doesn't make a huge difference--the tuple will be surfaced when `Enum.flat_map` raises on it:
```
** (Protocol.UndefinedError) protocol Enumerable not implemented for {:error, "foo"} of type Tuple
<stacktrace snipped>
```
so I guess it's fine.
Would prefer to have `do_paginated_query` just `raise` directly if all of the call sites will end up raising anyway.

jzimbel-mbta (approved):

PR #856, opened 2025-01-30
Update cause enum for alert

ErinLMoore:
#### Summary of changes

**Asana Ticket:** [Update cause list in API repo](https://app.asana.com/0/1204819229687089/1209281257904883/f)

The hard-coded Cause enum used in alerts is outdated, this updates it to be the same as the enum we use in AlertsUI.

Feel free to assign this to a different reviewer as needed.


rudiejd:
@ErinLMoore if these are going to change, could you please update the GTFS documentation too? https://github.com/google/transit/blob/master/gtfs-realtime/spec/en/reference.md#enum-cause

Whoops (approved):

ErinLMoore:
> Oops, I meant to link the [MBTA GTFS docs](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs-realtime.md#enhanced-fields) for alerts. If the values are only changing in the API and not GTFS, it's fine not to update that. But in that case, we should still have a list of all the potential causes somewhere, since the API documentation links to the GTFS-RT Alert entity, but the causes in this enum are not listed in that specification

These values come from the GTFS-RT enhanced field called "cause detail", which gets called "cause" in the API, and the actual "cause" field is dropped. In the GTFS, cause detail isn't defined as an enum, it's just a string. That's the reason none of these values have been in the GTFS docs. If you think it's a good idea, I can update the API docs to reflect that.

PR #858, opened 2025-01-30
feat(commuter_rail_occupancies): read commuter rail occupancies from S3 instead of firebase

rudiejd:
…s from S3

#### Summary of changes

**Asana Ticket:** [🍎 API reads Commuter Rail Occupancy from S3 instead of Firebase](https://app.asana.com/0/inbox/1208877353871379/1209182129817552/1209276288985685/f)

Problem:
Keolis needs to decommission their Firebase feed

Solution:
We're having them move the contents of that Firebase feed to S3.

Tactically, I added `StateMediator` for S3 that fetches files via `ExAws`. I made a temporary change to handle a new format (for some reason `cTrainNo` is relabeled as `Trip Name`, but I'm going to hold off on making that more permanent until I hear back from Keolis about whether this change was intentional.

### DIFF
--- a/apps/parse/lib/parse/commuter_rail_occupancies.ex
+++ b/apps/parse/lib/parse/commuter_rail_occupancies.ex
@@ -32,7 +32,6 @@ defmodule Parse.CommuterRailOccupancies do
     with {:ok, flag} <- density_flag(flag),
          {:ok, percentage} <- percentage(density),
          {:ok, name} <- trip_name(train) do
-
       [
         %Model.CommuterRailOccupancy{
           percentage: percentage,
@@ -62,7 +61,9 @@ defmodule Parse.CommuterRailOccupancies do
         status: flag,
         trip_name: name
       }
+
       Logger.error("OK we parsed #{inspect(model)}")
+
       [
         %Model.CommuterRailOccupancy{
           percentage: percentage,
--- a/apps/parse/lib/parse/commuter_rail_occupancies.ex
+++ b/apps/parse/lib/parse/commuter_rail_occupancies.ex
@@ -13,6 +13,9 @@ defmodule Parse.CommuterRailOccupancies do
       {:ok, %{"data" => data}} when is_list(data) ->
         Enum.flat_map(data, &parse_record/1)
 
+      {:ok, data} when is_list(data) ->
+        Enum.flat_map(data, &parse_record/1)
+
       e ->
         Logger.warning("#{__MODULE__} decode_error e=#{inspect(e)}")
         []
@@ -29,6 +32,37 @@ defmodule Parse.CommuterRailOccupancies do
     with {:ok, flag} <- density_flag(flag),
          {:ok, percentage} <- percentage(density),
          {:ok, name} <- trip_name(train) do
+
+      [
+        %Model.CommuterRailOccupancy{
+          percentage: percentage,
+          status: flag,
+          trip_name: name
+        }
+      ]
+    else
+      error ->
+        Logger.warning("#{__MODULE__} parse_error error=#{inspect(error)} #{inspect(record)}")
+        []
+    end
+  end
+
+  defp parse_record(
+         %{
+           "MedianDensity" => density,
+           "MedianDensityFlag" => flag,
+           "Trip Name" => train
+         } = record
+       ) do
+    with {:ok, flag} <- density_flag(flag),
+         {:ok, percentage} <- percentage(density),
+         {:ok, name} <- trip_name(train) do
+      model = %Model.CommuterRailOccupancy{
+        percentage: percentage,
+        status: flag,
+        trip_name: name
+      }
+      Logger.error("OK we parsed #{inspect(model)}")
       [
         %Model.CommuterRailOccupancy{
           percentage: percentage,
--- a/apps/parse/lib/parse/commuter_rail_occupancies.ex
+++ b/apps/parse/lib/parse/commuter_rail_occupancies.ex
@@ -45,7 +45,7 @@ defmodule Parse.CommuterRailOccupancies do
         []
     end
   end
-  
+
   # new format that keolis started providing when they switched this data over to S3
   defp parse_record(
          %{
--- a/apps/parse/lib/parse/commuter_rail_occupancies.ex
+++ b/apps/parse/lib/parse/commuter_rail_occupancies.ex
@@ -22,58 +22,43 @@ defmodule Parse.CommuterRailOccupancies do
     end
   end
 
-  defp parse_record(
-         %{
-           "MedianDensity" => density,
-           "MedianDensityFlag" => flag,
-           "cTrainNo" => train
-         } = record
-       ) do
-    with {:ok, flag} <- density_flag(flag),
-         {:ok, percentage} <- percentage(density),
-         {:ok, name} <- trip_name(train) do
-      [
-        %Model.CommuterRailOccupancy{
-          percentage: percentage,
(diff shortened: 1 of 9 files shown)


apps/parse/lib/parse/commuter_rail_occupancies.ex
```diff
@@ -43,6 +46,38 @@ defmodule Parse.CommuterRailOccupancies do
     end
   end
 
+  defp parse_record(
+         %{
+           "MedianDensity" => density,
+           "MedianDensityFlag" => flag,
+           "Trip Name" => train
```
rudiejd, line 53: image

apps/state_mediator/lib/state_mediator.ex
```diff
@@ -7,9 +7,14 @@ defmodule StateMediator do
...
   # for more information on OTP Applications
   def start(_type, _args) do
+    crowding_source = app_value(:commuter_rail_crowding, :source)
+
     children =
       children(Application.get_env(:state_mediator, :start)) ++
-        crowding_children(app_value(:commuter_rail_crowding, :enabled) == "true")
+        crowding_children(
+          app_value(:commuter_rail_crowding, :enabled) == "true",
+          :erlang.binary_to_atom(crowding_source)
```
lemald, line 16: **suggestion:** it's less of an issue here because it's just coming from an environment variable, but in general any input from outside of an application should go through [`String.to_existing_atom/1`](https://hexdocs.pm/elixir/1.18.2/String.html#to_existing_atom/1) to make sure that we're not creating arbitrary atoms, since atoms are never garbage collected.

apps/state_mediator/lib/state_mediator/s3_mediator.ex
```diff
@@ -0,0 +1,115 @@
...
+  @spec init(Keyword.t()) :: {:ok, __MODULE__.t()} | no_return
+  def init(options) do
+    state_module = Keyword.fetch!(options, :state)
+
+    bucket_arn = Keyword.fetch!(options, :bucket_arn)
+    object = Keyword.fetch!(options, :object)
+    sync_timeout = options |> Keyword.get(:sync_timeout, 5000)
+    interval = options |> Keyword.get(:interval, 5000)
+
+    send(self(), :initial)
```
lemald, line 56: **suggestion:** for this sort of pattern, you can instead return something of the form `{:ok, state, {:continue, continue_arg}}` and implement the [`handle_continue/2`](https://hexdocs.pm/elixir/1.18.2/GenServer.html#c:handle_continue/2) callback.

apps/state_mediator/lib/state_mediator/s3_mediator.ex
```diff
@@ -0,0 +1,115 @@
...
+       module: state_module,
+       bucket_arn: bucket_arn,
+       object: object,
+       sync_timeout: sync_timeout
+     }}
+  end
+
+  @spec handle_info(:initial | :timeout, t) :: {:noreply, t} | {:noreply, t, :hibernate}
+  def handle_info(:initial, %{module: state_module} = state) do
+    _ = Logger.debug(fn -> "#{__MODULE__} #{state_module} initial sync starting" end)
```
lemald, line 70: **question:** do we still need this and the other debug-level logging?

config/runtime.exs
```diff
@@ -28,7 +28,9 @@ if is_prod? and is_release? do
       port: "DYNAMO_PORT" |> System.get_env("443") |> String.to_integer(),
       scheme: System.get_env("DYNAMO_SCHEME", "https://"),
       host: System.fetch_env!("DYNAMO_HOST")
-    ]
+    ],
+    json_codec: Jason
```
lemald, line 32: **question:** is this necessary? [The documentation](https://hexdocs.pm/ex_aws/readme.html#json-codec-configuration) indicates that Jason is already the default.

rudiejd: ah, I can remove this then. I just noticed [RTR was doing this](https://github.com/mbta/rtr/blob/23eae66cf459c18e838be5b027c83acb28b4a538/config/config.exs#L30) for its S3 requests and copied from there

apps/state_mediator/lib/state_mediator/s3_mediator.ex
lemald: Related question: if the logging isn't necessary, can the two clauses of `handle_info/2` be combined?

lemald: One last related question: if there's just one `handle_info/2` clause handling everything, could you make `handle_response/2` private and instead just test `handle_info/2` directly?

apps/state_mediator/lib/state_mediator.ex
rudiejd: gotcha. honestly I can probably just use the string/binary value. I thought having the value be a constant could be nice, but seems like we're going to have to convert from a string to an atom anyway

config/runtime.exs
lemald: Oh, huh - maybe it wasn't the default in an older version of ex_aws?

apps/state_mediator/lib/state_mediator/s3_mediator.ex
rudiejd: that seems to be semantically closer to what is actually going on. I just copied this from the other `StateMediator` modules, which use a similar pattern with a timeout:  https://github.com/mbta/api/blob/3cb32a593feb017df98e9783a6de77c210567384/apps/state_mediator/lib/state_mediator/mediator.ex#L84

lemald: I know the `handle_continue/2` callback was added just a few years ago and wasn't present in some older versions of Elixir, so it may have just not existed when other parts of the API were written.

apps/state_mediator/lib/state_mediator/s3_mediator.ex
rudiejd: for the debug logging: this is how all of our other `StateMediator` modules work, so I just did this for parity sake. I ended up switching the `handle_info(:initial, ...)` to a `handle_continue`.

mix.exs
```diff
@@ -63,7 +63,8 @@ defmodule ApiUmbrella.Mixfile do
       {:lcov_ex, "~> 0.2", only: [:dev, :test], runtime: false},
       {:mix_audit, "~> 2.0", only: [:dev], runtime: false},
       # Generate docs with `mix docs`
-      {:ex_doc, "~> 0.20", only: [:dev, :test], runtime: false}
+      {:ex_doc, "~> 0.20", only: [:dev, :test], runtime: false},
+      {:mox, "~> 1.0", only: :test}
```
lemald, line 67: **questions:** A couple of things:

1. What's the rationale for this being needed at both the root level and also in the state mediator app? Sorry, I don't actually work with umbrella apps very often.
1. Was there a corresponding `mix.lock` change that should also be committed?

rudiejd: 1. this was an accident, and I forgot to revert. We only need `mox` to be added in the `state_mediator` project.
2. there is only one `mix.lock` file for the entire project, and `mox` is already included in that file. each sub-project within this umbrella project defines its lockfile as the root `mix.lock` , e.g. [state_mediator](https://github.com/mbta/api/blob/6150aae4606f4da4ccfbf99ff94db1ed88cbcb54/apps/state_mediator/mix.exs#L14)

lemald (approved):

PR #873, opened 2025-04-15
chore: Update explicit stops for Boat-F6 & F7.

cmaddox5:
#### Summary of changes

**Asana Ticket:** [⛴️ Boat-F6 becomes F6, F7, and F8](https://app.asana.com/1/15492006741476/project/584764604969369/task/1209912077451333?focus=true)

Given the number of stop sequence combos, this code helps dotcom by giving it the exact list they should display in the Schedule Finder. Updated the list for Boat-F6 and added one for Boat-F7.

### DIFF
--- a/apps/state/lib/state/stops_on_route.ex
+++ b/apps/state/lib/state/stops_on_route.ex
@@ -188,14 +188,26 @@ defmodule State.StopsOnRoute do
 
   defp gather_explicit_overrides(%{id: "Boat-F6"}, 1),
     do: [
-      {"Boat-F6", 1, :all, "Boat-F6-Wdy-Smr-23", true,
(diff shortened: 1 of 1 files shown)


cmaddox5:
Sorry, was a little premature declaring this the solution. I noticed that the origin stops for the new Boat-F7 route look fine, so I'm wondering if the better solution is to actually through this override logic out. It's on its way to dev-green so will give an update on this soon.

cmaddox5:
Looks like removing the overrides is the better solution here. Schedule Finder looks fine without them:
Screenshot 2025-04-16 at 10 16 27 AM
Screenshot 2025-04-16 at 10 16 20 AM

sloanelybutsurely (approved):
love to see deleted code!

PR #884, opened 2025-07-08
chore: connect Winthrop landing and terminal

Whoops:
#### Summary of changes

**Asana Ticket:** [🍎 Add connection between Winthrop Ferry Terminal + Winthrop Landing](https://app.asana.com/1/15492006741476/project/584764604969369/task/1210690613924915?focus=true)

### DIFF
--- a/apps/state/lib/state/connecting_stops.ex
+++ b/apps/state/lib/state/connecting_stops.ex
@@ -34,7 +34,8 @@ defmodule State.ConnectingStops do
     "place-harsq" => %{add: ~w(110)},
     "place-aqucl" => %{add: ~w(Boat-Long Boat-Long-South-4 Boat-Aquarium)},
     "Boat-Charlestown" => %{add: ~w(12859 12856)},
-    "place-ER-0115" => %{add: ~w(ER-0117-01 ER-0117-02)}
+    "place-ER-0115" => %{add: ~w(ER-0117-01 ER-0117-02)},
+    "Boat-Winthrop" => %{add: ~w(129863)}
   }
 
   def start_link(_opts \\ []) do


firestack (approved):
Because ~I don't know what I'm doing, I double checked the stop ids just to be sure it made sense 😅 
- https://www.mbta.com/stops/129863
- https://www.mbta.com/stops/Boat-Winthrop

LGTM 👍🏻

---

I am left with a question, that doesn't block merging but I'll put here anyway; is there something in GTFS-Schedule that would model this relationship as well? I can't find anything, and I don't see anything that would point to modeling this _that is applicable in this scenario_ (i.e., pathways.txt and transfers.txt don't really apply here)

Whoops:
> Because ~I don't know what I'm doing, I double checked the stop ids just to be sure it made sense
If you wanted to check the behavior, you can fetch the stop including the connecting stops and validate it's there. So after deploying to dev-green you can see that 
https://api-dev-green.mbtace.com/stops/Boat-Winthrop?include=connecting_stops
returns Winthrop Terminal as well.

>  is there something in GTFS-Schedule that would model this relationship as well?
Connecting stops for whatever reason is an API-only feature. @digitalcora may have some history as to why that is, versus putting it in a GTFS extension (or setting up transfers), that's before my time, so I don't really have the history. Transfers and pathways are similar, but different ;-)

digitalcora:
This being an API-only feature is my fault, and no, there's no reason that has to be the case — it was just more expedient to implement this way at the time, and there were no GTFS consumers who needed the data.

> is there something in GTFS-Schedule that would model this relationship as well?

No, afaik there is nothing in the standard that allows modeling this specific kind of relationship; it would have to be our own custom thing. Though to atone for my sins I've been working on a proposal for just such a thing (moving the source of this data into GTFS), which I'll share with the team shortly!

PR #885, opened 2025-09-17
feat: parse alert notification timing info

boringcactus:
#### Summary of changes

**Asana Ticket:** [Notifications | Alert consumer plan](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211209894303568?focus=true)

We’re getting ready to work on notifications in MBTA Go, and it’d be nice if we didn’t have to read GTFS directly just to get these fields.

When an alert is marked as closed and the all clear notification goes out, the alert has no active periods. Currently, alerts with no active periods are discarded in the parser, and I don’t know for sure whether or not anything is assuming that closed alerts disappear from the feed instantly and that alerts in the feed always have at least one active period, but I don’t think those are reasonable assumptions to make even if they’ve always been true until now.


digitalcora:
> In other words, this logic _did_ make, and rely on, the assumption that closed alerts disappear from the feed.

I can confirm it doesn't do this currently, and Screenplay should be unaffected by this change; we're filtering alerts based on active period at an earlier step in the process. (In other words our "diff" is between the set of alerts _that are active_ right now and the set that were active a few seconds ago.)

PR #994, opened 2026-04-08
feat(ApiWeb): prohibit filtering only by `direction_id`

runkelcorey:
## Summary of changes

Currently, requests to filter by `direction_id` alone succeed but takes a lot of time and, presumably, lots of resources. If a user wants half the `stop_events`, then they should just go to the feed instead. This PR

1. prohibits `stop_event` queries that only filter by `direction_id`
2. adds a new error view that explains it's not allowed

I toyed with adding a reference to the feed but, since that's not live in prod yet, I'm not sure referencing it is important now. Thoughts?


jadrian (approved):
LGTM. And I agree it's fine to leave out the mention of the feed since it's not yet in prod.

PR #1009, opened 2026-05-11
fix(StopEvent): assign date when including schedule

runkelcorey:
## Summary of changes

**Asana Ticket:** [🐞 Unhandled stop_events query](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1214494536112681)

1. Fixes typo in README
2. Assigns `start_date` as the schedule date when including schedules in `stop_events` queries. Not assigning the date caused this bug
3. Updates `ScheduleView` to format dates using a date field
4. Exposes `service_id` as attribute of schedules; this enables (2) and also seems useful for consumers (although I guess it hadn't been needed thus far)

## How were these changes validated?

1. Updates `StopEventControllerTest` to check that the service ids returned by each schedule match the expected service ids

## What questions should reviewers consider?

1. Are there edge cases where returning the `service_id` could mislead consumers?
2. Is there a cleaner way to assign the date? I feel like `date_seconds` and `date` represent just enough difference that a bigger refactor would be needed but idk


apps/api_web/lib/api_web/views/schedule_view.ex
```diff
@@ -104,17 +106,19 @@ defmodule ApiWeb.ScheduleView do
...
     |> DateHelpers.add_seconds_to_date(seconds)
     |> DateTime.to_iso8601()
   end
 
-  defp format_time(nil, _) do
-    nil
-  end
+  defp format_time(_schedule, nil, _conn), do: nil
+
+  defp get_date_for_schedule(%{date: date}, _conn) when not is_nil(date), do: date
```
jzimbel-mbta: question:
When do we expect `schedule` to have a `:date` field? The `Model.Schedule` struct def doesn't include that field.

(I believe `schedule` here is a `Model.Schedule` struct, but I definitely could be wrong)

apps/api_web/lib/api_web/views/stop_event_view.ex
```diff
@@ -70,12 +66,24 @@ defmodule ApiWeb.StopEventView do
...
 
+  defp attach_schedule(stop_event, schedules) do
+    schedule = Map.get(schedules, {stop_event.trip_id, stop_event.stop_sequence})
+    attach_schedule_with_date(stop_event, schedule)
+  end
+
+  defp attach_schedule_with_date(stop_event, nil), do: Map.put(stop_event, :schedule, nil)
+
+  defp attach_schedule_with_date(stop_event, schedule) do
+    schedule_with_date = Map.put(schedule, :date, stop_event.start_date)
```
jzimbel-mbta: issue:
You're putting a new field on a `Model.Schedule` struct here, which is unconventional. Would it make sense to update the struct definition to include `:date`? Or is there another way you could associate a stop event date with the schedule without having it be a kind of "stowaway" on the struct?

jzimbel-mbta (commented):
I think this looks good overall, but (as you also pointed out in the PR description) I wonder if there's a different way to associate a stop event date with a schedule than `Map.put`-ing it right on the struct.

apps/api_web/lib/api_web/views/stop_event_view.ex
runkelcorey: I think the better answer is to handle this downstream in the StopEventView rather than the only other way I can conceive, which would be to open 2 `conn`s for the purposes of formatting the date

runkelcorey:
> I think this looks good overall, but (as you also pointed out in the PR description) I wonder if there's a different way to associate a stop event date with a schedule than `Map.put`-ing it right on the struct.

Agreed. I moved the logic to associate dates and schedules out of `Schedule` and into `StopEvent`. That's *also* unconventional but respects the fact that stop events from 2 dates may occur

apps/api_web/lib/api_web/views/stop_event_view.ex
```diff
@@ -51,31 +51,64 @@ defmodule ApiWeb.StopEventView do
...
 
-  def preload(stop_event, conn, _opts) do
+  defp attach_schedules_if_needed(stop_event, conn) do
     if split_included?("schedule", conn) do
       schedule = State.Schedule.schedule_for(stop_event)
-      Map.put(stop_event, :schedule, schedule)
+      attach_formatted_schedule(stop_event, schedule)
     else
       stop_event
     end
```
jzimbel-mbta: suggestion:
Would it make sense to just wrap the stop event in a list for this clause? Like:
```suggestion
    [stop_event] = attach_schedules_if_needed([stop_event], conn)
    stop_event
```

(since the logic between the two clauses is otherwise identical)

In fact you could even condense `attach_schedules_if_needed/2` to one clause if you used `List.wrap/1`, since it is a no-op if the value is already a list:
```ex
defp attach_schedules_if_needed(stop_events, conn) do
  stop_events = List.wrap(stop_events)
  # ... rest of the original body
end
```

apps/api_web/lib/api_web/views/stop_event_view.ex
```diff
@@ -51,31 +51,64 @@ defmodule ApiWeb.StopEventView do
...
   end
 
+  defp attach_formatted_schedule(stop_event, schedule) do
+    formatted = if schedule, do: format_schedule_times(schedule, stop_event.start_date)
+    Map.put(stop_event, :schedule, formatted)
```
jzimbel-mbta, line 88: comment (no action needed):
I still don't love tacking extra fields onto a struct, but I reviewed some of our other `JaSerializer` views and it seems to be a common pattern, mainly used to attach associations to an entity that's about to be serialized. Given that, your approach is fine.

jzimbel-mbta (approved):
Looks good, just one non-blocking suggestion.

apps/api_web/lib/api_web/views/stop_event_view.ex
runkelcorey: thank you! I spent 10 minutes looking at this knowing there was an elegant way but I couldn't see it!

runkelcorey: I'll leave it as 2 clauses since I'd need to handle the list somewhere else if it's a single stop event

PR #1013, opened 2026-05-12
feat: update `cause_detail` and `effect_detail` to expect a translated string

allisonking:
#### Summary of changes

**Asana Ticket:** [Make `cause_detail` and `effect_detail` translated strings](https://app.asana.com/1/15492006741476/project/1167196461144361/task/1213724340503424?focus=true)

This is the other side of https://github.com/mbta/alerts_ui/pull/1030 (which was merged, then [reverted](https://github.com/mbta/alerts_ui/pull/1038) when we realized it broke API!) where alerts manager will now set `cause_detail` and `effect_detail` in the feed to be translated strings instead of regular strings. This PR makes sure the API can accept _both_ a string or a translated string for these fields. Once all instances of alerts manager is running and we are fully switched over to translated strings, we can remove the support for regular strings. 

To remove the backwards compatibility, should be able to revert [5ea59df](https://github.com/mbta/api/pull/1013/commits/5ea59df6f0ce626d9641ff51e166a51bcc270d2c)

### DIFF
--- a/apps/parse/lib/parse/alerts.ex
+++ b/apps/parse/lib/parse/alerts.ex
@@ -2744,7 +2744,7 @@ defmodule Parse.Alerts do
   def parse_alert(alert) do
     %Alert{
       id: Map.get(alert, "id"),
-      effect: alert |> Map.get("effect_detail") |> copy,
+      effect: alert |> Map.get("effect_detail") |> translated_text,
       cause: cause(alert),
       header: alert |> Map.get("header_text") |> translated_text,
       short_header: alert |> Map.get("short_header_text") |> translated_text,
@@ -2775,7 +2775,7 @@ defmodule Parse.Alerts do
   end
 
   defp cause(%{"cause_detail" => cause}) do
-    copy(cause)
+    translated_text(cause)
   end
 
   defp cause(%{"cause" => cause}) do
--- a/apps/parse/lib/parse/alerts.ex
+++ b/apps/parse/lib/parse/alerts.ex
@@ -2744,7 +2744,9 @@ defmodule Parse.Alerts do
   def parse_alert(alert) do
     %Alert{
       id: Map.get(alert, "id"),
-      effect: alert |> Map.get("effect_detail") |> translated_text,
+      # Backwards compatability-can change to `translated_text` after all instances of
+      # alerts manager are returning a translated string
+      effect: alert |> Map.get("effect_detail") |> maybe_translated_text,
       cause: cause(alert),
       header: alert |> Map.get("header_text") |> translated_text,
       short_header: alert |> Map.get("short_header_text") |> translated_text,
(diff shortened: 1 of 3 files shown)


apps/parse/lib/parse/alerts.ex
```diff
@@ -2744,7 +2744,9 @@ defmodule Parse.Alerts do
   def parse_alert(alert) do
     %Alert{
       id: Map.get(alert, "id"),
-      effect: alert |> Map.get("effect_detail") |> copy,
+      # Backwards compatability-can change to `translated_text` after all instances of
+      # alerts manager are returning a translated string
+      effect: alert |> Map.get("effect_detail") |> maybe_translated_text,
```
allisonking, line 2749: I chose to make this a new func `maybe_translated_text` instead of altering `translated_text` just in case it had unintended consequences to `translated_text` which is used in many more places

apps/parse/test/parse/alerts_test.exs
```diff
@@ -443,6 +467,43 @@ defmodule Parse.AlertsTest do
...
+      map = %{
+        "timestamp" => "1496832813",
+        "alerts" => [
+          %{
+            "active_period" => [],
+            "alert_lifecycle" => "NEW",
+            "cause" => "UNKNOWN_CAUSE",
+            "cause_detail" => [
+              %{"translation" => %{"language" => "en", "text" => "UNKNOWN_CAUSE"}}
+            ],
```
lemald: **question:** I might be missing something, but for the purposes of this text, shouldn't this key and `"effect_detail"` both be strings rather than ~maps~ lists of maps?

allisonking: you're right, that's pretty unintuitive! I was looking at how the other fields handle translated strings, for example `description_text` and it's this list of maps:

https://github.com/mbta/api/blob/5c2e966eb7cb2e4809a400779a6cf43e00166730/apps/parse/test/parse/alerts_test.exs#L392-L395

here's [the gtfs docs](https://gtfs.org/documentation/realtime/reference/#message-translatedstring) for `Translation`, where it says it has cardinality 'Many' on the key `translation`

allisonking: also, `cause` and `effect` are still just strings, but the spec says `cause_detail` and `effect_detail` should be `TranslatedString`s (not sure if that's what you were asking!)

lemald: I guess what's tripping me up is just that this test is called `"string cause_detail and effect_detail are still accepted"` but `cause_detail` and `effect_detail` aren't actually strings here like they were in the other tests before this change.

allisonking: OH yes you are totally right—I had both cases when I was developing and ended up deleting the wrong one... 🤦 I'll fix that now!

lemald (approved):
Looks good to me!

allisonking:
hmm I just rebased main but now the dev tasks check is failing. any ideas why?

lemald:
> hmm I just rebased main but now the dev tasks check is failing. any ideas why?

It looks like it's complaining about the version of the `decimal` package. I've opened a separate PR (#1014) to upgrade that, then once that's merged you can rebase again and hopefully it should pass.

lemald:
> > hmm I just rebased main but now the dev tasks check is failing. any ideas why?
> 
> It looks like it's complaining about the version of the `decimal` package. I've opened a separate PR (#1014) to upgrade that, then once that's merged you can rebase again and hopefully it should pass.

Okay, I got that other PR merged - you should be good to rebase and hopefully CI will pass!

PR #1018, opened 2026-05-20
feat: 6/13/26 service_date extended to 4:59 am on 6/14

dks-mbta:
#### Summary of changes

**Asana Ticket:** [🍎🐛 API should return late-night service when schedules endpoint called with no service date](https://app.asana.com/1/15492006741476/project/584764604969369/task/1214240195251422?focus=true)

6/13/26 service_date extended to 4:59 am on 6/14


rudiejd (approved):
I couldn't really figure out a satisfying way to test this in a deployed environment since I can't change the day. but this does look to be the way that 3 AM service works

PR #1036, opened 2026-06-12
fix(otp): upgrade OTP patch version to allow us to build

rudiejd:
#### Summary of changes

**Asana Ticket:** ad-hoc

Problem:
Builds were failing with an error about the TLS certificate. See https://mbta.slack.com/archives/GP5QJM98A/p1781291931302989

Solution:
Upgrade OTP to the newest patch version of OTP 27. This seems to fix the certificate errors.


lemald:
@rudiejd you should probably also upgrade the version the real application uses (via the image in the `Dockerfile`) to match.

lemald (approved):

PR #1042, opened 2026-06-29
feat(alerts): Remove logic for adding parent informed entities.

allisonking:
#### Summary of changes

**Asana Ticket:** [🧑‍🧒 Publish additional informed_entities for parent stops](https://app.asana.com/1/15492006741476/project/1167196461144361/task/1214457338889814?focus=true)

This is the other half of https://github.com/mbta/alerts_ui/pull/1051. Alerts manager now adds parent entities itself, so we can remove the API logic. They've been running concurrently up until now, but that was causing [problems](https://mbta.slack.com/archives/C03K6NLKKD1/p1782489011275529) because we were getting duplicate parent entities in cases where the activities differed.

I recreated the blue line shuttle issue in staging [here](https://alerts-manager-staging.mbtace.com/alerts/edit/1003231)

To confirm this fixes the duplicate parent entities issue:

1. Point your local API to the staging alerts manager feed by editing 

https://github.com/mbta/api/blob/b5368824a32e0f846e03e79d834172058cc934bc/apps/state_mediator/config/config.exs#L25 

to https://mbta-alerts-ui.s3.amazonaws.com/staging/alerts_feed_enhanced.json

2. Visit localhost:4000/alerts
3. Confirm that there are only 2 instances of `place-wondl` (one for each direction) and not 4 for alert `1003231`


apps/state/test/state/alert/hooks_test.exs
```diff
@@ -238,315 +165,5 @@ defmodule State.Alert.HooksTest do
...
-                   route: "main-route",
-                   direction_id: 0,
-                   trip: "trip1",
-                   activities: ["BOARD", "USING_WHEELCHAIR"]
-                 }
-               ]
-             } = apply_hook(informed_entities)
-    end
-
-    test "handles specific case from bugfix ticket" do
```
allisonking, line 348: I did recreate this alert to make sure alerts manager is handling this test case right and I think it is, but wouldn't mind a second pair of eyes. Here's the alert in staging: https://alerts-manager-staging.mbtace.com/alerts/edit/1003230

And here's the equivalent feed: https://mbta-alerts-ui.s3.amazonaws.com/staging/alerts_feed_enhanced.json

Search 1003230, and there should be 12 informed entities, of which 6 are parent entities, and 6 are the original child ones. The 6 original ones match what in this test is in `alerts_enhanced_json_excerpt`, and the 6 parent entities match what is in `expected_new_informed_entities` 

(actually, the parent entities' activities won't match exactly, but this is a good thing—the alerts manager is a bit stricter about which activities to associate with each parent entity. for example, the activities at North Station and Oak Grove won't be all 3 activities, it'll depend on direction, since these are the beginning/end of the suspension)

jzimbel-mbta:
Hi @allisonking, thanks for the PR!

Your changes look good, I've just been slow to review because I was working on addressing the issues that `mix deps.audit` was raising about our `hackney` dep version.

I'll have a fix done for this shortly, which should stop CI from failing on this and all the other branches.

jzimbel-mbta (approved):

allisonking:
Thanks for the help @jzimbel-mbta ! we realized that it would make more sense to batch this together with a change we are currently working on to remove the alternate trip bit as well, that way we can delete the entire hooks file and monitor the deploy in one go. So I'll close this PR but open another one shortly!

PR #1045, opened 2026-07-06
Remove alerts hook

allisonking:
#### Summary of changes

**Asana Ticket:** [☄️ Create additional informed entities for multi-route trips](https://app.asana.com/1/15492006741476/project/1167196461144361/task/1214283628700415?focus=true) and [🧑‍🧒 Publish additional informed_entities for parent stops](https://app.asana.com/1/15492006741476/project/1167196461144361/task/1214457338889814?focus=true)

Alerts manager now handles adding informed entities for both parent entities (https://github.com/mbta/alerts_ui/pull/1051) and alternate routes (https://github.com/mbta/alerts_ui/pull/1209). This takes out the hook so that API is no longer responsible for this behavior.

Some more context: previously I had removed parent entities only in https://github.com/mbta/api/pull/1042, but we decided we should just remove both parent and alternate entities in one go. It is mostly okay for the API to still be trying to add these, it'll just result in occasional duplicates like in [this instance](https://mbta.slack.com/archives/C03K6NLKKD1/p1782489011275529). I don't think alternate entities would ever be duplicated... 🤞 Alerts adding parents has been in prod for about a month now, and we just deployed alerts adding alternate routes to staging, so `api-dev-blue` is handling it there. 

We need to trigger all the alerts to rebuild their informed entities with their new logic before this API code can safely go away. So I won't be merging this quite yet, but wanted it to be ready!


apps/state/test/state/alert_test.exs
```diff
@@ -178,7 +178,13 @@ defmodule State.AlertTest do
...
       alert =
         put_in(@alert.informed_entity, [
-          %{activities: ["BOARD"], trip: "trip"}
+          %{
+            activities: ["BOARD"],
+            trip: trip,
+            route: route.id,
+            route_type: route.type,
+            direction_id: 1
+          }
```
allisonking, line 181: tbh I don't totally understand why I needed to make this change for the test to pass. I'm guessing the hook was adding this data in from the "trip" connection. but now we should get the full informed entity from alerts, so I put the whole informed entity here and that seemed to make the test happy

rudiejd: I think this was a result of the `include_alternate_trip_entities/1` function. you probably don't need to test that alternate trips are included anymore, right?

rudiejd:
deploying to dev-blue so we can check it out there!

rudiejd:
not seeing anything suspicious on dev-blue to speak of, so this lgtm!

rudiejd (approved):

PR #1055, opened 2026-07-21
chore: upgrade erlang and elixir versions

zyxw59:
#### Summary of changes

**Asana Ticket:** [💳 Take stock of which TD applications need Elixir / Erlang version upgrades soon and make tickets](https://app.asana.com/1/15492006741476/project/584764604969369/task/1216071585578544?focus=true)

Upgrade erlang version to 28.5.0.3
Upgrade elixir version to 1.19.5-otp-28


zyxw59:
I tried upgrading to erlang 29/elixir 1.20, but I was getting a lot more errors that seemed nontrivial to fix

lemald:
You'll also need to change the versions in the `Dockerfile`.

apps/state/test/state/server_test.exs
```diff
@@ -251,7 +251,7 @@ defmodule State.ServerTest do
         %Example{id: 2, data: :other}
       ])
 
-      assert [%{data: 38}, %{data: 44}, %{data: :other}] = HooksServer.all()
+      assert [%{id: 2, data: :other}, %{id: 1, data: 38}, %{id: 1, data: 44}] = HooksServer.all()
```
lemald: **suggestion (non-blocking):** if the ordering isn't significant, and this change was only necessary because of some arbitrary change to ordering as part of the version upgrade, you could consider splitting this out to a series of assertions that each of the expected elements is `in` the array, along with an assertion on its length. That way the order wouldn't matter.

lemald (approved):
This is looking good to me! Before merging, though, I think you should let this run on dev-blue / dev-green for a bit and make sure there are no new crashes or anything.

PR #1058, opened 2026-07-29
Tagged prod releases

lemald:
#### Summary of changes

**Asana Ticket:** [Set up tag/release-based prod deploys - api](https://app.asana.com/1/15492006741476/project/584764604969369/task/1216046041063091?focus=true)

A couple of notes on this one:

1. The previous deploy workflows were directly calling the relevant actions instead of making use of the `deploy-ecs.yml` workflow from the `workflows` repo. I decided to go ahead and change that.
2. This repo has a regularly-scheduled daily deploy to dev-blue. I wanted to simply schedule that in `deploy-lower-env.yml`, but it sounds like you can't pass settings into `cron`-scheduled workflows. I ended up creating a separate workflow for this that shares the same concurrency group but is only ever triggered via `cron`. I hope that's all right?


lemald:
@jzimbel-mbta looks like @npatel-mbta got auto-assigned, but tagging you specifically because of your familiarity with this stuff!

jzimbel-mbta:
1d15b0fda8295729d2899b209ddf0eb87236e679 is correct, that's another thing I wanted to go back and fix over in rtr. Good catch!

I thought I was being clever back when I wrote this—I wanted to reduce the amount of times `checkout` was called and thought that it could be called once and depended on by other jobs that needed the checked-out code, but each job runs in its own runner / container even if it depends on outputs of another job, so the `checkout` job I added is basically pointless. 😅

.github/workflows/deploy-lower-env.yml
```diff
@@ -0,0 +1,49 @@
+name: Deploy to lower environment (dev etc)
+
+on:
+  workflow_call:
+    inputs:
+      environment:
+        type: string
+        default: dev
+        required: true
```
jzimbel-mbta: This workflow isn't called by another* so the `workflow_call` block can be removed.

\* (though maybe it would make sense for `deploy-dev-blue-scheduled` to call this one?)

jzimbel-mbta (approved):
This looks good overall. I tried coming up with a nicer way to configure the scheduled dev-blue deploys, but I think your approach is the least-janky option available.

.github/workflows/deploy-lower-env.yml
lemald: cd69bc3

I considered your parenthetical note, but decided that `deploy-dev-blue-scheduled.yml` is already so simple that there isn't much point in pointing it to `deploy-lower-env,yml` as far as I can tell.

PR #1064, opened 2026-08-04
feat(state): Enable `:set` tables

runkelcorey:
#### Summary of changes

[🐞 State.StopEvent times out on refresh](https://app.asana.com/1/15492006741476/project/1189492770004753/task/1214717320822000?focus=true)

This PR adds the option of making a `:set` table instead of a `:bag`. For conservatism, I opted for only allowing `:set` and `:bag`.

>[Insert and lookup times in tables of type set are constant, regardless of the table size. For table types bag and duplicate_bag time is proportional to the number of objects with the same key.](https://www.erlang.org/doc/apps/stdlib/ets.html)

After deploying to `dev-blue`, Splunk showed that the `init_table` log time (which approximates the full reload) dropped 27% against `dev-green`, the baseline:

| measure         | dev-green | dev-blue | 𝚫        |
|-----------------|-----------|----------|----------|
| 5th percentile  | 6,442ms   | 5,952ms  | -8%      |
| median          | 13,347ms  | 9,695ms  | **-27%** |
| 95th percentile | 16,336ms  | 14,589ms | -10%     |

This change won't eliminate the issue entirely—9.6s is still much longer than it takes to reload `Prediction`—so I'm working on more PRs to address the issue.

### DIFF
--- a/apps/state/lib/state/server.ex
+++ b/apps/state/lib/state/server.ex
@@ -72,6 +72,12 @@ defmodule State.Server do
   defmacro __using__(opts) do
     indices = Keyword.fetch!(opts, :indices)
     hibernate? = Keyword.get(opts, :hibernate, true)
+    table_type = Keyword.get(opts, :table_type, :bag)
+
+    unless table_type in [:bag, :set] do
+      raise CompileError,
+        description: "invalid :table_type #{inspect(table_type)}, expected :bag or :set"
+    end
 
     recordable =
       case Keyword.fetch!(opts, :recordable) do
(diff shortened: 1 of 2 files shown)


lemald:
Interesting! So were there a lot of `StopEvents` with the same key? That would explain the poor performance with a `:bag` table type, but then wouldn't that imply that there will be a problem with the `:set` table type due to `:set` only allowing one value per key?

runkelcorey:
> Interesting! So were there a lot of `StopEvents` with the same key? That would explain the poor performance with a `:bag` table type, but then wouldn't that imply that there will be a problem with the `:set` table type due to `:set` only allowing one value per key?

No, the `id` is unique and [enforced in `lamp`](https://github.com/mbta/lamp/blob/9c7646b8bed8d2009a1e38f86dff67a12d3bd781/src/lamp_py/flashback/events.py#L13-L19). I _think_ the insert performance is not only proportional to the number of objects with the same key but, by implication, the number of objects. Also, perhaps other performance is also affected by table type?

lemald (approved):
Okay, makes sense. I wish we had a slightly more certain story about what's improving the performance, but since this is a very targeted change that only applies to `StopEvent`s I'm okay with merging it.

PR #1065, opened 2026-08-06
feat(state,parse): Perform partial updates on `StopEvent`

runkelcorey:
**Asana Ticket:** [🐞 State.StopEvent times out on refresh](https://app.asana.com/1/15492006741476/project/1210964140773280/task/1214717320822000?focus=true)

## What changes does this PR propose?

Changes refresh strategy for `StopEvent` from a full load to an incremental update by:

1. Implementing `handle_new_state` callback that filters incoming stop event JSONs for records that are newer than a configurable timestamp (defaulting to 2 hours, the same that Flashback aims for)
2. Implementing `post_commit_hook` callback that evicts records from the mnesia table if they are older than the configurable timestamp

Since `StopEvent` is a `:set` table, incoming values with the same keys will overwrite existing values.

## How were these changes validated?

1. Added tests to `State.StopEvent` and `Parse.StopEvent` that different cases of partial updates
2. [Deployed to dev-blue](https://github.com/mbta/api/actions/runs/31115360507) and [saw dramatic differences between the refresh times in `dev-blue` and `dev-green`](https://mbta.splunkcloud.com/en-US/app/search/search?earliest=1786030981&latest=1786032781&q=search%20index%3Dapi-dev-*%20took%20NOT%20fetching%0A%7C%20rex%20field%3D_raw%20%22(%3F%3Cfunction%3E%5Cw%2B)%20Elixir.State.StopEvent%20(%3F%3Cpid%3E%5CS%2B)%20took%20(%3F%3Cduration_ms%3E%5B%5Cd.%5D%2B)ms%22%0A%7C%20stats%20count%2C%20p05(duration_ms)%20as%20lowest%2C%20median(duration_ms)%20as%20med%2C%20p95(duration_ms)%20as%20highest%20by%20index%2C%20function%0A%7C%20table%20index%2C%20function%2C%20count%2C%20lowest%2C%20med%2C%20highest%0A%7C%20sort%20-med&display.page.search.mode=fast&dispatch.sample_ratio=1&workload_pool=&display.page.search.showFields=0&display.events.fields=%5B%22host%22%2C%22source%22%2C%22sourcetype%22%2C%22validation_errors%22%2C%22error_type%22%2C%22error_records%22%2C%22bucket_name%22%2C%22is_recent%22%2C%22status%22%2C%22process_name%22%2C%22index%22%2C%22duration_ms%22%5D&display.page.search.tab=statistics&display.general.type=statistics&sid=1786032800.25638) (which is running the current, full-load strategy):

| measure         | `dev-green` | `dev-blue` | 𝚫        |
|:----------------|-----------|----------|----------|
| 5th percentile  | 15,681ms   | 11ms  | -99.9%      |
| median          | 17,459ms  | 14ms  | **-99.9%** |
| 95th percentile | 19,957ms  | 27ms | -99.9%     |

These are crazy improvements so I feel like I'm missing a logging statement to capture the full overhead of the partial load. However, we should expect something like this: of the 116,740 events in the stop events dataset I downloaded yesterday, only 356 (0.3%) were updated or inserted the last 30 seconds of the dataset.

distribution of stop events over time

As a sanity check, I also counted the number of times each of these function were executed: since this is an event loop dictated by the time it takes to refresh data + sleep time, the more times something is invoked can indicate performance impacts. `dev-blue` completed 1,059 refreshes of `StopEvent` in 30 minutes, which translates to once per 1.7s, while `dev-green` completed 432 refreshes in the same time period, a rate of one refresh every 4s.

Performance-wise, these changes seem very impactful.

## What questions should reviewers consider?

1. This logic duplicates how we create the dataset in Lamp: upsert new records and evict old ones; is this a sign that the API is doing *too* much?
2. Do you think that the logic for a full load is really just a special case of a partial load?
3. Stylistically, should `timestamp` be a datetime or is it ok to have as a Unix timestamp?
4. Is there an better way to evict that would work better here? Like, to fully embrace the actor model, I imagine having each key track its own eviction but not sure if Mnesia invalidates some of those actor model assumptions


apps/state/lib/state/stop_event.ex
```diff
@@ -25,6 +25,131 @@ defmodule State.StopEvent do
...
+      |> then(&[StopEvent | &1])
+      |> List.to_tuple()
+
+    # Match spec format: {pattern, guards, result}
+    [{pattern, [{:andalso, {:"/=", :"$1", nil}, {:<, :"$1", cutoff}}], [:"$2"]}]
+  end
+
+  # Get the maximum timestamp from existing data in the table.
+  # Uses a dynamic match spec based on the StopEvent struct to avoid brittleness
+  # from hardcoded field positions.
```
lemald, line 112: **nitpick:** could make this a `@doc` block instead of just a comment.

apps/state/lib/state/stop_event.ex
```diff
@@ -25,6 +25,131 @@ defmodule State.StopEvent do
...
 
+  @impl State.Server
+  def handle_new_state(binary) when is_binary(binary) do
+    # Get the maximum timestamp from existing data
+    max_timestamp = get_max_timestamp()
+
+    # Parse with timestamp filtering if we have existing data
+    opts = if max_timestamp, do: [newer_than: max_timestamp], else: []
+
+    parser = Parse.StopEvents
```
lemald: **nitpick:** this is only used in one place and doesn't have any possible value other than `Parse.StopEvents` and not any test module or anything - could the variable be removed?

apps/state/lib/state/stop_event.ex
```diff
@@ -25,6 +25,131 @@ defmodule State.StopEvent do
...
+
+    parsed_data =
+      try do
+        parser.parse(binary, opts)
+      rescue
+        e ->
+          # log_parse_error returns nil, so we return nil on error to avoid passing
+          # invalid data to super/1. State.Server expects nil to mean
+          # "no data to insert" and will handle it gracefully.
```
lemald: **question:** I'm kind of confused by this comment because we don't actually pass `nil` to `super` below because we intercept it in the `case` statement and pass `:ok` instead.

apps/state/lib/state/stop_event.ex
```diff
@@ -25,6 +25,131 @@ defmodule State.StopEvent do
...
+  end
+
+  @spec evict_old_records(non_neg_integer()) :: :ok
+  def evict_old_records(retention_seconds) when retention_seconds >= 0 do
+    cutoff = System.system_time(:second) - retention_seconds
+    match_spec = build_eviction_match_spec(cutoff)
+
+    __MODULE__
+    |> :mnesia.dirty_select(match_spec)
+    |> Enum.each(&:mnesia.dirty_delete(__MODULE__, &1))
```
lemald: **question:** could anything weird possibly happen because of using `dirty_delete` here instead of deleting in a transaction? Could we have an API query return some of the `StopEvents` that are to be deleted but not others?

apps/state/test/state/stop_event_test.exs
```diff
@@ -391,4 +398,377 @@ defmodule State.StopEventTest do
...
+      # Trigger eviction with 2-hour retention (7200 seconds)
+      State.StopEvent.evict_old_records(7200)
+
+      # Old event should be evicted, borderline is exactly at boundary (should be kept),
+      # recent should remain
+      assert State.StopEvent.size() == 2
+      assert by_id("old-event") == nil
+      assert by_id("borderline-event") != nil
+      assert by_id("recent-event") != nil
+    end
```
lemald: **nitpick:**
```suggestion
      assert is_nil(by_id("old-event"))
      refute is_nil(by_id("borderline-event"))
      refute is_nil(by_id("recent-event"))
    end
```

apps/state/test/state/stop_event_test.exs
```diff
@@ -391,4 +398,377 @@ defmodule State.StopEventTest do
...
+
+      State.StopEvent.new_state([old_event, recent_event])
+      assert State.StopEvent.size() == 2
+
+      # Evict with 40 minute retention (2400 seconds)
+      State.StopEvent.evict_old_records(2400)
+
+      assert State.StopEvent.size() == 1
+      assert by_id("old-event") == nil
+      assert by_id("recent-event") != nil
```
lemald: **nitpick:**
```suggestion
      assert is_nil(by_id("old-event"))
      refute is_nil(by_id("recent-event"))
```

apps/state/lib/state/stop_event.ex
runkelcorey: [Official docs](https://elixir.hexdocs.pm/1.12.3/writing-documentation.html#documentation-code-comments) recommend against `doc`s in private functions and compilation will discard their contents

apps/state/lib/state/stop_event.ex
runkelcorey: probably; locally, it looks like it doesn't affect delete performance much, so I used the safe delete

apps/state/lib/state/stop_event.ex
lemald: Ah, right, I totally missed that this was a private function.

lemald (approved):

PR #1074, opened 2026-08-28
feat(ApiWeb): add `/transfers` endpoint

thecristen:
#### Summary of changes

**Asana Ticket:** [Follow up on "🛜 🐞 Daily Schedules sometimes doesn't show school trips "](https://app.asana.com/1/15492006741476/project/555089885850811/task/1217929539836227?focus=true) prerequisite

👋🏼 To facilitate displaying in-seat transfers to riders, I embarked on an effort to finally learn V3 API development in order to expose the `transfers.txt` GTFS data. Please let me know if there are pieces I overlooked or misplaced.

At a high level, this PR:
- Adds ingestion and parsing of GTFS `transfers.txt`
- Adds a controller that lets you query the transfers by trip (to `:from_trip_id`) and by type (i.e. `:transfer_type`). Not my prettiest code, most of it is copied from the `FacilitiesController` but it works ✨ 
  - Also requires at least one filter to use (e.g. it won't return all the transfers) - this was @lemald 's suggestion and I think it's a good one.
- Adds a view which shows the attributes, along `from_trip`, `to_trip`, `from_stop`, and `to_stop` modeled as relationships... open to feedback on that one. I had an earlier draft which showed these as `*_id` attributes instead, just not sure what the best choice was there.
- I used 🤖 only to add the unit tests, they looked reasonable to me but please keep me honest on that
- Finally, the `/status` endpoint is updated to show last updated timestamp for transfers

I also tried to flesh out the documentation as thoroughly as I could think of, happy to field feedback on that too!

### DIFF
--- a/apps/api_web/lib/api_web/controllers/transfer_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/transfer_controller.ex
@@ -49,22 +49,18 @@ defmodule ApiWeb.TransferController do
   end
 
   def index_data(conn, params) do
-    with :ok <- Params.validate_includes(params, @includes, conn) do
-      case Params.filter_params(params, @filters, conn) do
-        {:ok, filters} when map_size(filters) > 0 ->
-          filters
-          |> format_filters()
-          |> State.Transfer.filter_by()
-          |> State.all(Params.filter_opts(params, @pagination_opts, conn))
-
-        {:error, _, _} = error ->
-          error
-
-        _ ->
-          {:error, :filter_required}
-      end
+    with :ok <- Params.validate_includes(params, @includes, conn),
+         {:ok, filters} when map_size(filters) > 0 <- Params.filter_params(params, @filters, conn) do
+      filters
+      |> format_filters()
+      |> State.Transfer.filter_by()
+      |> State.all(Params.filter_opts(params, @pagination_opts, conn))
     else
-      {:error, _, _} = error -> error
+      {:error, _, _} = error ->
+        error
+
+      _ ->
+        {:error, :filter_required}
     end
   end
 
--- a/apps/api_web/lib/api_web/controllers/transfer_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/transfer_controller.ex
@@ -3,11 +3,10 @@ defmodule ApiWeb.TransferController do
   Controller for Transfers. Filterable by:
 
   * from_trip (multiple)
-  * type (multiple)
   """
   use ApiWeb.Web, :api_controller
 
-  @filters ~w(from_trip type)s
+  @filters ~w(from_trip)s
   @pagination_opts ~w(offset limit)a
 
   def state_module, do: State.Transfer
@@ -16,17 +15,17 @@ defmodule ApiWeb.TransferController do
     get(path(__MODULE__, :index))
 
     description("""
-    **NOTE:** `filter[type]` or `filter[from_trip]` **MUST** be present for any transfers to be returned.
+    **NOTE:** `filter[from_trip]` **MUST** be present for any transfers to be returned.
 
     List of transfers. Transfer specifies additional rules and overrides for a transfer between trips, routes, and/or stops.
-
-    ## Transfers of a certain type
-
-    `/transfers?filter[type]=TYPE`
 
     ## Transfers from a certain trip
 
     `/transfers?filter[from_trip]=TRIP_ID`
+
+    ## Transfers from a set of trips
+
+    `/transfers?filter[from_trip]=TRIP_ID1,TRIP_ID2,TRIP_ID3`
     """)
 
     common_index_parameters(__MODULE__, :transfer)
@@ -36,13 +35,6 @@ defmodule ApiWeb.TransferController do
       :query,
       :string,
       "Filter by trip ID. Multiple trips #{comma_separated_list()}."
-    )
-
-    parameter(
-      "filter[type]",
-      :query,
-      :string,
-      "Filter by transfer type. Multiple types #{comma_separated_list()}."
     )
 
     consumes("application/vnd.api+json")
@@ -82,16 +74,6 @@ defmodule ApiWeb.TransferController do
 
       trip_ids ->
         %{from_trip_ids: trip_ids}
-    end
-  end
-
-  defp do_format_filter({"type", type_string}) do
-    case Params.split_on_comma(type_string) do
-      [] ->
-        []
-
-      types ->
-        %{types: types}
     end
   end
 
--- a/apps/api_web/lib/api_web/controllers/transfer_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/transfer_controller.ex
@@ -8,6 +8,7 @@ defmodule ApiWeb.TransferController do
 
   @filters ~w(from_trip)s
   @pagination_opts ~w(offset limit)a
+  @includes ~w(from_trip to_trip from_stop to_stop)
 
   def state_module, do: State.Transfer
 
@@ -30,6 +31,8 @@ defmodule ApiWeb.TransferController do
 
     common_index_parameters(__MODULE__, :transfer)
 
+    include_parameters()
+
     parameter(
       "filter[from_trip]",
       :query,
@@ -46,18 +49,22 @@ defmodule ApiWeb.TransferController do
   end
 
   def index_data(conn, params) do
-    case Params.filter_params(params, @filters, conn) do
-      {:ok, filters} when map_size(filters) > 0 ->
-        filters
-        |> format_filters()
-        |> State.Transfer.filter_by()
-        |> State.all(Params.filter_opts(params, @pagination_opts, conn))
+    with :ok <- Params.validate_includes(params, @includes, conn) do
+      case Params.filter_params(params, @filters, conn) do
+        {:ok, filters} when map_size(filters) > 0 ->
+          filters
+          |> format_filters()
+          |> State.Transfer.filter_by()
+          |> State.all(Params.filter_opts(params, @pagination_opts, conn))
 
-      {:error, _, _} = error ->
-        error
+        {:error, _, _} = error ->
+          error
 
-      _ ->
-        {:error, :filter_required}
+        _ ->
+          {:error, :filter_required}
+      end
+    else
+      {:error, _, _} = error -> error
     end
   end
 
@@ -160,5 +167,20 @@ defmodule ApiWeb.TransferController do
         end,
       Transfer: page(:TransferResource)
     }
+  end
+
+  defp include_parameters(schema) do
+    ApiWeb.SwaggerHelpers.include_parameters(
+      schema,
+      @includes,
+      description: """
+      | include     | Description |
+      |-------------|-------------|
+      | `from_trip` | The trip where a connection between routes begins |
+      | `to_trip`   | The trip where a connection between routes ends |
+      | `from_stop` | The stop or station where a connection between routes begins |
+      | `to_stop`   | The stop or station where a connection between routes ends |
+      """
+    )
   end
 end
--- a/apps/api_web/lib/api_web/controllers/transfer_controller.ex
+++ b/apps/api_web/lib/api_web/controllers/transfer_controller.ex
@@ -2,12 +2,12 @@ defmodule ApiWeb.TransferController do
   @moduledoc """
   Controller for Transfers. Filterable by:
 
-  * trip (multiple)
+  * from_trip (multiple)
   * type (multiple)
   """
   use ApiWeb.Web, :api_controller
 
-  @filters ~w(trip type)s
+  @filters ~w(from_trip type)s
   @pagination_opts ~w(offset limit)a
 
   def state_module, do: State.Transfer
@@ -16,7 +16,7 @@ defmodule ApiWeb.TransferController do
     get(path(__MODULE__, :index))
 
     description("""
-    **NOTE:** `filter[type]` or `filter[trip]` **MUST** be present for any transfers to be returned.
+    **NOTE:** `filter[type]` or `filter[from_trip]` **MUST** be present for any transfers to be returned.
 
     List of transfers. Transfer specifies additional rules and overrides for a transfer between trips, routes, and/or stops.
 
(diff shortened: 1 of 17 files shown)


apps/api_web/lib/api_web/controllers/transfer_controller.ex
```diff
@@ -0,0 +1,182 @@
...
+              """,
+              enum: Enum.to_list(0..2),
+              example: 0
+            )
+          end
+
+          relationship(:from_stop)
+          relationship(:to_stop)
+          relationship(:from_trip)
+          relationship(:to_trip)
```
KaylaBrady, line 177: Thank you for adding transfers to the API 🙌 

question(non-blocking): With this PR establishing the relationship from transfer => trips, do you have a sense of if it would be possible to later add a relationship from trip => transfer?

Definitely out of scope for this PR, but I wonder if it is feasible to support a query like `/trips/trip_id?include=transfers`

apps/model/lib/model/transfer.ex
```diff
@@ -0,0 +1,73 @@
...
+  | 0 or empty | Recommended between routes |
+  | 1          | Timed between two routes |
+  | 2          | Requires a minimum amount of time to ensure connection |
+  | 3          | Transfers not possible between routes at the location |
+  | 4          | In-seat transfer between sequential trips |
+  | 5          | In-seat transfers not allowed between sequential trips |
+
+  See [GTFS `transfers.txt` `transfer_type`](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#transferstxt)
+  """
+  @type transfer_type :: 0..5
```
thecristen, line 38: Oh! I forgot to note I was considering whether we'd want to adjust the `:transfer_type` enum to be more descriptive than integers, like we've done for a few other fields. But I'm leaning towards "meh, probably not" because the meaning of each one feels nuanced enough that it's hard to capture in a pithy phrase anyway.

Like, every phrase I could think of feels a bit awkward 😂 

  | Value      | Description | Potential "friendly" value |
  |------------|-------------|-----|
  | 0 or empty | Recommended between routes | `RECOMMENDED_BETWEEN_ROUTES` |
  | 1          | Timed between two routes | `TIMED_BETWEEN_ROUTES` |
  | 2          | Requires a minimum amount of time to ensure connection | `REQUIRES_MIN_TIME` |
  | 3          | Transfers not possible between routes at the location | `NO_TRANSFER_WITHIN_LOCATION` |
  | 4          | In-seat transfer between sequential trips | `IN_SEAT_BETWEEN_TRIPS` |
  | 5          | In-seat transfers not allowed between sequential trips | `OUT_SEAT_BETWEEN_TRIPS` |

apps/api_web/lib/api_web/views/transfer_view.ex
```diff
@@ -0,0 +1,70 @@
...
+          from_stop_id: from_stop_id,
+          to_trip_id: to_trip_id,
+          to_stop_id: to_stop_id
+        },
+        _conn
+      ) do
+    from_trip = from_trip_id || ""
+    to_trip = to_trip_id || ""
+    from_route = route_id(from_trip)
+    to_route = route_id(to_trip)
```
lemald: **question:** maybe I'm missing something, but what is the purpose of including the route ID(s) in the ID? Is it just for improved legibility?

apps/model/lib/model/transfer.ex
lemald: Yeah I think leaving it this way is probably fine. There are other places like the `type` field on routes where we similarly leave an enum in integer form in the API response so it's not breaking existing patterns or anything.

apps/api_web/lib/api_web/controllers/transfer_controller.ex
```diff
@@ -0,0 +1,182 @@
...
+    produces("application/vnd.api+json")
+    response(200, "OK", Schema.ref(:Transfer))
+    response(400, "Bad Request", Schema.ref(:BadRequest))
+    response(403, "Forbidden", Schema.ref(:Forbidden))
+    response(429, "Too Many Requests", Schema.ref(:TooManyRequests))
+  end
+
+  def index_data(conn, params) do
+    case Params.filter_params(params, @filters, conn) do
+      {:ok, filters} when map_size(filters) > 0 ->
```
lemald, line 58: **issue:** so I worry a bit that by allowing only a filter by type, and allowing the type filter to take a comma-separated list of types, it's still trivially easy to write a query that would return all of the transfers. I'm not sure what to do about it, though, because we've presumably got to allow _some_ filter more general than trip IDs for your use cases on dotcom.

lemald: Maybe we could deploy to a dev environment and see what response times are actually like when filtering by multiple types to get a sense of how much of an issue it actually is.

apps/api_web/lib/api_web/controllers/transfer_controller.ex
```diff
@@ -0,0 +1,182 @@
...
+              "x-nullable": true,
+              example: 7
+            )
+
+            suggested_buffer_time(
+              :integer,
+              "Experimental. Recommended buffer time to allow to make a successful transfer between two services. This is also partly based on the significance of missing the transfer (due to service frequency).",
+              "x-nullable": true,
+              example: 5
+            )
```
lemald, line 138: **suggestion:** it would be nice to also mention the units of these fields in the docs.

apps/api_web/lib/api_web/views/transfer_view.ex
thecristen: ah, meant to leave a comment about this, thanks for asking! I did this based on the GTFS documentation's description of this table's [primary key](https://github.com/google/transit/blob/master/gtfs/spec/en/reference.md#transferstxt). It's a bit odd to account for here since our `transfers.txt` doesn't actually have `from_route_id` or `to_route_id` columns (though it _could_...).
However, it seems that all rows in our current transfers have unique combinations of `from_stop_id`, `to_stop_id`, `from_trip_id`, and `to_trip_id`, so maybe it makes sense to drop the route IDs?

thecristen: funny enough, it doesn't improve legibility one bit in some cases :joy:

image

apps/api_web/lib/api_web/controllers/transfer_controller.ex
thecristen: I admittedly don't know how implementing includes works, but I think with the rest of these pieces in place, it'd be a relatively light lift!

apps/api_web/lib/api_web/controllers/transfer_controller.ex
thecristen: Will add. Just realized [our own doc](https://github.com/mbta/gtfs-documentation/blob/master/reference/gtfs.md#transferstxt) doesn't say, but I'm going to assume our experimental fields use seconds to be consistent with the standard `min_transfer_time` field.

apps/api_web/lib/api_web/views/transfer_view.ex
thecristen: Removed in  d3c129b1928b337d446ccb17a79c1fa1d32961fb

apps/api_web/lib/api_web/controllers/transfer_controller.ex
thecristen: Added in [docs(TransferController): specify seconds](https://github.com/mbta/api/pull/1074/commits/8267d694f8f729da9336803f1cd099a8384af045)

apps/parse/lib/parse/transfers.ex
```diff
@@ -0,0 +1,60 @@
...
+      from_stop_id: copy(row["from_stop_id"]),
+      to_stop_id: copy(row["to_stop_id"]),
+      min_transfer_time: to_integer_if_not_blank(row["min_transfer_time"]),
+      min_walk_time: to_integer_if_not_blank(row["min_walk_time"]),
+      min_wheelchair_time: to_integer_if_not_blank(row["min_wheelchair_time"]),
+      suggested_buffer_time: to_integer_if_not_blank(row["suggested_buffer_time"]),
+      wheelchair_transfer: to_integer(row["wheelchair_transfer"]),
+      from_trip_id: copy_if_not_blank(row["from_trip_id"]),
+      to_trip_id: copy_if_not_blank(row["to_trip_id"]),
+      transfer_type: String.to_integer(row["transfer_type"])
```
lemald: **question:** should this also be `to_integer_if_not_blank`, since  `transfer_type` can in theory be empty (not sure if we actually do that in our current GTFS)?

thecristen: Ahh I think so - I wrote this before realizing that although the field is required its value can be empty. Thanks for catching 👍🏼

thecristen: Refined in [fixup(Parse.Transfers): possibly empty values](https://github.com/mbta/api/pull/1074/commits/faa74f966eeb72a8f515591f03e9b2146ef3c0b9)

apps/api_web/lib/api_web/controllers/transfer_controller.ex
thecristen: I ended up figuring out how to implement it in [feat(TripView): include transfers](https://github.com/mbta/api/pull/1074/commits/4816ce39203674653123b87202cd9a1dc2014432)

apps/api_web/lib/api_web/controllers/transfer_controller.ex
thecristen: Let's send this to a dev environment! Maybe you can pick which environment/when? Don't want to interfere with anything else that might be going on.

apps/api_web/lib/api_web/controllers/transfer_controller.ex
KaylaBrady: Amazing!!

apps/api_web/lib/api_web/controllers/transfer_controller.ex
lemald: It looks like `dev-blue` auto-deploys the latest `master` every night, so you probably want `dev-green` if you don't want to have to redeploy every day. It doesn't look like anyone else is using `dev-green` right now.

apps/api_web/lib/api_web/views/trip_view.ex
```diff
@@ -105,6 +113,13 @@ defmodule ApiWeb.TripView do
     optional_relationship("route_pattern", route_pattern_id, &RoutePattern.by_id/1, conn)
   end
 
+  def transfers(%{id: trip_id}, _conn) do
+    case Transfer.by_from_trip_id(trip_id) do
```
lemald: **question:** so to confirm, including transfers for a given trip will give you transfers _from_ that trip, and not to it? If so I think I would argue for calling the `include` argument something else if possible, like `from_trip_transfers`.

apps/api_web/lib/api_web/controllers/transfer_controller.ex
thecristen: Standby while I work on making it ready for prime time - I'm missing a few attributes

image

apps/api_web/lib/api_web/views/trip_view.ex
thecristen: image

Yeah, I just realized that might've been flawed now that I'm trying to use it... 😩 . Will re-do

apps/api_web/lib/api_web/controllers/transfer_controller.ex
lemald: Re: the volume concern, it looks like our current GTFS (with the just-recently-released Fall rating included) is 13930 entries, so that's the kind of scale we're talking about.

apps/api_web/lib/api_web/controllers/transfer_controller.ex
thecristen: Actually the above was just a parsing issue on Dotcom's end.

It's on dev-green if you'd like to try it! e.g. https://api-dev-green.mbtace.com/trips/76334596?include=transfers

thecristen: Renamed this include to `from_trip_transfers`, in case you look at this later :)

apps/api_web/lib/api_web/views/trip_view.ex
thecristen: Updated in [feedback: rename include to from_trip_transfers](https://github.com/mbta/api/pull/1074/commits/3b92cce72b4d5e7af861173c352671ec624fe5de), excellent tip

apps/api_web/lib/api_web/controllers/transfer_controller.ex
thecristen: Put it on dev-green! Here's some observations:

```
/transfers?filter[type]=0	167 ms
/transfers?filter[type]=1	162 ms
/transfers?filter[type]=2	217 ms
/transfers?filter[type]=3	27 ms
/transfers?filter[type]=4	1.01 s
```

As a reference point, even these available hefty queries still are much quicker:

```
/vehicles - 323ms
/schedules?filter[route]=1&filter[direction_id]=0 - 312ms
```

So yeah, I guess it is too much! Your point about it being trivially easy to query for all transfers anyway... haha, oops!

Some ways forward I see

1. Make `/transfers` more performant / limit the query more somehow
2. Don't expose `/transfers` directly, but make it available as an include. 

https://api-dev-green.mbtace.com/trips/76334596?include=transfers (to be imminently renamed to `include=from_trip_transfers`) works, and seems like it could meet Dotcom's and Mobile App's needs. Curious what we think!

KaylaBrady: From the app side, I agree that the include approach should meet our needs!

lemald: I support the idea of making it an include-able resource without a top-level endpoint if that works for both teams! Alternatively we could keep the endpoint but only with the `from_trip_id` filter and not `type`.

lemald (approved):
Looks good to me with all the latest changes!

PR #1077, opened 2026-09-01
chore(deps): update bandit and xml_builder

lemald:
#### Summary of changes

**Asana Ticket:** none, ad-hoc

Updates a couple of packages for which `mix hex.audit` was returning issues.


rudiejd (approved):
verified diffs for each upgrade and didn't see anything blocking or dangerous, seems good to me 👍

